PR #14, opened 2017-09-05
This change to cmake is needed in order for me to configure on Fedora.

stevenrbrandt:
This change is needed for my Phylanx to pass configure on Fedora.


CMakeLists.txt
```diff
@@ -97,7 +97,9 @@ else()
   phylanx_error("HPX_DIR has not been specified, please set it to help locating HPX")
 endif()
 
-find_package(Eigen3 REQUIRED NO_CMAKE_PACKAGE_REGISTRY)
+find_package(PkgConfig)
+pkg_search_module(Eigen3 REQUIRED eigen3)
```
hkaiser: This will not work on platforms not supporting pkgconfig. We could have your code as a fallback, but it will not work as the main detection mechanism.

CMakeLists.txt
```diff
@@ -192,11 +194,11 @@ if(NOT MSVC)
   # Check if our libraries have unresolved symbols
   #if(NOT APPLE AND NOT HPX_WITH_APEX)
   if(NOT APPLE AND NOT WIN32 AND NOT PHYLANX_WITH_SANITIZERS)
-    phylanx_add_link_flag_if_available(-Wl,-z,defs TARGETS SHARED EXE)
+    hpx_add_link_flag_if_available(-Wl,-z,defs TARGETS SHARED EXE)
```
hkaiser: This shouldn't be necessary anymore, the missing cmake macro was added here: 72117850cc7154b20edc7e5397cfb50f78d84853.

CMakeLists.txt
```diff
@@ -192,11 +194,11 @@ if(NOT MSVC)
...
   endif()
 
   if("${HPX_PLATFORM_UC}" STREQUAL "BLUEGENEQ")
-    phylanx_add_compile_flag_if_available(-Wno-deprecated-register LANGUAGES CXX C)
+    hpx_add_compile_flag_if_available(-Wno-deprecated-register LANGUAGES CXX C)
```
hkaiser: Same as above.

stevenrbrandt:
I've updated the CMakeList.txt file to use PkgConfig as a fallback mechanism.

PR #15, opened 2017-09-05
Tests

rtohid:
Minor changes to get tests compile and pass.

### DIFF
--- a/phylanx/ir/node_data.hpp
+++ b/phylanx/ir/node_data.hpp
@@ -155,13 +155,13 @@ namespace phylanx { namespace ir
             dimensions_type dimensions_;
 
             // support for intrusive_ptr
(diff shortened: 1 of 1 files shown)


phylanx/ir/node_data.hpp
```diff
@@ -155,13 +155,13 @@ namespace phylanx { namespace ir
             dimensions_type dimensions_;
 
             // support for intrusive_ptr
-            template <typename T>
-            friend void intrusive_ptr_add_ref(node_data_storage_base<T>* p)
+            template <typename _T>
```
hkaiser: Please don't use leading underscores in identifiers. Those names are reserved for the implementation. Could you use `T_` instead, for instance, please?

PR #18, opened 2017-09-07
Fixing gcc7

hkaiser:


hkaiser:
Initially I was planning to wait with this until the CircleCI tests are in place, but master is currently exposing a lot of problems for Linux users - so I will go ahead and merge this.

PR #20, opened 2017-09-10
fixed Eigen::Constant type reference issue

ct-clmsn:
fixes type resolution issue when calling the Eigen static Constant method.

### DIFF
--- a/phylanx/ir/node_data.hpp
+++ b/phylanx/ir/node_data.hpp
@@ -273,6 +273,8 @@ namespace phylanx { namespace ir
             using storage1d_type =
                 Eigen::Matrix<T, Eigen::Dynamic, 1>;
             using base_type = node_data_storage_base<T>;
+            using constant_type =
+                typename Eigen::DenseBase<storage1d_type>;
 
             node_data_storage()
               : node_data_storage_base<T>(1)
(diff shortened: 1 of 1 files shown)


ct-clmsn:
found a small hpx header file missing issue. fixed the issue in hpx and added a [PR](https://github.com/STEllAR-GROUP/hpx/pull/2892).

src/primitives/add_operation.cpp
```diff
@@ -175,7 +175,7 @@ namespace phylanx { namespace primitives
     // implement '+' for all possible combinations of lhs and rhs
     hpx::future<ir::node_data<double>> add_operation::eval()
     {
-        return hpx::dataflow(hpx::util::unwrapping(
+        return hpx::dataflow(hpx::util::unwrapped(
```
hkaiser: This has been recently renamed. What version of HPX should we target with Phylanx? Would it be ok to stick with HPX master for the foreseeable future?

ct-clmsn: my poor git utilization strikes again! apologies, will amend.

ct-clmsn: @hkaiser do i need to do anything to drop that last commit for 'hpx::util::unwrapping'?

hkaiser: @ct-clmsn it should be fine to remove the commit from your local branch (reset to b18e6332b070e17a7f32297ea0a5f88e91d765f8) and force-push to github.

ct-clmsn:
@hkaiser should be rolled back now. thanks!

PR #21, opened 2017-09-13
Circle-CI

rtohid:
This successfully builds Phylanx and all of it's dependencies on Circle-CI, however, tests still fail!


hkaiser:
@rtohid Thanks a lot! This is very important!

PR #27, opened 2017-09-18
Adding file_read and file_write primitives

hkaiser:
- the file_read pattern is `file_read(_1)`, where `_1` is representing the
  file name to read data from
- the file_write pattern is `file_write(_1, _2)`, where `_1` is 
  representing the file name to write the data to that is represented
  by `_2`

- both facilities operate on serialized `ir::node_data<double>` byte streams

- adding corresponding tests
- introduced literal values in AST (bool, double, uint64, string)
- this changes construction API for primitives, those now expect to take two
  arguments: a vector of literal values and a vector of other primitives
- updated add_primitive to be able to handle literal values in addition
  to other primitives

- flyby: adding literal_value test


hkaiser:
@hapoo That's fixed now, thanks!

hapoo:
LGTM

PR #33, opened 2017-09-22
Use Phylanx from Python

stevenrbrandt:
This pull request creates ...
(1) a python file, Phylanx2.py, which constructs an AST without using a parser.
(2) a function called util.append_operation() because, apparently, appending to the rest member of expressions does not work.
(3) a function util.stringify() which renders an AST in human-readable form.


src/ast/node.cpp
```diff
@@ -376,6 +377,10 @@ namespace phylanx { namespace ast
     std::ostream& operator<<(std::ostream& out, operand const& op)
     {
         out << "operand";
+        int ind = op.index();
+        if(ind == 1) {
+          auto ident = phylanx::util::get<1>(op.get());
+        }
```
hkaiser: Shouldn't this print something for `ind == 2` as well? Wouldn't it be better to use visit()?

src/util/serialization/ast.cpp
```diff
@@ -142,6 +143,139 @@ namespace phylanx { namespace util
         detail::unserialize_helper(input, expr);
         return expr;
     }
-}}
 
+    // You cannot add to expression.rest unless you
+    // use this function. Note sure why.
+    void append_operation(ast::expression& ast,ast::operation const& o) {
+      ast.rest.push_back(o);
+    }
```
hkaiser: `expression has a member `append()` just for this (see: https://github.com/STEllAR-GROUP/phylanx/blob/master/phylanx/ast/node.hpp#L472-L487)

hkaiser:
Please use the `.clang-format` file in the root of this repository to format your code.

hkaiser:
Also, could you please add a test verifying the `stringify` functionality?

src/ast/node.cpp
stevenrbrandt: Sorry, that was a mistake. I reverted it.

src/util/serialization/ast.cpp
stevenrbrandt: OK, I thought that was the built-in pybind11 command for lists. It didn't work for me. Calling it left the list empty.

hkaiser: Ok, good point - I'll have a look and fix this. Thanks!

stevenrbrandt:
What verison of clang do I need to use for the .clang-format? I get an error from both 3.8.1 and 3.9.0 about 
```
YAML:26:23: error: unknown key 'AlignEscapedNewlines'
AlignEscapedNewlines: Right
                      ^~~~~
```

aserio:
@stevenrbrandt I got clang-format to work with Clang v5.0.0

tests/unit/python/ast/python_builds_ast.py
```diff
@@ -0,0 +1,86 @@
+import phylanx.ast as ast
```
hkaiser: Please add copyright and license to this file

stevenrbrandt:
Hartmut, I've added the copyrights.

hkaiser:
@stevenrbrandt AppVeyor is still failing. I think you missed to explicitly export the stringify functions (add a `PHYLANX_EXPORT`) for them to be visible outside of the shared library.

hkaiser:
@stevenrbrandt What was your rationale of adding another set of functions (`stringify_xxx`) instead of extending the existing `operator<<()` functionality to give you the textual representation of the expressions?

stevenrbrandt:
You were already using that for a different representation, one that showed the tree structure more, and I didn't want to mess that up.

hkaiser:
> You were already using that for a different representation, one that showed the tree structure more, and I didn't want to mess that up.

@stevenrbrandt This was just a preliminary implementation which has been updated since (see https://github.com/STEllAR-GROUP/phylanx/blob/master/src/ast/node.cpp#L423-L541). Could you please verify if these (new) streaming operators expose the functionality you need? Should we amend them in any way?

This should also bind those operators to the existing Python `__str__` functionality (see https://github.com/STEllAR-GROUP/phylanx/blob/master/python/src/phylanx.cpp#L113-L119).

...sorry for any possible misunderstandings, I should have made clear (at least in a comment or so) that the old streaming operators were just placeholders.

phylanx/util/serialization/ast.hpp
```diff
@@ -33,6 +33,15 @@ namespace phylanx { namespace util
     PHYLANX_EXPORT std::vector<char> serialize(ast::expression const&);
     PHYLANX_EXPORT std::vector<char> serialize(ast::function_call const&);
 
+    PHYLANX_EXPORT void append_operation(ast::expression &,ast::operation const &);
+    PHYLANX_EXPORT std::string stringify_optoken(ast::optoken);
```
hkaiser: I think these declarations can be removed now.

py/Phylanx2.py
```diff
@@ -0,0 +1,86 @@
+# Copyright (c) 2017 Hartmut Kaiser
```
hkaiser: Shouldn't this copyright be under your name?

tests/unit/python/ast/python_builds_ast.py
```diff
@@ -0,0 +1,91 @@
+# Copyright (c) 2017 Hartmut Kaiser
```
hkaiser: Same here, please attribute the copyright to yourself.

stevenrbrandt:
It won't pass the test I put in until the fix_lists PR is in. On the other hand, this PR will provide the test for fix_lists. Should I combine these two pull requests?

hkaiser:
No need to combine things., Please fix the small issues I mentioned above and we'll merge this..

PR #36, opened 2017-09-27
The build failed because a directory was missing. This corrects the p…

stevenrbrandt:
This simple change fixes the build problem on linux.


hkaiser:
It still fails installing:
```
/usr/bin/python3.5: can't open file '/phylanx/build/python/setup_.py': [Errno 2] No such file or directory
```

stevenrbrandt:
My initial fix didn't fix the install, only the build. This change was more than 1 line of code, but install works.

python/CMakeLists.txt
```diff
@@ -40,18 +41,18 @@ if (PYTHON_EXECUTABLE)
...
       set(PYTHON_PACKAGE_NAME "_phylanx")
     endif()
-    configure_file(${SETUP_PY_IN} "${SETUP_PY}_${__config}.py" @ONLY)
+    configure_file(${SETUP_PY_IN} "${SETUP_PY}_${__configs}.py" @ONLY)
   endforeach()
 
   configure_file(${INIT_PY_IN} "${INIT_PY}" @ONLY)
 
-  set(__setup_script "${SETUP_PY}_$<CONFIG>.py")
+  set(__setup_script "${SETUP_PY}_${__configs}.py")
```
hkaiser: Those changes `__config` --> `__configs` are definitely not correct. On multi-target platforms (i.e. XCode or Visual Stdio you have more than one possible configuration, for instance `${__configs} == "Debug;Release"`

stevenrbrandt: I see.
Well, __config was not getting set.

stevenrbrandt: Ah, it just needs to be in a loop

hkaiser: It was getting set [here](https://github.com/STEllAR-GROUP/phylanx/pull/36/files/f34b9d6a1aaf4f8052f144b7efde431b92a06a6c#diff-79b695dff65b8b0a69bfed14e824cb18R43)

hkaiser: Is `CMAKE_BUILD_TYPE` set for you?

stevenrbrandt: Yes, I set CMAKE_BUILD_TYPE explicitly, give me a sec....

stevenrbrandt: OK, see if this new one works better for you.

python/CMakeLists.txt
```diff
@@ -41,38 +41,41 @@ if (PYTHON_EXECUTABLE)
...
+    source_group("Setup" FILES ${setup_files})
+    set_source_files_properties(${setup_files} PROPERTIES HEADER_FILE_ONLY TRUE)
 
-  # for the install step it does not matter which setup.py we use
-  install(CODE
-    "execute_process(COMMAND ${PYTHON_EXECUTABLE} ${SETUP_PY}_${__configs}.py install --user)")
+    # for the install step it does not matter which setup.py we use
+    install(CODE
+      "execute_process(COMMAND ${PYTHON_EXECUTABLE} ${SETUP_PY}_${__config}.py install --user)")
+  endforeach()
```
hkaiser: I think the installation does not need to be performed more than once. The initial idea was to execute only the `setup_${__config}.py` script for one of the available configuration targets as the generated scripts all do the same thing for the installation (they differ for the build step only) - hence the comment...

stevenrbrandt: I agree, but I figured there was no harm. In any event, I've pushed a change that only runs it once. I think that my changes have nothing to do with the failing tests, however.

hkaiser: The failing C++ tests are a (known) HPX problem, please disregard those. The Python tests however seem to run fine.

stevenrbrandt:
What's preventing us from merging this? AFAIK, I have addressed all the issues. Thanks.

hkaiser:
Well, now the appveyor build is broken...

PR #38, opened 2017-09-29
adding multiply primitive

hapoo:
This adds multiply primitive. It works for vector of operands when:
1. all operands are double
2. all operands are matrices
3. when there is mix of doubles and matrices; except when the first operand is double, in this case it only multiplies first to operands and throws exception for 3 operands and more.

issue: when tested with matrices larger than 100 by 100 got this:
Stack overflow in coroutine at address 0x00007f1286f2fbb8. Configure the hpx runtime to allocate a larger coroutine stack size. But the logic works and tests pass for smaller matrices.


phylanx/execution_tree/primitives/mul_operation.hpp
```diff
@@ -0,0 +1,49 @@
...
+            protected:
+            ir::node_data<double> mul0d(operands_type const& ops) const;
+            ir::node_data<double> mulxd(operands_type const& ops) const;
+
+            private:
+            std::vector<ast::literal_value_type> literals_;
+            std::vector<primitive> operands_;
+        };
+    }}}
+
```
hkaiser: Something went wrong with the indentation here.

src/execution_tree/primitives/mul_operation.cpp
```diff
@@ -0,0 +1,200 @@
...
+                                                                 ast::detail::literal_value(val));
+                                                     }
+
+                                                     HPX_ASSERT(p.valid());
+                                                     return p.eval();
+                                                 })
+                );
+            }
+        }}}
+
```
hkaiser: Same here, indentation gone wrong

hkaiser:
May I suggest you reformat all source files using clang-format (there is a .clang-format configuration file in the root of this repository).

PR #42, opened 2017-10-03
Fix lists

stevenrbrandt:
Calling the append function on expression.rest and function_call.args does nothing. This pull request fixes the problem. I will add a test once my other pull request is approved.

### DIFF
--- a/python/src/phylanx.cpp
+++ b/python/src/phylanx.cpp
@@ -328,7 +328,7 @@ PYBIND11_MODULE(_phylanx, m)
     // list of phylanx::ast::operations
     pybind11::class_<std::list<phylanx::ast::operation>>(ast,"operation_list","A list of operations")
         .def(pybind11::init<>())
-        .def("pop_back",&std::list<phylanx::ast::operation>::clear)
+        .def("pop_back",&std::list<phylanx::ast::operation>::pop_back)
         .def("append",[](std::list<phylanx::ast::operation>& v,const phylanx::ast::operation& f) { v.push_back(f); })
         .def("__len__",[](const std::list<phylanx::ast::operation>& v) { return v.size(); })
         .def("__iter__",[](std::list<phylanx::ast::operation>& v) {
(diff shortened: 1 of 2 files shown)


python/src/phylanx.cpp
```diff
@@ -321,6 +325,26 @@ PYBIND11_MODULE(_phylanx, m)
...
+        .def("append",[](std::list<phylanx::ast::operation>& v,const phylanx::ast::operation& f) { v.push_back(f); })
+        .def("__len__",[](const std::list<phylanx::ast::operation>& v) { return v.size(); })
+        .def("__iter__",[](std::list<phylanx::ast::operation>& v) {
+            return pybind11::make_iterator(v.begin(), v.end());
+        }, pybind11::keep_alive<0,1>());
+
+    // list of phylanx::ast::expressions
+    pybind11::class_<std::list<phylanx::ast::expression>>(ast,"expression_list","A list of expressions")
+        .def(pybind11::init<>())
+        .def("pop_back",&std::list<phylanx::ast::expression>::clear)
```
hkaiser: This should probably call `std::list<>::pop_back()`, not `std::list<>::clear()`.

python/src/phylanx.cpp
```diff
@@ -321,6 +325,26 @@ PYBIND11_MODULE(_phylanx, m)
             &phylanx::bindings::pickle_helper<phylanx::ast::expression>,
             &phylanx::bindings::unpickle_helper<phylanx::ast::expression>));
 
+    // list of phylanx::ast::operations
+    pybind11::class_<std::list<phylanx::ast::operation>>(ast,"operation_list","A list of operations")
+        .def(pybind11::init<>())
+        .def("pop_back",&std::list<phylanx::ast::operation>::clear)
```
hkaiser: This should probably call `std::list<>::pop_back()`, not `std::list<>::clear()`.

stevenrbrandt: I fixed this.

python/src/phylanx.cpp
stevenrbrandt: I fixed this.

PR #52, opened 2017-10-08
Adding unary_minus and unary_not primitives

hkaiser:


hkaiser:
The reported problems on circleci are caused by a problem in HPX

PR #53, opened 2017-10-08
Making sure primitives don't trigger any unneeded allocations and/or argument copy operations

hkaiser:
This PR changes all primitives to avoid allocations and copy operations as much as possible


hkaiser:
The remaining problems are related to a problem in HPX

PR #57, opened 2017-10-09
Buildbot scripts

khuck:
Requesting to merge the buildbot scripts directory into master.  The only changes in this merge are to add the directory of configuration files and build scripts.


hkaiser:
Is this really something which should go into this repository? I.e. should everybody interested in Phylanx have to download this?

khuck:
If you want us to be the single point of success/failure of buildbot support, then no. :)

That depends - this directory can be excluded when creating releases, so it wouldn't have to be in tarball downloads.  If you have strong objections to having it in the repo, then we don't have to add it.

hkaiser:
> If you want us to be the single point of success/failure of buildbot support, then no. :)

LOL - any repository would do the trick for you to plausibly deny responsibility 

> That depends - this directory can be excluded when creating releases, so it wouldn't have to be in tarball downloads. If you have strong objections to having it in the repo, then we don't have to add it.

That's true enough - github creates tarballs from the whole repo, though (...we could however exclude this directory from the respective release branches...). So you might be right - it's not perfect either way. What do others think about this?

khuck:
I have to put the buildbot configuration files into a private repo anyway, so I could put them there, and make it a submodule of the Phalanx project?

khuck:
Actually, the more I think about it, the more a submodule makes sense.  I'll withdraw this pull request...

PR #62, opened 2017-10-11
Logistic regression algorithm

hkaiser:
This adds a couple of primitives (see also #58 and #61) and implements a first version of the logistic regression algorithm based on the Phylanx infrastructure.

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -4,6 +4,7 @@
 # file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
 
 set(subdirs
+    algorithms
     ast
    )
 
--- a/examples/algorithms/CMakeLists.txt
+++ b/examples/algorithms/CMakeLists.txt
@@ -1,0 +1,35 @@
+# Copyright (c) 2017 Hartmut Kaiser
+#
+# Distributed under the Boost Software License, Version 1.0. (See accompanying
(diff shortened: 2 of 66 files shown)


ct-clmsn:
@hkaiser just noticed that ya'll have implemented the training version of this algorithm. There is a prediction algorithm which uses the output weights from the training algorithm to make predictions over a new input set of features (a set of vectors with dimension N, typically stored into a matrix). The output from the prediction algorithm should yield a vector of integer values. Pseudo-code: weights = train(X,Y); newY = predict(newX, weights); ... newY should contain integer values in the range {min(Y), max(Y)}.

justwagle:
@ct-clmsn @hkaiser  Hi Chris, yes, the algorithm at the moment trains the model and outputs the weights. It is my fault for missing it in the first algorithm. It was my understanding that once we have the correct weights from the training set in Phylanx, implementing the prediction function would not require any new primitives as the prediction function would simply need to calculate the dot product of the testing set with the weights obtained from the training part. 
The predict function should be very similar to:
predict(test_x, weights_from_training):
        final_scores = np.dot(test_x, weights_from_training)
        pred = sigmoid(final_scores)
If it is ok , I can create a ticket noting that prediction function needs to be added to the logistic regression algorithm once we have verified the weights obtained in training in Phylanx matches reference implementation in python.

ct-clmsn:
@justwagle not a problem! your understanding is correct (if you have enough primitives to train, you have enough primitives to predict). i wasn't sure if that was something being tracked internally or if it needed to be annotated in the issue tracker. thank you for the quick response!

PR #64, opened 2017-10-11
Enable to specify more than one match-pattern for a primitive

### LINKED ISSUE #63: Multiple Matches Required
In order to match the two supported use cases of `if` (i.e. `if(cond, true_case, false_case)` and `if(cond, true_case)` ) we currently use `if(_1, __2)`. This will not warn the user that she is passing too many arguments to if once she exceeds three. 

A novel way to express this use case or the ability to describe multiple map conditions is desired.

hkaiser:
- this fixes #63

### DIFF
--- a/phylanx/execution_tree/primitives/base_primitive.hpp
+++ b/phylanx/execution_tree/primitives/base_primitive.hpp
@@ -170,9 +170,7 @@ namespace phylanx { namespace execution_tree
     ///////////////////////////////////////////////////////////////////////////
     // Factory functions
     using factory_function_type =
-        hpx::util::function_nonser<
-            primitive(hpx::id_type, std::vector<primitive_argument_type>&&)
-        >;
+        primitive (*)(hpx::id_type, std::vector<primitive_argument_type>&&);
 
     template <typename Primitive>
     primitive create(hpx::id_type locality,
@@ -183,7 +181,7 @@ namespace phylanx { namespace execution_tree
 
     ///////////////////////////////////////////////////////////////////////////
     using match_pattern_type = std::pair<std::string, factory_function_type>;
-    using pattern_list = std::vector<match_pattern_type>;
+    using pattern_list = std::vector<std::vector<match_pattern_type>>;
 }}
 
 namespace phylanx { namespace execution_tree { namespace primitives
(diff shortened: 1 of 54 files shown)


aserio:
I am getting a weird error when testing your branch:

[ 97%] Building CXX object tests/unit/execution_tree/CMakeFiles/generate_tree_test_exe.dir/generate_tree.cpp.o
/home/aserio/git/phylanx/tests/unit/execution_tree/generate_tree.cpp: In function ‘void test_if_conditional()’:
/home/aserio/git/phylanx/tests/unit/execution_tree/generate_tree.cpp:212:5: error: could not convert ‘{phylanx::execution_tree::primitives::if_conditional::match_data}’ from ‘<brace-enclosed initializer list>’ to ‘phylanx::execution_tree::pattern_list {aka std::vector<std::vector<std::pair<std::__cxx11::basic_string<char>, phylanx::execution_tree::primitive (*)(hpx::naming::id_type, std::vector<mpark::variant<phylanx::ast::nil, bool, long int, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, phylanx::ir::node_data<double>, phylanx::execution_tree::primitive> >&&)> > >}’
     };
     ^
Do you think this is related to the changes on this branch?

hkaiser:
@aserio This is expected ;-) The changes require for you to expose a list of pattern matchers, instead of a single one. See for instance here for an example: https://github.com/STEllAR-GROUP/phylanx/blob/fixing_63/src/execution_tree/primitives/add_operation.cpp#L35-L38

aserio:
I got it to work!

phylanx/execution_tree/primitives/base_primitive.hpp
```diff
@@ -169,9 +169,7 @@ namespace phylanx { namespace execution_tree
     ///////////////////////////////////////////////////////////////////////////
     // Factory functions
     using factory_function_type =
-        hpx::util::function_nonser<
-            primitive(hpx::id_type, std::vector<primitive_argument_type>&&)
-        >;
+        primitive (*)(hpx::id_type, std::vector<primitive_argument_type>&&);
```
aserio: What does this change do?

hkaiser: It was a `function<>` object before and is a plain pointer to a (C-style) function now. This was a workaround for a problem in HPX which hopefully will be solved soon (see https://github.com/STEllAR-GROUP/hpx/pull/2945), at which point we can roll back this change.

PR #71, opened 2017-10-17
Migration to Blaze - CMake scripts updated to find/verify Blaze.

parsa:


cmake/Phylanx_SetupBlaze.cmake
```diff
@@ -0,0 +1,33 @@
...
+	  phylanx_error("Blaze could not be found. Please set blaze_DIR to help locating it.")
+	else()
+	  phylanx_info("Skipping Blaze presence check.")
+	endif()
+  endif()
+
+
+  phylanx_info("Blaze was found.")
+
+endmacro()
```
hkaiser: This file contains tabs, please convert those to spaces

parsa: Fixed.

src/CMakeLists.txt
```diff
@@ -98,9 +98,10 @@ set_target_properties(
     CLEAN_DIRECT_OUTPUT 1
     FOLDER "Core")
 
+# NOMINMAX is needed for Windows to circumvent issues with std::numeric_limits<T>
 set_property(TARGET phylanx_component APPEND
   PROPERTY COMPILE_DEFINITIONS
-  "PHYLANX_EXPORTS")
+  "PHYLANX_EXPORTS;NOMINMAX")
```
hkaiser: Please set this on Windows platforms only. Also, the comment is misleading.

parsa: How is the comment misleading? The code won't compile without it.

cmake/Phylanx_SetupBlaze.cmake
hkaiser: I tried this with a standalone Blaze installation (outside of vcpkg). I did specify the `blaze_INCLUDE_DIR` while running cmake. In this case cmake fails to find the library, but if I specify `PHYLANX_SKIP_BLAZE_CHECK` (what a strange way to circumvent error messages) the code still compiles.

parsa: It worked on my Mac, I'll check on my Windows machine.
> what a strange way to circumvent error messages
I don't see a way to have vcpkg tell me where to find its include directory so the check can proceed. Do you have a better idea?

src/CMakeLists.txt
hkaiser: I didn't say it would compile without NOMINMAX. I meant that the issue is not related to `std::numeric_limits<T>`. Th eproblem will occur whenever `min()` or `max()` are used in an unguarded way.

cmake/Phylanx_SetupBlaze.cmake
hkaiser: > It worked on my Mac, I'll check on my Windows machine. How did you set up your standalone Blaze installation? Can I see the errors?

It simply indicated that `HAVE_BLAZE_H` or similar was not defined.

> I don't see a way to have vcpkg tell me where to find its include directory so the check can proceed. Do you have a better idea?

Yes, I know - I created a ticket on their end (https://github.com/Microsoft/vcpkg/issues/1996).

src/CMakeLists.txt
```diff
@@ -98,10 +98,15 @@ set_target_properties(
     CLEAN_DIRECT_OUTPUT 1
     FOLDER "Core")
 
-# NOMINMAX is needed for Windows to circumvent issues with std::numeric_limits<T>
```
parsa: @hkaiser is this better or is it still misleading?

src/CMakeLists.txt
```diff
@@ -98,10 +98,15 @@ set_target_properties(
...
 set_property(TARGET phylanx_component APPEND
   PROPERTY COMPILE_DEFINITIONS
-  "PHYLANX_EXPORTS;NOMINMAX")
+  "PHYLANX_EXPORTS")
+
+if(WIN32)# NOMINMAX is needed to resolve a conflict between Windows.h and std::numeric_limits
```
parsa: @hkaiser is this better or is it still misleading?

parsa:
Note: These CI builds failing have nothing to do with the code in this PR.

hapoo:
Worked on my machine

hkaiser:
I will update the appveyor base files asap. That might make it work there. Why is CircleCI failing?

parsa:
It says `-- Blaze was found.` but also says `include could not find load file: /blaze-src/cmake/blaze-targets.cmake`

I suppose it means Blaze wasn't installed through CMake. That's fine, I'll try using `blaze_INCLUDE_DIR`

parsa:
@hkaiser I don't have a solution for that failure. The scripts are working as they supposed to. The Docker image's Blaze installation needs to be fixed.

hkaiser:
I have added blaze to the vcpkg modules on appveyor now. Unfortunately this does not work still.

hkaiser:
Your latest change seem to have fixed it for AppVeyor (the current failures are unrelated), great!

PR #72, opened 2017-10-17
Fixing dot-product, adding test

hkaiser:
- this also fixes the LRA example

@justwagle this fixes the LRA example for me. Could you please verify it fixes things for you as well?


justwagle:
@hkaiser Yes, this works for me. Thank you very much.

PR #73, opened 2017-10-18
Simple for loop primitive

justwagle:
This pull request tries to create a simple for loop primitive that could be extended upon in the future.
At the moment, the for loop has the following: 
for( initialization, condition_to_check , block_to_execute).


hkaiser:
> At the moment, the for loop has the following: for( initialization, condition_to_check , block_to_execute).

@justwagle In order for this to be similar to the C++ `for(init; cond; reinit) body;` - wouldn't you need to add a fourth argument?

justwagle:
@hkaiser  Yes, I will add the reinint.

src/execution_tree/primitives/for_operation.cpp
```diff
@@ -0,0 +1,142 @@
...
+                    // evaluate body of for statement
+                    auto this_ = this->shared_from_this();
+                    return literal_operand(operands_[3]).then(
+                        [this_](
+                            hpx::future<primitive_result_type> && result
+                        ) mutable
+                        {
+                            this_->result_ = result.get();
+                            //do the reinit statement
+                            auto hold = this_->reinit().get();
```
hkaiser: Wouldn't it be better to evaluate the reinit asynchronously as well?

justwagle: @hkaiser  Yes, that needs fixing. I will fix it.

justwagle:
@hkaiser CircleCI is finally green for this for_loop. The while test also passes now.

PR #74, opened 2017-10-18
Adding if2

aserio:
Adding the if primitive.

### DIFF
--- a/src/execution_tree/primitives/if_conditional.cpp
+++ b/src/execution_tree/primitives/if_conditional.cpp
@@ -72,25 +72,47 @@ namespace execution_tree {
         }
 
     ///////////////////////////////////////////////////////////////////////////
+    // struct containing implementation of evaluate 
+    // (used to resolve lifetime issues)
+    namespace detail
+    {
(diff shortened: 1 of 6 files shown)


aserio:
To-do:
[] Create a test in generate_tree.cpp which checks the case where if(cond, case) returns `nil` when case = false

src/execution_tree/primitives/if_conditional.cpp
```diff
@@ -0,0 +1,97 @@
...
+                    "given "
+                    "by the operands array is valid");
+            }
+        }
+
+    ///////////////////////////////////////////////////////////////////////////
+    // evaluate 'true_case' or 'false_case' based on 'cond'
+        hpx::future<primitive_result_type> if_conditional::eval() const
+        {
+            hpx::future<std::uint8_t> cond_eval = boolean_operand(operands_[0]);
```
hkaiser: I think this should be protected from lifetime issues. if the primitive goes out of scope while the evaluation is being performed you'd see a segfault.

aserio: The latest commit should solve the issue

Issue #83, opened 2017-10-27
Matrix dimensions are used inconsistent in API


hkaiser:
The dimensions used/returned from `node_data` have inconsistent ordering (partially `[rows, columns]`, partially `[column, rows]`). This is confusing and error prone and needs fixing.

parsa:
Addressed in #84

PR #84, opened 2017-10-27
Fix dot product

parsa:
- 2D dot product fixed
- Row/Column inconsistencies fixed.
- Eigen removed from Dockerfile.
- `BLAZE_INCLUDE_DIR` and Vcpkg hacks removed from FindBlaze.cmake


hkaiser:
This does not fully address #83, see for instance here: https://github.com/STEllAR-GROUP/phylanx/blob/master/phylanx/ir/node_data.hpp#L225 or here https://github.com/STEllAR-GROUP/phylanx/blob/master/phylanx/ir/node_data.hpp#L221 where `column` and `row` are flipped.

parsa:
This PR combined too many separate issues into one PR.

Issue #79, opened 2017-10-25
Add BLAS support


parsa:
BLAS is required by some Blaze functions we use in Phylanx. e.g. [matrix inversion](https://bitbucket.org/blaze-lib/blaze/wiki/Matrix%20Operations#!matrix-inversion) and [matrix determinant](https://bitbucket.org/blaze-lib/blaze/wiki/Matrix%20Operations#!det)

Needs CMake support.

ct-clmsn:
@parsa @hkaiser would the existing [HPX blas infrastructure](https://github.com/STEllAR-GROUP/hpxla) be worth revisiting?

hkaiser:
@ct-clmsn interesting idea. Please understand however that this was never a) verified to properly work, nor b) never optimized for performance. But as a fallback this might be something to look into. Thanks for reminding me of this work - I had completely forgotten that it existed.

ct-clmsn:
@hkaiser could blaze (or an hpx-integrated blaze) fit into it as a matrix-template type?

hkaiser:
This has been fixed by merging #88.

PR #88, opened 2017-10-28
Remove Blaze CMake hacks, Add BLAS/LAPACK Support

parsa:
* Remove `blaze_INCLUDE_DIR` flag
* Microsoft/vcpkg#1996 is resolved 
    * Vcpkg properly installs Blaze through CMake 
    * Remove the use of undocumented variables to handle Blaze installations available through Vcpkg
* Add BLAS/LAPACK support (#79)


tests/unit/execution_tree/primitives/inverse_operation.cpp
```diff
@@ -29,7 +29,7 @@ void test_transpose_operation_0d()
         transpose.eval();
```
hkaiser: Why's the inverse test actually testing transpose?

parsa: @hkaiser :)))) Who wrote this test? It does test transpose and functionality is correct but most of the method and variable names are wrong

parsa: Renamed in 42b1a0af2076deefdf48d22333442bc92de4045f

hkaiser: Thanks!

phylanx/config.hpp
```diff
@@ -11,8 +11,30 @@
...
 #define NOMINMAX
-#endif
+#endif // NOMINMAX
+
+// Use HPX threads. Do not use C++11 threads
+#if !defined(BLAZE_USE_HPX_THREADS)
+#define BLAZE_USE_HPX_THREADS 1
+
+#if !defined(BLAZE_USE_SHARED_MEMORY_PARALLELIZATION) || BLAZE_USE_SHARED_MEMORY_PARALLELIZATION != 1
+#define BLAZE_USE_SHARED_MEMORY_PARALLELIZATION 1
```
hkaiser: This will cause problems if `BLAZE_USE_SHARED_MEMORY_PARALLELIZATION == 0` or similar.

phylanx/config.hpp
```diff
@@ -11,8 +11,30 @@
...
+#endif BLAZE_USE_SHARED_MEMORY_PARALLELIZATION
 
+#ifdef BLAZE_USE_CPP_THREADS
+#undef BLAZE_USE_CPP_THREADS
+#endif // BLAZE_USE_CPP_THREADS
+
+#ifdef BLAZE_USE_BOOST_THREADS
```
hkaiser: No need to guard the `#undef`. It's perfectly valid to in-define a non-defined variable.

phylanx/config.hpp
parsa: If `BLAZE_USE_SHARED_MEMORY_PARALLELIZATION ` is not 1 then it may not use HPX. Why is being equal to 0 a problem?

phylanx/config.hpp
parsa: Guards removed.

phylanx/config.hpp
hkaiser: if  `BLAZE_USE_SHARED_MEMORY_PARALLELIZATION` is defined but `!= 1` you will see macro redefinition errors from the preprocessor.

PR #90, opened 2017-10-29
Slicing Primitive

justwagle:
This pull request creates a slicing primitive:

**result slicing(input, row_start, row_stop, col_start , col_stop)**

Here, input can be a **0d, 1d or 2d** data. 

*If the input is a 0d data and slicing is called, the input itself is returned as the result. The arguments row_start, row_stop, col_start, col_stop are ignored internally.

*If the input is a 1d data and slicing is called, the part of the vector is  returned as the result. The arguments col_start and col_stop determine the result. The arguments row_start, row_stop are set to 0 and 1 respectively internally and any values provided by the user is ignored.

*If the input is a 2d data and slicing is called, part of the matrix is returned as the result. The arguments row_start, row_stop, col_start, col_stop determine the result.


justwagle:
@hkaiser : I realized I made a mistake while translating the index from Phalanx to blaze and have fixed it in the following commit. Thanks.

PR #92, opened 2017-10-30
sequential transforms

ct-clmsn:
Removed unnecessary comment.


ct-clmsn:
first (draft) version of transducer w/state transition tables.

ct-clmsn:
working on a more reasonable solution.

Issue #44, opened 2017-10-04
Phylanx tests fail with JEMalloc 5.0


khuck:
When HPX and Phylanx are built with JEMalloc 5.0, three of the tests fail:

```
11/15 Test #11: tests.unit.python.ast.generate_ast ........***Failed    0.11 sec
Couldn't load _phylanx(d) extension module:  /usr/local/jemalloc/5.0.1/lib/libjemalloc.so.2: cannot allocate memory in static TLS block
sys.path:  ['/var/lib/buildbot/slaves/phylanx/ktau-debug/build/tests/unit/python/ast', '/var/lib/buildbot/slaves/phylanx/ktau-debug/build/buildbot/build-ktau-x86_64-Linux-gcc/phylanx-Debug/python/build/lib.linux-x86_64-3.5', '/usr/lib/python35.zip', '/usr/lib/python3.5', '/usr/lib/python3.5/plat-x86_64-linux-gnu', '/usr/lib/python3.5/lib-dynload', '/usr/local/lib/python3.5/dist-packages', '/usr/lib/python3/dist-packages', '/var/lib/buildbot/slaves/phylanx/ktau-debug/build/buildbot/build-ktau-x86_64-Linux-gcc/phylanx-Debug/python/build/lib.linux-x86_64-3.5/phylanx']
os.environ['PATH']:  /var/lib/buildbot/slaves/phylanx/ktau-debug/build/buildbot/build-ktau-x86_64-Linux-gcc/hpx-Debug/lib:/var/lib/buildbot/slaves/phylanx/ktau-debug/build/buildbot/build-ktau-x86_64-Linux-gcc/phylanx-Debug/lib:/usr/local/bin:/usr/bin:/bin:/usr/local/games:/usr/games:/snap/bin:/usr/local/java/jdk1.8.0_11/bin
phylanx_bindir:  /var/lib/buildbot/slaves/phylanx/ktau-debug/build/buildbot/build-ktau-x86_64-Linux-gcc/phylanx-Debug/lib
hpx_bindir:  /var/lib/buildbot/slaves/phylanx/ktau-debug/build/buildbot/build-ktau-x86_64-Linux-gcc/hpx-Debug/lib

      Start 12: tests.unit.python.ast.node
12/15 Test #12: tests.unit.python.ast.node ................***Failed    0.09 sec
[...]

PR #95, opened 2017-11-02
Warning added if HPX is configured with JEMALLOC

monil01:
This PR addresses #44.


hkaiser:
Is there a way to check the version of jemalloc used? ... just asking - there might be no way to figure that out from what HPX exposes through its `HPXConfig.cmake`.

hkaiser:
@monil01 do you plan to address @khuck's  comment above?

monil01:
@hkaiser Yes I will do this today.

monil01:
@hkaiser I addressed Kevin's comment and pushed it.

hkaiser:
@khuck are you now fine with this PR?

khuck:
Yes, looks good

PR #99, opened 2017-11-02
Cross product primitive implemented

parsa:
Part of issue #2

### DIFF
--- a/tests/unit/execution_tree/primitives/cross_operation.cpp
+++ b/tests/unit/execution_tree/primitives/cross_operation.cpp
@@ -50,7 +50,7 @@ void test_one_vector_with_dimension_2()
     blaze::DynamicVector<double, blaze::rowVector> v1{1.0, 2.0};
     blaze::DynamicVector<double, blaze::rowVector> v2{4.0, 5.0, 6.0};
 
-    blaze::DynamicMatrix<double> expected{ {12.0, -6.0, -3.0} };
+    blaze::DynamicMatrix<double> expected{{12.0, -6.0, -3.0}};
 
     phylanx::execution_tree::primitive lhs =
         hpx::new_<phylanx::execution_tree::primitives::variable>(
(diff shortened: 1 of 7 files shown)


hkaiser:
The NumPy docs say (see https://docs.scipy.org/doc/numpy-1.13.0/reference/generated/numpy.cross.html):

> The cross product of a and b in R^3 is a vector perpendicular to both a and b. If a and b are arrays of vectors, the vectors are defined by the last axis of a and b by default, and these axes can have dimensions 2 or 3. Where the dimension of either a or b is 2, the third component of the input vector is assumed to be zero and the cross product calculated accordingly. In cases where both input vectors have dimension 2, the z-component of the cross product is returned.

IOW, we should add support for a cross product of 2 dimensional arrays as well.

parsa:
After experimenting with NumPy a bit I found these differences with my current implementation that I have to address:
- Vectors must have exactly 2 or 3 elements (current code accepts vectors with fewer than 3 elements)
- It's possible to perform cross product on a vector and a matrix (Not supported)
- The matrices can have 2 columns, too

tests/unit/execution_tree/primitives/cross_operation.cpp
```diff
@@ -0,0 +1,81 @@
...
+    blaze::DynamicMatrix<double> expected{{-3.0, 0.0, 0.0}};
+}
+
+void test_multiple_vector_cross_products()
+{
+    blaze::DynamicMatrix<double> v1{{1.0, 2.0, 3.0}, {4.0, 5.0, 6.0}};
+    blaze::DynamicMatrix<double> v2{{4.0, 5.0, 6.0}, {1.0, 2.0, 3.0}};
+
+    blaze::DynamicMatrix<double> expected{{-3.0, 6.0, -3.0}, {3.0, -6.0, 3.0}};
+}
```
hkaiser: These three tests above don't look complete.

PR #100, opened 2017-11-02
Square root and Power primitives added

parsa:
Part of issue #2

### DIFF
--- a/src/execution_tree/primitives/power_operation.cpp
+++ b/src/execution_tree/primitives/power_operation.cpp
@@ -1,0 +1,139 @@
+//  Copyright (c) 2017 Parsa Amini
+//
+//  Distributed under the Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+#include <phylanx/config.hpp>
+#include <phylanx/execution_tree/primitives/power_operation.hpp>
+#include <phylanx/ir/node_data.hpp>
+#include <phylanx/util/serialization/blaze.hpp>
+
+#include <hpx/include/components.hpp>
+#include <hpx/include/lcos.hpp>
+#include <hpx/include/util.hpp>
+
+#include <cstddef>
+#include <memory>
+#include <utility>
+#include <vector>
+#include <cmath>
+
+///////////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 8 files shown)


src/execution_tree/primitives/power_operation.cpp
```diff
@@ -0,0 +1,139 @@
...
+            using operand_type = ir::node_data<double>;
+            using operands_type = std::vector<operand_type>;
+
+            primitive_result_type power0d(operands_type && ops) const
+            {
+                ops[0][0] = std::pow(ops[0][0], ops[1][0]);
+                return std::move(ops[0]);
+            }
+
+            // lhs_num_dims == 1
```
hkaiser: This comment is misleading

src/execution_tree/primitives/power_operation.cpp
```diff
@@ -0,0 +1,139 @@
...
+            primitive_result_type powerxd(operands_type && ops) const
+            {
+                operand_type& lhs = ops[0];
+                operand_type& rhs = ops[1];
+
+                lhs.matrix() = blaze::pow(lhs.matrix(), rhs[0]);
```
hkaiser: Should we add an exception for the case if the rhs is not a scalar?

src/execution_tree/primitives/square_root_operation.cpp
```diff
@@ -0,0 +1,133 @@
...
+
+    ///////////////////////////////////////////////////////////////////////////////
+    namespace phylanx {
+    namespace execution_tree {
+        namespace primitives
+        {
+            ///////////////////////////////////////////////////////////////////////////
+            std::vector<match_pattern_type> const square_root_operation::match_data =
+            {
+                hpx::util::make_tuple("square_root", "square_root(_1, _2)", &create<square_root_operation>)
```
hkaiser: Please avoid long lines.

src/execution_tree/primitives/power_operation.cpp
parsa: No. `rhs` is guaranteed to be a scalar when the code reaches here. See `eval()` (line 97)

src/execution_tree/primitives/power_operation.cpp
parsa: Removed.

PR #113, opened 2017-12-01
Add LRA with CSV

justwagle:
This pull request adds another example for LRA where data is read from csv file rather than being hard coded. This pull request also modifies file_read_csv primitive so that it skips any header information in the csv file. Also, slicing now returns a vector if only a single row or column is sliced.


examples/algorithms/lra_csv.cpp
```diff
@@ -0,0 +1,122 @@
...
+           "number of iterations (default: 750)")
+          ("alpha,a",
+           boost::program_options::value<double>()->default_value(1e-5),
+           "alpha (default: 1e-5)")
+          ("data_csv", boost::program_options::value<std::string>(),
+           "file name for reading data")
+          ;
+
+  return hpx::init(desc, argc, argv);
+}
```
hkaiser: Not sure, but this indentation looks messed up. Do you use 4 spaces to indent code?

hkaiser: Could you add the actual CSV file as well? Also, please add a comment to this example explaining where to get the data from and how it is supposed to be formatted.

ct-clmsn:
@justwagle @hkaiser have ya'll done a trivial comparative analysis of the python version and the phylanx version?

justwagle:
@ct-clmsn I do not have the data for that at the moment, however, I will be doing some comparisons over the Weekend/Monday.

hkaiser:
> have ya'll done a trivial comparative analysis of the python version and the phylanx version? 

@ct-clmsn we have not done a thorough performance comparison yet. We have however verified that our implementation provides the same answers. In general I'd expect our implementation to run faster than the Python base version as a) it's 100% C++, and b) we can run things parallelized.

ct-clmsn:
@justwagle would a longer term solution be worth discussing with @khuck? @hkaiser not a problem. curiosity motivated the question. I've similar expectations.

khuck:
@ct-clmsn Doing an explicit performance comparison is one of the motivations for what Monil is looking into.  However, he was looking at measuring TensorFlow and Spartan - do we want to compare the Phylanx solution with a vanilla Python implementation, too?

hkaiser:
We will benefit most from using APEX once we have figured out how to best make it aware of Phylanx. Currently APEX works on the level of HPX tasks, I strongly believe however that raising the abstraction level here will give us better insights about Phylanx itself.

ct-clmsn:
@khuck good deal!

ct-clmsn:
@hkaiser understood. thanks!

PR #115, opened 2017-12-04
Fix linux startup crash

### LINKED ISSUE #116: Release build crash when phylanx imported into Python
The Release and RelWithDebInfo builds of Phylanx crash when the phylanx module is imported into python.

This was caused by `__argc` and `__argv` not getting set correctly before passing them to HPX.

khuck:
Fixing initialization bug when ELF parsing fails
When getting command line options from somewhere other than main,
Linux requires us to parse /proc/self/cmdline because using the
built in ELF support fails - python is forking the process at some
point (so GDB tells me) which means that the ELF .init_array
support doesn't get called.

I copied the similar windows support and changed the line that splits the string.

This fixes #116


khuck:
By the way, my fix worked before I pulled the latest code, now there are other crashes. :(

python/src/init_hpx.cpp
```diff
@@ -16,17 +16,59 @@
...
+    {
+        // parse the command line proc file
+        std::ifstream cmdline("/proc/self/cmdline");
+        std::string tmpstr((std::istreambuf_iterator<char>(cmdline)),
+                         std::istreambuf_iterator<char>());
+        // make a stringstream and split into a vector of strings
+        std::stringstream ss(tmpstr);
+        std::istream_iterator<std::string> begin(ss);
+        std::istream_iterator<std::string> end;
+        std::vector<std::string> args_(begin, end);
```
hkaiser: Wouldn't it be possible to directly initialize the vector from the ifstream:

    std::ifstream cmdline("/proc/self/cmdline"); 
    if (!cmdline.is_open())
    {
        // proc fs is not available, raise error etc.
    }
    std::vector<std::string> args{
        std::istream_iterator<char>{cmdline},  
        std::istream_iterator<char>{}; 
    };

?

hkaiser:
@khuck thanks for catching this!

python/src/init_hpx.cpp
khuck: Yes, that is easier. :)

But I had to change the {} to () to get past the "most vexing parse" problem...

khuck:
@hkaiser I cleaned up the patch, as suggested.  Thanks!

hkaiser:
> By the way, my fix worked before I pulled the latest code, now there are other crashes. :(

@khuck AppVeyor and CircleCI are green. Could you provide a link to the failures you're seeing, please?

khuck:
I had to pull and rebuild HPX, now everything is working.  Thanks!

PR #118, opened 2017-12-05
Use full dataset

justwagle:
This pull request makes sure LRA uses the full dataset and also adds a high resolution timer to time the execution.

### DIFF
--- a/examples/algorithms/lra_csv.cpp
+++ b/examples/algorithms/lra_csv.cpp
@@ -29,7 +29,7 @@ char const* const read_x_code = R"(block(
     // Read X-data from given CSV file
     //
     define(read_x, filepath,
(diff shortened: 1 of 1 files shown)


ct-clmsn:
@justwagle @hkaiser would a timer primitive be worth providing?

hkaiser:
@ct-clmsn if you have a simple timer primitive in mind - what would you like to see in addition to `hpx::util::high_resolution_timer` (see here: https://github.com/STEllAR-GROUP/hpx/blob/master/hpx/util/high_resolution_timer.hpp)?

ct-clmsn:
@hkaiser @justwagle high_res, interval, or the time logger? I'm not completely sold on idea; seems like a temporary solution that would be superceded when the longer term solution is available.

Issue #128, opened 2017-12-14
Extend `slice()` primitives to support negative indicies


hkaiser:
This should allow for accessing columns/rows relative to the end of the array instead of relative to the beginning.

hkaiser:
This has been fixed by merging #133

PR #133, opened 2017-12-15
Extending `slice()` primitives to support negative indicies

justwagle:
This pull request tries to address #128 by adding support for negative indices in the slice primitive. 

***slice ( matrix, row_start , row_stop, col_start , col_stop)*** now allows ***row_start*** to have negative values which effectively slices the matrix ***row_start*** rows from the end. 
similarly, negative ***col_start*** effectively slices the matrix ***col_start*** columns from the end. 
The parameters ***row_stop / col_stop*** does not have any effect when using negative indices to slice from the end. Any values passed to these parameters are simply ignored. 

This pull request also modifies the lra_csv example to remove hard coded values.

### DIFF
--- a/src/execution_tree/primitives/column_slicing.cpp
+++ b/src/execution_tree/primitives/column_slicing.cpp
@@ -93,6 +93,26 @@ namespace phylanx { namespace execution_tree { namespace primitives
                 // column = col_start
                 // n = (col_stop - col_start)
 
+                if (col_start < 0 && col_stop >= 0)    //slice from the end
+                {
+                    auto size = args[0].vector().size();
+
+                    subvector_type sv = blaze::subvector(
+                            args[0].vector(), size + col_start, (-col_start));
+
+                    return ir::node_data<double>{vector_type{std::move(sv)}};
+                }
+
+                if (col_start < 0 && col_stop < 0)    //slice from the end
+                {
+                    auto size = args[0].vector().size();
+
+                    subvector_type sv = blaze::subvector(args[0].vector(),
+                                                         size + col_start, (-col_start) + col_stop);
+
+                    return ir::node_data<double>{vector_type{std::move(sv)}};
+                }
+
                 subvector_type sv =
                     blaze::subvector(args[0].vector(),
                         col_start, (col_stop - col_start));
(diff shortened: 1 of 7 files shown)


hkaiser:
> The parameters row_stop / col_stop does not have any effect when using negative indices to slice from the end. Any values passed to these parameters are simply ignored.

What is if the user wants to extract the two columns before the last one, shouldn't `slice(a, ..., -3, -1)`be the right incantation for that use case?

Also, would it be possible to add the same functionality to the `row_slice` and `column_slice` primitives as well?

justwagle:
> What is if the user wants to extract the two columns before the last one, shouldn't slice(a, ..., -3, -1)be the right incantation for that use case?

> Also, would it be possible to add the same functionality to the row_slice and column_slice primitives as well?

@hkaiser , Yes, I will add these. They use cases completely slipped my mind.

src/execution_tree/primitives/column_slicing.cpp
```diff
@@ -93,6 +93,26 @@ namespace phylanx { namespace execution_tree { namespace primitives
                 // column = col_start
                 // n = (col_stop - col_start)
 
+                if (col_start < 0 && col_stop >= 0)    //slice from the end
```
hkaiser: What is the intended semantic for `start < 0 && stop => 0`? I'd have expected for this to be an error.

justwagle: I was thinking for `col_start < 0 && col_stop >=0`  if a user wants say get the last two column, he can simply say column_slice(a, -2, x) , here, x is any value greater than or equal to zero, that is any positive value  can be passed for the argument col_stop.

hkaiser: Do you mean that if `start < 0 && end >= 0`, then the `end` specifies the number of columns instead of the index of the first column after the last?

justwagle: if `start < && end >= 0` , whatever end specifies is ignored internally and the value passed to start is used to extract the number of columns. 
For example,
if the user wants to extract the last 3 columns from the input array, he would say 
`column_slice(a, -3, some_positive_value_ignored_internally)`.
But if the user wants to extract two columns before the last one, 
`column_slice(a, -3, -1)`

hkaiser: Ok. May I suggest not to implement this API an such an inconsistent way? This seems to be very confusing and does not follow the 'rule of the least surprise'.

justwagle: Yes sir, how about I throw an error if `start <0 and stop >0` ? If the user wants to extract the last two columns he will have to say `column_slice(a, -2, 0)`  and if the user wants to extract two columns before the last one, `column_slice(a, -3,-1)`. Will that be an acceptable solution ?

hkaiser: That sounds about right. Thanks!

hkaiser:
@justwagle could you rebase your changes on top of master, please? That should fix the test failures.

PR #142, opened 2018-01-10
Python interface dev

stevenrbrandt:
Work by Rod and myself to make the Python3 interface work. This PR adds many pybind11 bindings.


ct-clmsn:
@stevenrbrandt you and Rod have done an impressive amount of work!

examples/phyast.py
```diff
@@ -0,0 +1,390 @@
...
+        self.f = f
+        # Get the source code
+        src = inspect.getsource(f) #getsource(f)
+        # Before recompiling the code, take
+        # off the decorator from the source.
+        src = re.sub(r'^\s*@\w+\s*','',src)
+
+        # Create the AST
+        tree = ast.parse(src)
+        #astdump.indented(tree)
```
hkaiser: Please remove code that is commented out

examples/phyast.py
```diff
@@ -0,0 +1,390 @@
...
+        #return self.nf(*args)
+        #print('len=',len(args))
+        if len(args)==1:
+            return et.phylisp_eval(self.new_src,args[0])
+        elif len(args)==2:
+            return et.phylisp_eval(self.new_src,args[0],args[1])
+        else:
+            return et.phylisp_eval(self.new_src,*args)
+
+@phylanx
```
hkaiser: Could you please separate the implementation of the decorator from the tests?

python/src/phylanx.cpp
```diff
@@ -164,12 +170,59 @@ namespace phylanx { namespace bindings
...
+                return primitive{
+                    hpx::local_new<primitives::variable>(node_result)};
+            }
+            else if (ndx == 2)
+            {
+                auto node_result = phylanx::util::get<std::int64_t>(result);
+                return primitive{
+                    hpx::local_new<primitives::variable>(node_result)};
+            }
+        }
```
hkaiser: What if `ndx != 4 && ndx != 2`

python/src/phylanx.cpp
```diff
@@ -497,36 +535,390 @@ PYBIND11_MODULE(_phylanx, m)
...
+            return hpx::threads::run_as_hpx_thread([&]() {
+                using namespace phylanx::execution_tree;
+                return primitive{hpx::local_new<primitives::variable>(
+                    phylanx::ir::node_data<double>{d})};
+            });
         },
         "create a new variable from a floating point value");
 
+    execution_tree.def("linspace",
+        [](double xmin, double xmax, std::size_t nx) {
```
hkaiser: What does 'linspace' stand for?

python/src/phylanx.cpp
```diff
@@ -497,36 +535,390 @@ PYBIND11_MODULE(_phylanx, m)
...
+            return hpx::threads::run_as_hpx_thread([&]() {
+                using namespace phylanx::execution_tree;
+                return primitive{hpx::local_new<primitives::variable>(
+                    phylanx::ir::node_data<double>{m})};
+            });
+        },
+        "create a new variable from a matrix of zeros");
+
+    execution_tree.def(
+        "phylisp_eval", expression_compiler<>, expression_compiler_help);
```
hkaiser: We shouldn't call the language `phylisp`. Why I see where this comes from, I believe it's not a 'lisp-ish' language.

python/src/phylanx.cpp
```diff
@@ -497,36 +535,390 @@ PYBIND11_MODULE(_phylanx, m)
...
+        expression_compiler<double, phylanx::execution_tree::primitive const&,
+            double>,
+        expression_compiler_help);
+    execution_tree.def("phylisp_eval",
+        expression_compiler<phylanx::execution_tree::primitive const&, double,
+            double>,
+        expression_compiler_help);
+    execution_tree.def("phylisp_eval",
+        expression_compiler<double, double, double>,
+        expression_compiler_help);
```
hkaiser: Wouldn't it be sufficient defining overload of this function taking one or more instances of `primitive_argument_type`'s? Otherwise we'll get an exponential explosion of overloads for all possible combination of argument types.

python/src/phylanx.cpp
```diff
@@ -497,36 +535,390 @@ PYBIND11_MODULE(_phylanx, m)
...
+
+    execution_tree.def("add",
+        [](phylanx::execution_tree::primitive const& obj1,
```
hkaiser: Same here, using `primitive_argument_type` as the parameter types might be better.

python/src/phylanx.cpp
```diff
@@ -164,12 +170,59 @@ namespace phylanx { namespace bindings
...
+const char* expression_compiler_help =
+    "compile and evaluate a numerical expression in Phylanx lisp";
+template <typename... Ts>
+typename hpx::util::invoke_result<phylanx::execution_tree::primitive(Ts...),
+    Ts...>::type
+expression_compiler(std::string xexpr, Ts const&... ts)
+{
+    return hpx::threads::run_as_hpx_thread([&]() {
+        using namespace phylanx::execution_tree;
+        auto here = hpx::find_here();
```
hkaiser: `here` is not used anywhere, afaics.

python/src/phylanx.cpp
```diff
@@ -164,12 +170,59 @@ namespace phylanx { namespace bindings
...
+                auto node_result = phylanx::util::get<std::int64_t>(result);
+                return primitive{
+                    hpx::local_new<primitives::variable>(node_result)};
+            }
+        }
+        catch (const std::exception& ex)
+        {
+            PyErr_SetString(PyExc_RuntimeError, ex.what());
+        }
+        catch (...)
```
hkaiser: You might want to consider catching `hpx:exception`'s as well.

tests/unit/python/ast/builds_ast_ndarray.py
```diff
@@ -0,0 +1,57 @@
...
+b = ndarray(2)
+expr = a + b
+
+expr = a + expr
+print(expr)
+print(a)
+print(b)
+
+expr2 = a * b
+print(expr2)
```
hkaiser: This test does not 'test' anything. There is no way to tell (in a scripted environment) whether the code is doing what it is supposed to do. This code is more of an example demonstrating capabilities.

tests/unit/python/execution_tree/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
...
+
+  add_phylanx_python_unit_test("execution_tree" ${test}
+    SCRIPT ${script}
+    FOLDER "Tests/Python/Unit/Execution_Tree"
+    DEPENDS phylanx_py python_setup
+    WORKING_DIRECTORY ${PHYLANX_PYTHON_EXTENSION_LOCATION}
+    ENVIRONMENT "PYTHONPATH=${PHYLANX_PYTHON_EXTENSION_LOCATION}")
+
+  add_phylanx_pseudo_target(tests.unit.python_execution_treee.${test}_py)
+  #add_phylanx_pseudo_dependencies(tests.unit.python_execution_tree tests.unit.python_execution_tree.${test}_py)
```
hkaiser: If these pseudo-dependencies are not in place the test will not be executed by 'make tests'

examples/phyast.py
```diff
@@ -0,0 +1,390 @@
+import phylanx
+et = phylanx.execution_tree
+import inspect
+import ast
+import re
+
+def phy_print(m):
```
hkaiser: This function is duplicated in three files, should it be separated?

examples/phyast.py
hkaiser: Also, the decorator should probably go into the phylanx/python directory for everybody to use.

python/src/phylanx.cpp
```diff
@@ -1,29 +1,33 @@
...
+#include <hpx/runtime/threads/run_as_hpx_thread.hpp>
 
 #include <map>
-#include <string>
 #include <sstream>
+#include <string>
 #include <vector>
 
+#include <phylanx/execution_tree/primitives/file_read_csv.hpp>
+#include <phylanx/execution_tree/primitives/slicing_operation.hpp>
```
hkaiser: If we decide to expose the primitive components as python objects we need to add all of the related #includes. It might even be a good idea to create a separate file just for those bindings to avoid for this file to become too large.

hkaiser:
Greate work! Thanks!

Some general notes:

- Please add copyright and licenses to all files.
- Please don't reformat complete files (i.e. leave untouched code segments unchanged).
- What indentation depth will we use for our python codes, 2 or 4 spaces? Whatever we choose, it should be applied globally.

What is the purpose of exposing various primitives (like `add`) directly to the python code? What benefit do we get from this?

hkaiser:
May I suggest to separate the proposed changes into at least two pull requests? Leave the current one for the changes needed for the decorator and separate the changes proposed for exposing the primitive components to python into a different PR? I feel that these two changesets will require separate and independent discussions.

python/src/phylanx.cpp
stevenrbrandt: linspace() is a replication of a function in numpy that creates an array with nx values ranging from xmin to xmax.

python/src/phylanx.cpp
stevenrbrandt: hpx::exception subclasses std::exception, so it seems unnecessary.

python/src/phylanx.cpp
stevenrbrandt: Well, obviously, it will just return 0. :) Only numeric types are supported so far.

python/src/phylanx.cpp
stevenrbrandt: Good point.

python/src/phylanx.cpp
stevenrbrandt: Probably. Is that a blocker?

examples/phyast.py
stevenrbrandt: That should be done now.

examples/phyast.py
stevenrbrandt: It should only exist once now.

examples/phyast.py
stevenrbrandt: handled.

python/src/phylanx.cpp
stevenrbrandt: Removed.

python/src/phylanx.cpp
stevenrbrandt: Do you have a suggested name?

python/src/phylanx.cpp
stevenrbrandt: I don't see a way to do the autoconversion and it's a pain to do et.phylisp_eval("3") instead of just 3. Not sure what the right answer is here.

stevenrbrandt: Rereading your comment... So if I use primitive_argument_type, I need a way to create a python object of that type. Currently I don't have one and Python doesn't autoconvert double or anything else to that type.

tests/unit/python/ast/builds_ast_ndarray.py
stevenrbrandt: This wasn't part of of the commit AFAIK. Maybe Rod added it?

tests/unit/python/execution_tree/CMakeLists.txt
stevenrbrandt: But they are executed by "make tests." I checked.

python/src/phylanx.cpp
hkaiser: Just `eval` should be fine in this context. 

Wrt the name for our little language - I heard somebody calling it `PhySL`, but I'm really impartial and would like to hear other oppinions.

python/src/phylanx.cpp
hkaiser: `primitive_argument_type` should be constructable from `double` - so I assumed the binding would figure it out. Let me experiment a bit before getting back to this.

tests/unit/python/ast/builds_ast_ndarray.py
hkaiser: Shrug. It is part of this PR.

python/src/phylanx.cpp
hkaiser: `#including` all needed files is probably a good idea, if for nothing else than readability and robustness. Experience shows that relying on implicit availability of types makes the code base very brittle to unrelated changes (i.e. somebody removing an #include from a different header file might break the compilation of this source).

Separating things into a different source file is probably not a blocker, however as commented on earlier - I would like to see the two (independent) changesets that are currently proposed by this PR to be separated into independent pull requests.

python/src/phylanx.cpp
hkaiser: If that's the case the code should throw an exception if non-numeric types happen to be returned. Otherwise somebody will have a hard time figuring out what's wrong further down the road.

python/src/phylanx.cpp
hkaiser: `hpx::exception` might give you more error context information than what `std:;exception::what` returns. But sure - let's leave it as is for now.

tests/unit/python/execution_tree/CMakeLists.txt
hkaiser: How should that happen? AFAICS, without those there is no code in the cmake build system making the `tests` target depend on the python unit test added.

python/src/phylanx.cpp
```diff
@@ -551,7 +551,24 @@ PYBIND11_MODULE(_phylanx, m)
...
+        [](const std::vector<std::vector<double>>& d) {
+            std::size_t nx = d.size();
+            std::size_t ny = d[0].size();
+            blaze::DynamicMatrix<double> m{nx, ny};
+            for(int i=0;i<nx;i++) {
+              for(int j=0;j<ny;j++) {
+                m(i,j) = d[i][j];
+              }
+            }
+            return hpx::threads::run_as_hpx_thread([&]() {
```
hkaiser: If this was turned into a constructor for `ir::node_data<>` the conversion would be automatic.

examples/InitMatrix.py
```diff
@@ -0,0 +1,131 @@
...
+phy_print(m)
+
+print("Breast Cancer")
+
+m = et.file_read_csv("./algorithms/breast_cancer.csv")
+print(m)
+phy_print(m)
+phy_print(et.slice(m, 0, 3, 0, 3))
+three = et.phylisp_eval("3")
+four = et.phylisp_eval("4")
```
hkaiser: This probably has to be `et.eval(...)` now.

examples/algorithms/lra_csv_phyfun.py
```diff
@@ -0,0 +1,30 @@
+import phylanx
+et = phylanx.execution_tree
+from phylanx.util import *
+
+@phyfun
+def lra(x, y, alpha, iterations, enable_output):
+        weights=constant(0.0, shape(x, 1))
```
hkaiser: This file uses 8 spaces for indentation. Could we please agree on an indentation width for all Python files?

python/phylanx/util/__init__.py
```diff
@@ -8,3 +9,247 @@
...
+
+# Create the decorator
+class phyfun(object):
+    def __init__(self,f):
+        self.f = f
+        # Get the source code
+        src = inspect.getsource(f) #getsource(f)
+        # Before recompiling the code, take
+        # off the decorator from the source.
+        src = re.sub(r'^\s*@\w+\s*','',src)
```
hkaiser: This takes off all decorators, not only the `@phyfun` one.

python/src/phylanx.cpp
```diff
@@ -497,36 +535,404 @@ PYBIND11_MODULE(_phylanx, m)
...
+                using namespace phylanx::execution_tree;
+                return primitive{hpx::local_new<primitives::variable>(
+                    phylanx::ir::node_data<double>{m})};
+            });
+        },
+        "create a new variable from a matrix floating point values");
+
+    execution_tree.def("linspace",
+        [](double xmin, double xmax, std::size_t nx) {
```
hkaiser: As talked about this should be turned into a primitive instead of pre-filling the data in Python (-bindings) space.

python/src/phylanx.cpp
```diff
@@ -497,36 +535,404 @@ PYBIND11_MODULE(_phylanx, m)
...
+        .def("get",
+            [](phylanx::execution_tree::primitive const& p, int index) {
+                return hpx::threads::run_as_hpx_thread([&]() {
+                    using namespace phylanx::execution_tree;
+                    auto n = numeric_operand(p, {}).get();
+                    return n[index];
+                });
+            },
+            "Get the value at the specified index")
+        .def("get",
```
hkaiser: Should we access the matrix elements using the list syntax instead of `get()`?

phylanx/ir/node_data.hpp
```diff
@@ -114,6 +114,18 @@ namespace phylanx { namespace ir
...
+          const std::size_t nx = values.size();
+          const std::size_t ny = values[0].size();
+          storage2d_type store{nx,ny};
+          for(std::size_t i=0;i<nx;i++) {
+            for(std::size_t j=0;j<ny;j++) {
+              store(i,j) = values[i][j];
+            }
+          }
+          operator=(std::move(store));
+        }
```
hkaiser: Could you please stick to the formatting used in this file (i.e. 4 characters of whitespace used for indentation)?

phylanx/ir/node_data.hpp
```diff
@@ -114,6 +114,18 @@ namespace phylanx { namespace ir
...
+        {
+          const std::size_t nx = values.size();
+          const std::size_t ny = values[0].size();
+          storage2d_type store{nx,ny};
+          for(std::size_t i=0;i<nx;i++) {
+            for(std::size_t j=0;j<ny;j++) {
+              store(i,j) = values[i][j];
+            }
+          }
+          operator=(std::move(store));
```
hkaiser: Why do you implement the constructor in terms of the assignment operator?. That does not look right to me.

hkaiser: The following is how this constructor should be implemented:
```
        explicit node_data(std::vector<std::vector<T>> const& values)
          : data_(storage2d_type{values.size(), values[0].size()})
        {
            std::size_t const nx = values.size();
            for (std::size_t i = 0; i != nx; ++i)
            {
                std::vector<T> const& row = values[i];
                std::size_t const ny = row.size();
                for (std::size_t j = 0; j != ny; ++j)
                {
                    util::get<2>(data_)(i, j) = row[j];
                }
            }
        }
```

stevenrbrandt:
Hartmut, AFAIK, all issues you mentioned have been addressed and the build is succeeding.

hkaiser:
@stevenrbrandt This PR is in conflicted state now. 

Also, I think we should still add the exception for non-numeric types here: https://github.com/STEllAR-GROUP/phylanx/pull/142/files/c5fc204aac47a41dbe3ff5fd9f90717683e8307f#diff-72627d0c0e55a662e5966fe24191c190R211. 

In addition, there is still the question of how to access the matrix elements (see `phylanx.primitive.get()`).

python/src/phylanx.cpp
```diff
@@ -501,23 +547,82 @@ PYBIND11_MODULE(_phylanx, m)
...
+        expression_compiler_help);
+    execution_tree.def("eval",
+        expression_compiler<
+            phylanx::execution_tree::primitive const&,
+            phylanx::execution_tree::primitive const&,
+            phylanx::execution_tree::primitive const&,
+            phylanx::execution_tree::primitive const&,
+            phylanx::execution_tree::primitive const&,
+            phylanx::execution_tree::primitive const&>,
+        expression_compiler_help);
```
hkaiser: I just realized, that we have now two `eval()` functions. One exposed above, and one exposed by the `primitive` type below. Should this be properly consolidated?

stevenrbrandt:
OK, the examples are working, and exceptions are thrown for unsupported return types.

examples/algorithms/lra_csv_phyfun.py
```diff
@@ -0,0 +1,31 @@
+import phylanx
```
hkaiser: This file has no copyright and license

examples/algorithms/lra_csv_phyfun_np.py
```diff
@@ -0,0 +1,32 @@
+import phylanx
```
hkaiser: This file has no copyright and license

tests/unit/python/execution_tree/CMakeLists.txt
hkaiser: What should we do with this? I'd suggest to add the dependencies back as otherwise the build system might not work correctly.

stevenrbrandt:
OK, all issues are (once again) addressed and the CircleCI tests successful.

PR #143, opened 2018-01-11
Adding sequence numbers to the names generated for the primitives.

hkaiser:
This will enable identifying every primitive instance in an unique way (needed for performance counters).
The full name of every component is patterned after

     /phylanx/<primitive>:<sequence-nr>[:<instance>]/<compile_id>#<tag>

where:
```
     <primitive>:   the name of primitive type representing the given
                    node in the expression tree
     <sequence-nr>: the sequence number of the corresponding instance
                    of type <primitive>
     <instance>:    (optional), some primitives have additional instance
                    names, for instance references to function arguments
                    have the name of the argument as their <instance>
     <compile_id>:  the sequence number of the invocation of the
                    function phylanx::execution_tree::compile
     <tag>:         the index into the vector of iterators, where the
                    iterator refers to the point of usage of the
                    primitive in the compiled source code
```

### DIFF
--- a/phylanx/execution_tree/compiler/actors.hpp
+++ b/phylanx/execution_tree/compiler/actors.hpp
@@ -16,6 +16,7 @@
 #include <cstddef>
 #include <functional>
 #include <list>
+#include <map>
 #include <utility>
 #include <vector>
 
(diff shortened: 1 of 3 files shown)


hkaiser:
@parsa The sequence number should give you a means of ordering the primitive instances for the purpose of returning the performance counter information.

Issue #125, opened 2017-12-10
Possible visualization of expression evaluation


hkaiser:
This is meant to be a sketch of an idea how we could visualize the evaluation of an execution tree. Let's assume we have a simple expression `store(A, add(B, C))`, then the following visualization allows to quickly understand what part of the expression evaluation took how long:

![expr](https://user-images.githubusercontent.com/568063/33807778-757e8816-dda1-11e7-8e9b-c64dbcd6a47a.png)

An implementation of such a visualization would require to add some functionalities:

 - [x] associate a name with every instantiated primitive (possibly derived from the source code line the primitive was created from) (see #126)
 - [x] Generate topology of execution tree in a commonly used string format (see #148, #154)
 - [x] (optionally) measure the execution counts and times for each primitive (instrument `primitive::eval` and `primitive::eval_direct` actions), this could be done using performance counters (see #96)

@kisaacs, @khuck please add more tasks to this list from your perspective

Later on, the infrastructure built for this may be used to adaptively change the evaluation of parts of the execution tree by dynamically switching to direct execution in cases where the asynchronous (default) execution may impose too much overheads (see #121).

kisaacs:
It sounds like you know what data we need (tree specification, correspondence to code/variable names, metrics). Who do we see as the user for these? Phylanx developers or phylanx users? 

If it's developers first, as a first cut, we could write a script that outputs a static PDF -- this might match the quick command-line workflow (or even we could probably output them to the command line as ASCII or interactive ASCII with ncurses). If we find we want to look at more than one metric and we want something more general user-friendly, then we might want something like the Perfopticon visualization for distributed databases:

https://idl.cs.washington.edu/papers/perfopticon/
Perfopticon keeps the tree/graph simple and augments with quick interactive chart support.

We could also alter the tree drawing itself such that the length of the edges/nodes encodes time or counts. Some of these choices will be affected by how big we think these expressions are going to get.

hkaiser:
> It sounds like you know what data we need (tree specification, correspondence to code/variable names, metrics). Who do we see as the user for these? Phylanx developers or phylanx users?

For now it's mostly us developers being interested in getting our performance straight. Later on users might be interested as well, I'm not sure if that's too low level for them.

> Perfopticon keeps the tree/graph simple and augments with quick interactive chart support.

Looks nice.

> Some of these choices will be affected by how big we think these expressions are going to get.

[...]

PR #148, opened 2018-01-15
Generate newick tree

hkaiser:
This patch proposed to add the following additional facilities:

- Generate Newick tree from execution tree
- Parse primitive names into its parts

This PR is related to #125

### DIFF
--- a/examples/algorithms/lra_csv_instrumented.cpp
+++ b/examples/algorithms/lra_csv_instrumented.cpp
@@ -1,4 +1,4 @@
-//   Copyright (c) 2017 Hartmut Kaiser
+//   Copyright (c) 2017-2018 Hartmut Kaiser
 //
 //   Distributed under the Boost Software License, Version 1.0. (See accompanying
 //   file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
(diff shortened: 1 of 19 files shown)


hkaiser:
@kisaacs, @khuck please verify whether the generated tree topology is conforming to your Newick file format readers.

hkaiser:
Here is an example of a generated tree (slightly formatted):
```
(   (   (   (/phylanx/argument#0#filepath/0#7) /phylanx/file_read_csv#0/0#8,
            /phylanx/argument#1#row_start/0#9,
            /phylanx/argument#2#row_stop/0#10,
            /phylanx/argument#3#col_start/0#11,
            /phylanx/argument#4#col_stop/0#12
        ) /phylanx/slice#0/0#13
    ) /phylanx/define-function#0#read_x/0#1,
    /phylanx/function#0#read_x/0#15
) /phylanx/block#0/0#16;
```

where the string used for the tree nodes represent the names of the corresponding primitives (as registered with AGAS). This tree was generated from the following PhySL expression:
```
std::string const read_x_code = R"(block(
    //
    // Read X-data from given CSV file
    //
    define(read_x, filepath, row_start, row_stop, col_start, col_stop,
        slice(file_read_csv(filepath), row_start, row_stop, col_start, col_stop)
    ),
    read_x
))";
```

khuck:
@hkaiser I am not familiar with Newick trees... Kate and I had talked about graphviz trees, which are in "dot" format.  Unless there was a conversation that I don't remember.

kisaacs:
@khuck There are conversion utilities out there since they are both standard formats. (e.g., https://github.com/lindenb/newicktools )  Katy and I will try these soon so we can confirm with Hartmut.

hkaiser:
@kisaacs, @khuck well, if generating dot format would be more appropriate we can certainly change things.

kisaacs:
Here are two standard representations of trees using the sample data, right out of the d3js library. We were able to read it just fine. You may want to verify it is the intended structure.

The one with color is just using color to encode depth. The subtrees can be expanded and collapsed.

indentedexpressiontree
nodelinkexpressiontree

hkaiser:
@kisaacs The structure looks about right. Very nice! Thanks!

PR #150, opened 2018-01-16
Minimal performance information collection in base_primitive

parsa:


phylanx/execution_tree/primitives/base_primitive.hpp
```diff
@@ -177,7 +177,11 @@ namespace phylanx { namespace execution_tree { namespace primitives
         virtual primitive_result_type eval_direct(
             std::vector<primitive_argument_type> const& params) const
         {
-            return eval(params).get();
+            ++count_;
+            hpx::util::high_resolution_timer eval_timer(duration_);
+            const auto& ret = eval(params).get();
+            duration_ = eval_timer.elapsed_nanoseconds();
```
hkaiser: This will collect the execution time once only, overwriting the previous timings (which have to be accumulated).

phylanx/execution_tree/primitives/base_primitive.hpp
```diff
@@ -218,6 +222,8 @@ namespace phylanx { namespace execution_tree { namespace primitives
     protected:
         static std::vector<primitive_argument_type> noargs;
         std::vector<primitive_argument_type> operands_;
+        mutable std::int64_t count_;
+        mutable std::int64_t duration_;
```
hkaiser: Those new variables are not initialized anywhere

phylanx/execution_tree/primitives/base_primitive.hpp
```diff
@@ -177,7 +177,11 @@ namespace phylanx { namespace execution_tree { namespace primitives
         virtual primitive_result_type eval_direct(
             std::vector<primitive_argument_type> const& params) const
         {
-            return eval(params).get();
+            ++count_;
+            hpx::util::high_resolution_timer eval_timer(duration_);
+            const auto& ret = eval(params).get();
```
hkaiser: This should read:

    auto && ret = f.get();

otherwise the return value will not be moved.

Also,  using an RAII wrapper to collect the data might be more appropriate (see for instance https://github.com/STEllAR-GROUP/hpx/blob/master/hpx/util/scoped_timer.hpp).

hkaiser:
AFAIU, the data to be collected should reflect the numbers for the `eval()` function, not `eval_direct()`. So I think the function to instrument would be this:

https://github.com/STEllAR-GROUP/phylanx/blob/master/phylanx/execution_tree/primitives/base_primitive.hpp#L161-L165

phylanx/execution_tree/primitives/base_primitive.hpp
parsa: `eval_time` is initialized as`eval_timer(duration_)`. Why is this a problem?

phylanx/execution_tree/primitives/base_primitive.hpp
parsa: Resolved.

phylanx/execution_tree/primitives/base_primitive.hpp
parsa: Resolved.

hkaiser:
@parsa please remove the conflicts so this can be merged.

PR #158, opened 2018-01-19
String args

stevenrbrandt:
Support string variables within the Python interface.


phylanx/execution_tree/primitives/variable.hpp
```diff
@@ -24,7 +24,7 @@ namespace phylanx { namespace execution_tree { namespace primitives
     {
     public:
         variable() = default;
-        variable(std::string name);
+        variable(std::string data,std::string name);
```
hkaiser: I think , this is not necessary as a `primitive_argument_type` can be constructed from a `std::string` (see https://github.com/STEllAR-GROUP/phylanx/blob/master/phylanx/execution_tree/primitives/base_primitive.hpp#L112-L122).

src/execution_tree/primitives/console_output.cpp
```diff
@@ -43,7 +43,7 @@ namespace phylanx { namespace execution_tree { namespace primitives
     console_output::console_output(
             std::vector<primitive_argument_type>&& operands)
       : base_primitive(std::move(operands))
-    {}
+    {}
```
hkaiser: We will add the inspect checker to this repository soon. One of the things it will flag as an error is end-of-line-whitespace. Would you mind configuring your editor such that it removes that automatically?

src/execution_tree/primitives/console_output.cpp
```diff
@@ -91,11 +94,6 @@ namespace phylanx { namespace execution_tree { namespace primitives
     hpx::future<primitive_result_type> console_output::eval(
         std::vector<primitive_argument_type> const& args) const
     {
-        if (operands_.empty())
-        {
-            return std::make_shared<detail::console_output>()->eval(args, noargs);
-        }
-
```
hkaiser: Why have you removed this code?

src/execution_tree/primitives/variable.cpp
```diff
@@ -61,6 +62,7 @@ namespace phylanx { namespace execution_tree { namespace primitives
     variable::variable(std::vector<primitive_argument_type>&& operands,
             std::string name)
       : name_(std::move(name))
+      , data_(operands)
```
hkaiser: What's your rationale of adding this?

phylanx/execution_tree/primitives/variable.hpp
stevenrbrandt: The problem is, if the (std::string name) constructor exists, it somehow gets called. The variable 
 constructed by "return primitive{hpx::local_new<primitives::variable>(d,"")};" gets replaced by it, and the contents of data_ erased.

src/execution_tree/primitives/console_output.cpp
stevenrbrandt: In Python, if you do a print statement with no args, you should still print newline. This is for compatibility.

src/execution_tree/primitives/variable.cpp
stevenrbrandt: The data in operands was being discarded for no apparent reason. I assumed it was a bug.

src/execution_tree/primitives/console_output.cpp
stevenrbrandt: No problem.

stevenrbrandt:
Replied to all comments. Awaiting additional clarification.

phylanx/execution_tree/primitives/variable.hpp
hkaiser: Hmmm, sorry - I don't understand what the problem is. Could you elaborate, please?

src/execution_tree/primitives/console_output.cpp
hkaiser: Ok, this is a bug in the primitive, then (I created a ticket, #159). Please revert this change as it would possibly break lazy argument evaluation.

src/execution_tree/primitives/variable.cpp
hkaiser: `data_` is properly initialized here: https://github.com/STEllAR-GROUP/phylanx/blob/master/src/execution_tree/primitives/variable.cpp#L75, isn't it?

phylanx/execution_tree/primitives/variable.hpp
stevenrbrandt: I can only describe what happens, I can't make sense of it. If a constructor is present that only sets the name, then regardless of how I instantiate the variable I pass to hpx::local_new<primitives::variable>, the name is extracted and a copy of the variable with the data erased is passed into the component. Apparently, nothing currently requires the name-only constructor at present anyway.

hkaiser: That sounds like a bug to me. Could you create a small test case that reproduces the issue you're seeing, please?

src/execution_tree/primitives/console_output.cpp
hkaiser: @stevenrbrandt as assumed, this was a bug in the implementation of the actual output operation. Please see #160 for a proper fix.

hkaiser: Please remove this change from the PR as the problem you were referring to has already been fixed on master.

src/execution_tree/primitives/variable.cpp
hkaiser: Please remove this change from the proposed changeset as well. The initialization of `data_` here is not necessary if not wrong.

Issue #96, opened 2017-11-02
Create performance counters for primitives


parsa:
This should be done in a way that all primitives almost automatically have performance counters associated with each of them. We should at least create two performance counters for each of the primitives:

    /phylanx{locality#N/<primitive-name>}/invocation-count
    /phylanx{locality#N/<primitive-name>}/execution-time

hkaiser:
The names for the counters should follow the established practice. Here is a list of possible names for the counters we should create for each of the primitives:

    /phylanx{locality#N/total}/<primitive>/count/eval
    /phylanx{locality#N/total}/<primitive>/time/eval
    /phylanx{locality#N/total}/<primitive>/count/direct-eval
    /phylanx{locality#N/total}/<primitive>/time/direct-eval

where `<primitive>` represents the primitive type the counter will return data for. 

All returned values should be cumulative, i.e. the values should represent the accumulated counts/times since the counter was reset last. The returned times should have nanosecond resolution.

Also, while discussing this, we decided that the counters should return a `std::vector<std::int64_t>` of values (i.e. the counter types should be `hpx::performance_counter::raw_values_counter` - note that this type does not exist yet, see https://github.com/STEllAR-GROUP/hpx/issues/3102).

The values in the returned array should be ordered based on the sequence number that is encoded in the primitive names (as they are registered in AGAS), see https://github.com/STEllAR-GROUP/phylanx/pull/143 for a description of the structure of the generated primitive names.

Examples for the counter names used would be:

     /phylanx{locality#0/total}/define-function/count/eval
     /phylanx{locality#0/total}/add/time/direct-eval

hkaiser:
@parsa please have a look at https://github.com/STEllAR-GROUP/hpx/pull/3103, those changes will most probably needed for your work. Please comment over there if that's what you need for your work on the performance counters.

hkaiser:
This will also require for https://github.com/STEllAR-GROUP/hpx/pull/3118 to be merged in order to be able to (re-)initialize the counters with the current list of instantiated primitives.

hkaiser:
This was fixed by merging #176

PR #165, opened 2018-01-24
Primitive performance counter infrastructure

parsa:
This PR implements a performance counter for each primitive.

### DIFF
--- a/src/performance_counters/register_counters.cpp
+++ b/src/performance_counters/register_counters.cpp
@@ -149,6 +149,13 @@ namespace phylanx { namespace performance_counters {
                     name + " primitive",
                 &primitive_counter_creator,
                 &hpx::performance_counters::locality_counter_discoverer);
+            hpx::performance_counters::install_counter_type(
+                "/phylanx/primitives/" + name + "/count/eval",
+                hpx::performance_counters::counter_raw_values,
+                "returns the number of times of eval() function of each " +
+                name + " primitive instance was called",
+                &primitive_counter_creator,
+                &hpx::performance_counters::locality_counter_discoverer);
         }
     }
 
--- a/src/performance_counters/register_counters.cpp
+++ b/src/performance_counters/register_counters.cpp
@@ -5,6 +5,7 @@
 
 #include <phylanx/config.hpp>
 #include <phylanx/ir/node_data.hpp>
+#include <phylanx/execution_tree/compile.hpp>
 
 #include <hpx/runtime/startup_function.hpp>
 #include <hpx/include/util.hpp>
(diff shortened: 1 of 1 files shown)


src/performance_counters/register_counters.cpp
```diff
@@ -46,6 +132,31 @@ namespace phylanx { namespace performance_counters
...
+        {
+            auto const& pattern = *(patterns.begin());
+            std::string const& name = hpx::util::get<0>(pattern);
+            hpx::performance_counters::install_counter_type(
+                "/phylanx/primitives/" + name + "/time/eval",
+                hpx::performance_counters::counter_raw_values,
+                "returns the total execution time of eval() function of the " +
+                    name + " primitive",
+                &primitive_counter_creator,
+                &hpx::performance_counters::locality_counter_discoverer);
```
hkaiser: This counter type should have a unit of measure defined (probably `ns`).

src/performance_counters/register_counters.cpp
```diff
@@ -46,6 +150,38 @@ namespace phylanx { namespace performance_counters
...
+        // Iterate and register a time and count performance counter per each primitive
+        for (auto const& patterns : pattern_list)
+        {
+            // No need to process all entries in the list
+            auto const& pattern = *(patterns.begin());
+            // The name of the primitive
+            std::string const& name = hpx::util::get<0>(pattern);
+            // Register a primitive time performance counter
+            hpx::performance_counters::install_counter_type(
+                "/phylanx/time/primitives/" + name + "/eval",
```
hkaiser: May I suggest we stick to the naming conventions as outlined in #96? This would make the new counter names conform to other, already existing counters.

src/performance_counters/register_counters.cpp
```diff
@@ -46,6 +150,38 @@ namespace phylanx { namespace performance_counters
...
+                hpx::performance_counters::counter_raw_values,
+                "returns the total execution time of eval() function of the " +
+                    name + " primitive",
+                &primitive_counter_creator,
+                &hpx::performance_counters::locality_counter_discoverer,
+                HPX_PERFORMANCE_COUNTER_V1, "ns");
+            // Register a primitive count performance counter
+            hpx::performance_counters::install_counter_type(
+                "/phylanx/count/primitives/" + name + "/eval",
```
hkaiser: Same here, please stick to the existing naming conventions.

src/performance_counters/register_counters.cpp
```diff
@@ -46,6 +150,38 @@ namespace phylanx { namespace performance_counters
...
+                hpx::performance_counters::counter_raw_values,
+                "returns the number of times of eval() function of each " +
```
hkaiser: Could you fix the grammar of the help text, please?

src/performance_counters/register_counters.cpp
parsa: The naming convention in #96 is hard to read. e.g Upon seeing a list of performance counters `/phylanx{locality#*/total}/add/count/eval` does not suggest that this performance counter measures something about add primitives. In this respect `/phylanx{locality#N/total}/primitives/<primitive>/count/eval` is better.

src/performance_counters/register_counters.cpp
parsa: Updated. Check again, please.

src/performance_counters/register_counters.cpp
hkaiser: Ok, but the currently registered name is different from that.

parsa: Must have been a mistake I amended. That's what it looks like right now.

hkaiser: Ok.

PR #166, opened 2018-01-24
added node_data<bool> to argument_value_type

scheherzade:


tests/unit/ir/node_data.cpp
```diff
@@ -41,6 +41,22 @@ int main(int argc, char* argv[])
...
+    {
+        phylanx::ir::node_data<bool> single_value(false);
+        HPX_TEST_EQ(single_value[0], false);
+
+        //auto begin = hpx::util::begin(single_value);
+        //auto end = hpx::util::end(single_value);
+        //
+        //HPX_TEST(begin != end);
+        //HPX_TEST_EQ(*begin, false);
+        //HPX_TEST(++begin == end);
```
hkaiser: I think we can remove the commented code now as we have dropped `node_data::begin/end`.

src/execution_tree/primitives/base_primitives.cpp
```diff
@@ -493,7 +533,7 @@ namespace phylanx { namespace execution_tree
         case 0:     // nil
             return ast::nil{};
 
-        case 1:     // bool
+        case 1:    // phylanx::ir::node_data<bool>
             return util::get<1>(val);
```
hkaiser: This should be `return util::get<1>(val).ref();`

src/execution_tree/primitives/base_primitives.cpp
```diff
@@ -554,8 +594,8 @@ namespace phylanx { namespace execution_tree
     {
         switch (val.index())
         {
-        case 1:     // bool
-            return ir::node_data<double>{double(util::get<1>(val))};
+        case 1:    // phylanx::ir::node_data<bool>
+            return util::get<4>(val).ref();
```
hkaiser: This has to be `return util::get<1>(val).ref();`

src/execution_tree/primitives/base_primitives.cpp
```diff
@@ -581,8 +621,8 @@ namespace phylanx { namespace execution_tree
     {
         switch (val.index())
         {
-        case 1:     // bool
-            return ir::node_data<double>{double(util::get<1>(std::move(val)))};
+        case 1:    // phylanx::ir::node_data<bool>
+            return util::get<4>(std::move(val));
```
hkaiser: This as well: `return util::get<1>(std::move(val));`.

src/execution_tree/primitives/base_primitives.cpp
```diff
@@ -1195,7 +1235,7 @@ namespace phylanx { namespace execution_tree
         case 0:
             return ast::nil{};
 
-        case 1:     // bool
+        case 1:    // phylanx::ir::node_data<bool>
```
hkaiser: I think that this switch has to be left alone as it looks at the variant stored in an AST.

phylanx/ir/node_data.hpp
```diff
@@ -218,6 +242,14 @@ namespace phylanx { namespace ir
...
         }
 
+        template <typename U,
+            typename U1 =
+                typename std::enable_if<!std::is_same<T, U>::value>::type>
+        node_data(node_data<U> const& d)
+          : data_(init_data_from_type(d))
+        {
+        }
+
```
hkaiser: Shouldn't be something similar be added for `node_data& operator=(node_data<U> const&)`?

phylanx/ir/node_data.hpp
```diff
@@ -370,6 +371,17 @@ namespace phylanx { namespace ir
             return *this;
         }
 
+        template <typename U,
+            typename U1 =
+                typename std::enable_if<!std::is_same<T, U>::value>::type>
+        node_data& operator=(node_data const& d)
```
hkaiser: Shouldn't this be:

    node_data& operator=(node_data<U> const& d) {...}

?

PR #169, opened 2018-01-25
Collect values for primitive performance counter

parsa:
This PR connects primitive performance counter to the values collected inside `base_primitive`

### DIFF
--- a/phylanx/execution_tree/primitives/base_primitive.hpp
+++ b/phylanx/execution_tree/primitives/base_primitive.hpp
@@ -1,4 +1,5 @@
 //  Copyright (c) 2017-2018 Hartmut Kaiser
+//  Copyright (c) 2018 Parsa Amini
 //
 //  Distributed under the Boost Software License, Version 1.0. (See accompanying
 //  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
@@ -281,6 +282,16 @@ namespace phylanx { namespace execution_tree { namespace primitives
         std::uint32_t pin_count() const
         {
             return 0;
+        }
+
+        const std::int64_t get_count()
+        {
+            return count_;
+        }
+
+        const std::int64_t get_duration()
+        {
+            return duration_;
         }
 
     protected:
--- a/phylanx/execution_tree/primitives/base_primitive.hpp
+++ b/phylanx/execution_tree/primitives/base_primitive.hpp
@@ -284,12 +284,12 @@ namespace phylanx { namespace execution_tree { namespace primitives
             return 0;
         }
 
-        const std::int64_t get_count()
+        std::int64_t get_count() const
         {
             return count_;
         }
 
-        const std::int64_t get_duration()
+        std::int64_t get_duration() const
         {
             return duration_;
         }
--- a/src/performance_counters/register_counters.cpp
+++ b/src/performance_counters/register_counters.cpp
@@ -44,7 +44,11 @@ namespace phylanx { namespace performance_counters
             value.count_ = ++invocation_count_;
 
             // Need to call reinit here if it has never been called before.
-            reinit(false);
+            if (!first_init)
+            {
+                first_init = true;
+                reinit(false);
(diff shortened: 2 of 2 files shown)


src/performance_counters/register_counters.cpp
```diff
@@ -41,7 +43,14 @@ namespace phylanx { namespace performance_counters
             value.status_ = hpx::performance_counters::status_new_data;
             value.count_ = ++invocation_count_;
 
-            /*value.values_ = std::move(result);*/
+            std::vector<std::int64_t> result;
+            // Extract the values from instances_
```
hkaiser: You might want to reserve the required space in this vector (`result.reserve(instances_.size());`)

src/performance_counters/register_counters.cpp
```diff
@@ -41,7 +43,14 @@ namespace phylanx { namespace performance_counters
...
+            for (auto const& instance : instances_)
+            {
+                result.push_back(instance->get_duration());
```
hkaiser: This counter type is used for both, the timing and the count. That means that it will not return the proper results for the counts.

phylanx/execution_tree/primitives/base_primitive.hpp
```diff
@@ -283,6 +284,16 @@ namespace phylanx { namespace execution_tree { namespace primitives
             return 0;
         }
 
+        std::int64_t get_count()
```
hkaiser: This function can be `const` (same is valid for `get_duration()` below).

src/performance_counters/register_counters.cpp
```diff
@@ -41,7 +43,14 @@ namespace phylanx { namespace performance_counters
             value.status_ = hpx::performance_counters::status_new_data;
```
hkaiser: You need to call `reinit` here if it has never been called before.

phylanx/execution_tree/primitives/base_primitive.hpp
parsa: Done.

src/performance_counters/register_counters.cpp
parsa: Done.

src/performance_counters/register_counters.cpp
parsa: Resolved.

src/performance_counters/register_counters.cpp
parsa: Please check again.

src/performance_counters/register_counters.cpp
```diff
@@ -32,6 +32,9 @@ namespace phylanx { namespace performance_counters
           : hpx::performance_counters::base_performance_counter<
                 primitive_counter>(info)
         {
+            hpx::performance_counters::counter_path_elements paths;
+            hpx::performance_counters::get_counter_path_elements(info.fullname_, paths);
+            duration_counter = paths.countername_.find("count") != std::string::npos;
```
hkaiser: `duration_counter` will be true for performance counters that are supposed to report count numbers, not durations.

phylanx/execution_tree/primitives/base_primitive.hpp
```diff
@@ -283,6 +284,16 @@ namespace phylanx { namespace execution_tree { namespace primitives
             return 0;
         }
 
+        const std::int64_t get_count()
```
hkaiser: That's not what I meant, `const std::int64_t get_count()` means that the returned value is `const`, which is meaningless as it is an rvalue anyways. WhatI meant was that this member function should be callable for `const` objects: `std::int64_t get_count() const`.

src/performance_counters/register_counters.cpp
parsa: Fixed.

phylanx/execution_tree/primitives/base_primitive.hpp
parsa: Fixed.

PR #171, opened 2018-01-26
Changing the match_data structure to better reflect the requirements of its usage

hkaiser:
- this is a breaking change requiring adaptation of all code using match patterns


parsa:
It works and looks fine.

PR #172, opened 2018-01-26
Support Python's slice operator.

stevenrbrandt:
Map python `[:,:]` operator's to slice() in PhySL.


hkaiser:
Neat!

python/phylanx/util/__init__.py
```diff
@@ -56,6 +77,39 @@ def recompile(self,a,allowreturn=False):
...
+            s = "slice("
+            s += sname
+            s += ","
+            if xlo == None:
+                s += "0"
+            else:
+                s += self.recompile(xlo)
+            s += ","
+            if xhi == None:
+                s += "shape(" + sname + ",0)"
```
hkaiser: You could use `-1` as the index to refer to the last row.

python/phylanx/util/__init__.py
```diff
@@ -56,6 +77,39 @@ def recompile(self,a,allowreturn=False):
...
+            else:
+                s += self.recompile(xhi)
+            s += ","
+            if ylo == None:
+                s += "0"
+            else:
+                s += self.recompile(ylo)
+            s += ","
+            if yhi == None:
+                s += "shape(" + sname + ",1)"
```
hkaiser: You could use `-1` as the index to refer to the last column.

python/phylanx/util/__init__.py
```diff
@@ -35,6 +35,27 @@ def phy_print(m):
...
+        for i in range(len(args)):
+            a = args[i]
+            nm = a.__class__.__name__
+            if nm == name:
+                if num == None or num == i:
+                    return a
+    elif num != None:
+        if num < len(args):
+            return args[num]
+
```
hkaiser: Missing `else`?

python/phylanx/util/__init__.py
stevenrbrandt: Actually, seeing this comment makes me realize the semantics of slice() does not match Python. The -1 in the last position is not supposed to work. In addition, it should be possible to have a third argument.
```
>>> r = range(10)
>>> print r[:]
[0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
>>> print r[0:-1]
[0, 1, 2, 3, 4, 5, 6, 7, 8]
>>> print r[::2]
[0, 2, 4, 6, 8]
>>> print r[1::2]
[1, 3, 5, 7, 9]
>>> print r[::-1]
[9, 8, 7, 6, 5, 4, 3, 2, 1, 0]
````

python/phylanx/util/__init__.py
stevenrbrandt: The -1 in the last position should cut off the last argument, not show the whole array.

python/phylanx/util/__init__.py
stevenrbrandt: The intent was to return None in the else case. That's what happens the way the code is written.

hkaiser: Ok, understood. Wouldn't it be better to be explicit about this, however?

python/phylanx/util/__init__.py
hkaiser: Yes, you're certainly right.

python/phylanx/util/__init__.py
hkaiser: Yes, you're certainly right.

hkaiser:
Could we start adding thorough tests for the `@phyfun` decorator, please?

PR #175, opened 2018-01-29
Supporting direct positional (symbol) information in PhySL

hkaiser:
PhySL now allows for any symbol (identifier or function_call) to encode the line/column information
of the source code it was generated from: `<ident>#line#column`. This information is also encoded in the generated primitive names: 

    /phylanx/<primitive>#<sequence-nr>[#<instance>]/<compile_id>#<tag1>[#<tag2>]

where `tag1` is either the offset as before (if `tag2` is missing), otherwise `tag1`/`tag2` encode the line/column.

- started to add positional information to PhySL generated from `@phyfun` decorator
- added test for facilities related to primitive_name
- flyby: fixing ast.natvis

### DIFF
--- a/tests/unit/python/execution_tree/eval.py
+++ b/tests/unit/python/execution_tree/eval.py
@@ -1,4 +1,4 @@
-#  Copyright (c) 2017 Hartmut Kaiser
+#  Copyright (c) 2017-2018 Hartmut Kaiser
 #  Copyright (c) 2018 R. Tohid
 #  Copyright (c) 2018 Steven R. Brandt
 #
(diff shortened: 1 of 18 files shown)


hkaiser:
@stevenrbrandt this PR contains some changes to the `@phyfun` decorator code. May I ask you to verify that this doesn't break anything? Please note that this most likely does not change all required spots in the decorator. Feel free to point out those, I'd be happy to adapt things.

tests/unit/python/execution_tree/eval.py
```diff
@@ -40,10 +40,19 @@ def fib(n):
     else:
         return fib(n-1)+fib(n-2)
 
+assert fib.__physl_src__ == \
+    'block#1#0(define#1#0(fib#1#0, n#1#8, ' + \
+        'if(n#2#7 < 2, n#3#15, ' + \
+            '(fib((n#5#19 - 1)) + fib((n#5#28 - 2))))' + \
+        '), fib#1#0)\n'
+
 assert fib(10)[0] == 55.0
```
stevenrbrandt: Probably you want to run this change before you replace the source as well as after. Otherwise, you're disabling the previous test.

tests/unit/python/execution_tree/eval.py
```diff
@@ -40,10 +40,19 @@ def fib(n):
...
 
 @phyfun
 def pass_str(a):
     return a
 
+assert pass_str.__physl_src__ == \
+    'block#1#0(define#1#0(pass_str#1#0, a#1#13, a#2#11), pass_str#1#0)\n'
+
 assert "foo" == str(pass_str("foo"))
```
stevenrbrandt: Likewise, you probably want to copy the assertion to run before as well as after the src change.

tests/unit/python/execution_tree/eval.py
hkaiser: Sorry, I don't know what you're referring to - care to elaborate?

hkaiser: @stevenrbrandt could you elaborate, please?

hkaiser:
The Appveyor tests failed for unrelated reasons.

PR #179, opened 2018-01-29
Primitive linspace

rtohid:


src/execution_tree/primitives/linspace.cpp
```diff
@@ -0,0 +1,112 @@
...
+                if (num_samples < 1)
+                {
+                    HPX_THROW_EXCEPTION(hpx::bad_parameter,
+                        "phylanx::execution_tree::primitives::detail::linspace1d",
+                        "the linspace primitive requires at least one interval");
+                }
+                else if (1 == num_samples)
+                {
+                    vector_type result{start};            
+                    return operand_type{result};
```
hkaiser: You might want to move the `result`: `return operand_type{std::move(result)};`

src/execution_tree/primitives/linspace.cpp
```diff
@@ -0,0 +1,112 @@
...
+                }
+                else
+                {
+                  auto result = vector_type(num_samples);
+                  double dx = (stop - start) / (num_samples - 1);
+                  for (std::size_t i = 0; i < num_samples; i++)
+                  {
+                    result[i] = start + dx * i;
+                  }
+                  return operand_type{result};
```
hkaiser: You might want to move the `result`: `return operand_type{std::move(result)};`

src/execution_tree/primitives/linspace.cpp
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    hpx::future<primitive_result_type> linspace::eval(
+        std::vector<primitive_argument_type> const& args) const
+    {
+        if (operands_.empty())
+        {
+            return std::make_shared<detail::linspace>()->eval(args);
+        }
+        return std::make_shared<detail::linspace>()->eval(operands_);
```
hkaiser: You should pass along `args` for lazy argument evaluation. Otherwise constructs like

    define(somefunc, a, b, linspace(a, b, 1)),
    somefunc(1, 2)

will not work properly.Please see the other primitives for how it has to be done.

hkaiser:
Also, inspect is unhappy with your patch: https://862-101409977-gh.circle-artifacts.com/0/tmp/circle-artifacts.bS6qesy/phylanx_inspect_report.html.

rtohid:
@justwagle, thank you.
@hkaiser, What do you think? Should I also check here, or the check being done in extract_integer_value would suffice?

PR #187, opened 2018-02-01
Equal_elementwise Primitive

scheherzade:


hkaiser:
@scheherzade inspect is unhappy: https://911-101409977-gh.circle-artifacts.com/0/tmp/circle-artifacts.j1r2usa/phylanx_inspect_report.html

src/execution_tree/primitives/equal_elementwise.cpp
```diff
@@ -0,0 +1,416 @@
...
+                            "equal_elementwise::eval",
+                            "left hand side and right hand side are "
+                            "incompatible "
+                            "and can't be compared");
+                    }
+
+                    primitive_result_type operator()(
+                        ir::node_data<double>&& lhs,
+                        ir::node_data<bool>&& rhs) const
+                    {
```
hkaiser: This overload shouldn't be necessary as the overload above (`operator()(T1, T2)`) covers this case.

src/execution_tree/primitives/equal_elementwise.cpp
```diff
@@ -0,0 +1,416 @@
...
+                        HPX_THROW_EXCEPTION(hpx::bad_parameter,
+                            "equal_elementwise::eval",
+                            "left hand side and right hand side are "
+                            "incompatible "
+                            "and can't be compared");
+                    }
+
+                    primitive_result_type operator()(ir::node_data<bool>&& lhs,
+                        ir::node_data<double>&& rhs) const
+                    {
```
hkaiser: Same as above, this is not necessary.

hkaiser:
@scheherzade actually, how much different is this from the existing equal primitive? 

It seems to me (but I have not looked closer) that these two are very similar and differ only in how they handle the comparison results. If it turns out that the two primitive are nearly identical we could unify them into one primitive that handles both cases.

src/execution_tree/primitives/equal.cpp
```diff
@@ -55,24 +55,63 @@ namespace phylanx { namespace execution_tree { namespace primitives
...
-            bool equal1d1d(operand_type&& lhs, operand_type&& rhs) const
+            primitive_argument_type equal1d0d(
+                operand_type&& lhs, operand_type&& rhs) const
+            {
+                // TODO: SIMD functionality should be added, blaze implementation
+                // is not currently available
+                lhs.vector() = blaze::map(lhs.vector(),
+                    [&](double x) { return (x == rhs.scalar()); });
+
+                return primitive_argument_type(ir::node_data<bool>{lhs});
```
hkaiser: You might want to consider moving lhs into the result:

     return primitive_argument_type(ir::node_data<bool>{std::move(lhs)});

PR #189, opened 2018-02-02
Instrument Primitive Performance Information

parsa:
This PR demonstrates how to complement primitive performance counter information with information inside AGAS to produce and prints that information in CSV.

> **Note**:
> This PR includes a temporary workaround that needs to be removed after #190 is resolved.

### DIFF
--- a/examples/profiling/primitive_instrumentation.cpp
+++ b/examples/profiling/primitive_instrumentation.cpp
@@ -68,6 +68,11 @@ std::map<std::string, std::vector<std::int64_t>> retrieve_counter_data(
     // Return value
     std::map<std::string, std::vector<std::int64_t>> result;
 
+    // Reuse get_counter_values_array calls
+    //   key: primitive type
+    //   value: vector of counter_values_array instance
+    std::map<std::string, std::vector<pc::counter_values_array>> pc_values;
+
     // Iterate through all provided primitive instances
     for (auto const& name : primitive_instances)
     {
@@ -78,20 +83,23 @@ std::map<std::string, std::vector<std::int64_t>> retrieve_counter_data(
         // TODO: Ensure counter_name_last_part has at least one entry
 
         // Performance counter values
-        std::vector<pc::counter_values_array> counter_values;
-
-        // Iterate through the last parts of performance counter names
-        for (auto const& counter_name_last_part : counter_name_last_parts)
+        std::vector<pc::counter_values_array> &counter_values =
+            pc_values[tags.primitive];
+        if (counter_values.empty())
         {
-            // NOTE: Reuse the get_counter_values_array call?
-            // 
-            // Construct the name of the counter
-            std::string counter_name(
-                "/phylanx/primitives/" + tags.primitive + "/" + counter_name_last_part);
-            // The actual performance counter
-            pc::performance_counter counter(counter_name, locality_id);
-            counter_values.push_back(
-                counter.get_counter_values_array(hpx::launch::sync, false));
+            // Iterate through the last parts of performance counter names
+            for (auto const& counter_name_last_part : counter_name_last_parts)
+            {
+                // NOTE: Reuse the get_counter_values_array call?
+                //
+                // Construct the name of the counter
+                std::string counter_name("/phylanx/primitives/" +
+                    tags.primitive + "/" + counter_name_last_part);
+                // The actual performance counter
+                pc::performance_counter counter(counter_name, locality_id);
+                counter_values.push_back(
+                    counter.get_counter_values_array(hpx::launch::sync, false));
+            }
         }
 
         // HACK: block 0 does not appear in AGAS
@@ -102,12 +110,12 @@ std::map<std::string, std::vector<std::int64_t>> retrieve_counter_data(
             continue;
         }
 
-        std::vector<std::int64_t> data(counter_values.size());
+        std::vector<std::int64_t> data(counter_name_last_parts.size());
         for (int i = 0; i < counter_values.size(); ++i)
         {
             data[i] = counter_values[i].values_[tags.sequence_number];
         }
-        
+
         result[name] = data;
     }
 
--- a/examples/profiling/primitive_instrumentation.cpp
+++ b/examples/profiling/primitive_instrumentation.cpp
@@ -10,79 +10,31 @@
 #include <hpx/include/agas.hpp>
 #include <hpx/runtime_fwd.hpp>
 
-#include <cstddef>
-#include <cstdint>
 #include <iostream>
-#include <map>
-#include <vector>
 #include <string>
 #include <utility>
 
-#include <boost/program_options.hpp>
-#include <blaze/Math.h>
-
-//////////////////////////////////////////////////////////////////////////////////
-// This example uses part of the breast cancer dataset from UCI Machine Learning
-// Repository:
-//     https://archive.ics.uci.edu/ml/datasets/Breast+Cancer+Wisconsin+(Diagnostic)
-//
-// A copy of the full dataset in CSV format (breast_cancer.csv), obtained from
-// scikit-learn datasets, is provided in the same folder as this example.
-//
-// The layout of the data in the provided CSV file used by the example
-// is as follows:
-// 30 features per line followed by the classification
-// 569 lines of data
-//
-// This example also demonstrates how the generated primitives can be introspected
-// and linked back to the source code.
-//
-/////////////////////////////////////////////////////////////////////////////////
-
-std::string const read_x_code = R"(block(
-    //
-    // Read X-data from given CSV file
-    //
-    define(read_x, filepath, row_start, row_stop, col_start, col_stop,
-        slice(file_read_csv(filepath), row_start, row_stop, col_start, col_stop)
-    ),
-    read_x
-))";
-
-std::string const read_y_code = R"(block(
-    //
-    // Read Y-data from given CSV file
-    //
-    define(read_y, filepath, row_start, row_stop,
-        slice(file_read_csv(filepath), row_start, row_stop, -1, 0)
-    ),
-    read_y
-))";
-
 ///////////////////////////////////////////////////////////////////////////////
-std::string const lra_code = R"(block(
+char const* const lra_code = R"(block(
     //
     // Logistic regression analysis algorithm
     //
-    //   x: [N, M]
-    //   y: [N]
-    //
-    define(lra, x, y, alpha, iterations, enable_output,
+    //   x: [30, 2]
+    //   y: [30]
+    define(lra, x, y, alpha,
         block(
-            define(weights, constant(0.0, shape(x, 1))),            // weights: [M]
-            define(transx, transpose(x)),                           // transx:  [M, N]
+            define(weights, constant(0.0, shape(x, 1))),     // weights: [2]
+            define(transx, transpose(x)),                    // transx:  [2, 30]
             define(pred, constant(0.0, shape(x, 0))),
             define(error, constant(0.0, shape(x, 0))),
             define(gradient, constant(0.0, shape(x, 1))),
             define(step, 0),
             while(
-                step < iterations,
+                step < 10,
                 block(
-                    if(enable_output, cout("step: ", step, ", ", weights)),
-                    // exp(-dot(x, weights)): [N], pred: [N]
                     store(pred, 1.0 / (1.0 + exp(-dot(x, weights)))),
-                    store(error, pred - y),                         // error: [N]
-                    store(gradient, dot(transx, error)),            // gradient: [M]
+                    store(error, pred - y),                  // error: [30]
+                    store(gradient, dot(transx, error)),     // gradient: [2]
                     parallel_block(
                         store(weights, weights - (alpha * gradient)),
                         store(step, step + 1)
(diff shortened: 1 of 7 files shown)


parsa:
@hkaiser Current output looks like this:
```
primitive,sequence_number,instance_id,compile_id,tag1,tag2,count,time,direct_count,direct_time
"block",1,,0,141,-1,10,26915755,0,0
"parallel_block",0,,0,815,-1,10,34686249,0,0
"dot",0,,0,621,-1,10,10528868,0,0
"dot",1,,0,753,-1,10,12347257,0,0
"while",0,,0,514,-1,1,1543407,0,0
"constant",0,,0,176,-1,0,0,1,2706202
"constant",1,,0,331,-1,0,0,1,1792731
"constant",2,,0,386,-1,0,0,1,1675553
"constant",3,,0,444,-1,0,0,1,1648183
"exp",0,,0,616,-1,10,15969494,0,0
"shape",0,,0,190,-1,1,771917,0,0
"shape",1,,0,345,-1,1,440485,0,0
"shape",2,,0,400,-1,1,457591,0,0
"shape",3,,0,458,-1,1,439202,0,0
"transpose",0,,0,252,-1,0,0,1,1298361
"add",0,,0,610,-1,10,18653887,0,0
"add",1,,0,937,-1,10,6856167,0,0
"div",0,,0,603,-1,10,21809553,0,0
"mul",0,,0,881,-1,10,10935140,0,0
"sub",0,,0,674,-1,10,10950535,0,0
"sub",1,,0,870,-1,10,19380472,0,0
"lt",0,,0,537,-1,11,8021100,0,0
"store",0,,0,591,-1,10,23892234,0,0
"store",1,,0,661,-1,10,13003278,0,0
"store",2,,0,737,-1,10,14319176,0,0
"store",3,,0,855,-1,10,21392589,0,0
"store",4,,0,925,-1,10,8925590,0,0
"minus",0,,0,621,-1,10,13177765,0,0
```
Does it need to be transposed? Is it missing any other information?

hkaiser:
@parsa this doesn't look too bad, except that each line should refer to a primitive __instance__, not a primitive __type__.

Also, looking forward, this functionality should be encapsulated in a separate function (to be integrated with Phylanx later on). I think the following API would be a good starting point for discussions on how to expose things:
```
namespace phylanx { namespace execution_tree
{
    std::vector<std::vector<std::int64_t>> retrieve_counter_data(
        std::vector<std::string> const& primitive_instances,
        std::vector<std::string> const& counter_name_templates);
}}
```
where:
- `primitive_instances`: the list of primitive instance names for which to retrieve the counter data, e.g. `/phylanx/add#1/30#2`, or similar
- `counter_name_templates`: the list of counter name (templates) identifying the performance counters to retrieve. This could be something like `/phylanx{locality#0}/primitives/%s/count/eval`, and the `%s` should be replaced by the concrete counter type name (e.g. `add`).

The function then should return an array of counter values, one entry for each of the primitive instances sequenced as specified by the first argument. The counter values for each entry in that array should be sequenced as specified by the second argument.

Alternatively (or additionally), the first argument could be an instance of a `phylanx::execution_tree::topology` as returned by the `primitive::expression_topology()` function.

parsa:
> … except that each line should refer to a primitive instance, not a primitive type.

All information we have about the primitive instance is available in the each CSV line. What is missing?

> `primitive_instances`: the list of primitive instance names for which to retrieve the counter data, e.g. `/phylanx/add#1/30#2`, or similar

There needs to be API that lets developers list and filter primitives instead of telling them Phylanx registers them in AGAS and they have to query AGAS and process the list themselves.

> This could be something like `/phylanx{locality#0}/primitives/%s/count/eval`, and the `%s` should be replaced by the concrete counter type name (e.g. `add`).

Using `%s` is a different convention than what is used in HPX and Phylanx and I do not like it because it may confuse to users.

hkaiser:
> > … except that each line should refer to a primitive instance, not a primitive type.
>
> All information we have about the primitive instance is available in the each CSV line. What is missing?

Ok, I misread the output. I would have expected to see the full primitive instance name in the first column, though.

hkaiser:
> There needs to be API that allows users to list and filter primitives instead of telling them that Phylanx registers all primitives with the symbolic namespace and that they have to query AGAS and sort through the list themselves.

We have an API returning the topological structure of a given execution tree: https://github.com/STEllAR-GROUP/phylanx/blob/master/phylanx/execution_tree/primitives/base_primitive.hpp#L116-L117. Other APIs could be created, but that's orthogonal to the issue at hand. The function I mentioned is needed regardless where the user got the primitive instance names from.

> Using %s is a different convention than what is used in HPX and Phylanx and I do not like it because it may confuse users.

What API would you suggest instead?

> How would this help? Is the point that it would help differentiating between primitives from multiple expression trees? Would it guarantee that performance data from all primitives in an expression is available even if they are not instantiated?

These questions are mostly orthogonal to the issue at hand, aren't they? I'd be happy to discuss those independently.

hkaiser:
> Directly providing locality Id, primitive type and an enum for the performance counter type as arguments instead of a string. 

I don't think creating a separate enumerator type is a good solution. That would mean to encode the performance counter types in two separate ways, causing definite problems down the road.

> If you really don't like that then something like `/phylanx{locality#0}/primitives/count/eval` as a string and the name of the primitive as a separate argument

Isn't that essentially what I suggested?

> They'd used be used together if they both exist.

Sure, nobody objects to that. However as said, the function I proposed is needed regardless where the user got the primitive instance names from. So let's focus on this for now assuming that we somehow can retrieve the primitive instance names. Later we will solve the other half of the puzzle.

examples/profiling/primitive_instrumentation.cpp
```diff
@@ -0,0 +1,205 @@
...
+//                            to be queried from
+//   counter_name_last_parts: A vector containing the last part of the
+//                            performance counter names.
+//                            e.g. std::vector{ "count/eval", "count/direct_eval" }
+// NOTE: primitive_instances are not verified
+std::map<std::string, std::vector<std::int64_t>> retrieve_counter_data(
+    std::vector<std::string> const& primitive_instances,
+    hpx::naming::id_type locality_id,
+    std::vector<std::string>
+        counter_name_last_parts)
```
hkaiser: Passing the vector by `const&` might be sufficient - no need to create a copy.

examples/profiling/primitive_instrumentation.cpp
```diff
@@ -0,0 +1,205 @@
...
+            phylanx::execution_tree::compiler::parse_primitive_name(name);
+
+        // TODO: Ensure counter_name_last_part has at least one entry
+
+        // Performance counter values
+        std::vector<std::vector<std::int64_t>>& counter_values =
+            counter_values_pile[tags.primitive];
+        if (counter_values.empty())
+        {
+            // Iterate through the last parts of performance counter names
```
hkaiser: You could pre-allocate the memory for the needed number of entries: `counter_values.reserve(counter_name_last_parts.size())`

examples/profiling/primitive_instrumentation.cpp
```diff
@@ -0,0 +1,205 @@
...
+            for (auto const& counter_name_last_part : counter_name_last_parts)
+            {
+                // Construct the name of the counter
+                std::string counter_name("/phylanx/primitives/" +
+                    tags.primitive + "/" + counter_name_last_part);
+                // The actual performance counter
+                hpx::performance_counters::performance_counter counter(
+                    counter_name, locality_id);
+                counter_values.push_back(
```
hkaiser: You could request the counter values asynchronously:
```
        if (counter_values.empty())
        {
            // Iterate through the last parts of performance counter names
            std::vector<hpx::future<std::vector<counter_array_value>>> futures;
            futures.reserve(counter_name_last_parts.size());
            for (auto const& counter_name_last_part : counter_name_last_parts)
            {
                // Construct the name of the counter
                std::string counter_name("/phylanx/primitives/" +
                    tags.primitive + "/" + counter_name_last_part);
                // The actual performance counter
                hpx::performance_counters::performance_counter counter(
                    counter_name, locality_id);
                futures.push_back(counter.get_counter_values_array(false));
            }

            hpx::wait_all(futures());
            for (auto& f : futures)
            {
                counter_values.push_back(f.get().values));
            }
       }
```

hkaiser: You could even rewrite things such that you request all counters (for all primitives) first before calling `wait_all()`.

examples/profiling/primitive_instrumentation.cpp
```diff
@@ -0,0 +1,205 @@
...
+            }
+        }
+
+        std::vector<std::int64_t> data(counter_name_last_parts.size());
+        for (int i = 0; i < counter_values.size(); ++i)
+        {
+            data[i] = counter_values[i][tags.sequence_number];
+        }
+
+        result[name] = data;
```
hkaiser: This can be moved: `result[name] = std::move(data);`

hkaiser: Alternatively, using `emplace` might be even more efficient (as it avoids creating en empty vector before assigning the data): 
```
using maptype = std::map<std::string, std::vector<std::int64_t>>;
result.emplace(maptype::value_type(name, std::move(data)));
```

examples/profiling/primitive_instrumentation.cpp
```diff
@@ -0,0 +1,205 @@
...
+//                            is required
+//   locality_id:             The locality the performance counter data is going
+//                            to be queried from
+//   counter_name_last_parts: A vector containing the last part of the
+//                            performance counter names.
+//                            e.g. std::vector{ "count/eval", "count/direct_eval" }
+// NOTE: primitive_instances are not verified
+std::map<std::string, std::vector<std::int64_t>> retrieve_counter_data(
+    std::vector<std::string> const& primitive_instances,
+    hpx::naming::id_type locality_id,
```
hkaiser: The `hpx::id_type` could be passed by `const&`.

examples/profiling/primitive_instrumentation.cpp
```diff
@@ -0,0 +1,205 @@
...
+        << "count" << ","
+        << "time" << ","
+        << "direct_count" << ","
+        << "direct_time"
+        << std::endl;
+
+    // List of existing primitive instances
+    std::vector<std::string> existing_primitive_instances;
+    // Iterate over all primitive types
+    for (auto const& pattern :
```
hkaiser: Why do you iterate over all known patterns? Wouldn't it be sufficient to call `hpx::agas::find_symbols()` once? This should return all known primitive names that are registered at this point.

examples/profiling/primitive_instrumentation.cpp
```diff
@@ -0,0 +1,205 @@
...
+// Retrieve performance counter data for selected primitives
+//   primitive_instances:     The primitives for which performance counter data
+//                            is required
+//   locality_id:             The locality the performance counter data is going
+//                            to be queried from
+//   counter_name_last_parts: A vector containing the last part of the
+//                            performance counter names.
+//                            e.g. std::vector{ "count/eval", "count/direct_eval" }
+// NOTE: primitive_instances are not verified
+std::map<std::string, std::vector<std::int64_t>> retrieve_counter_data(
```
hkaiser: I'd suggest to move the locality-id to the last position as this would allow to leave it with a default value 
```
    ..., hpx::id_type const& locality_id = hpx::find_here())
```

examples/profiling/primitive_instrumentation.cpp
```diff
@@ -0,0 +1,205 @@
...
+        for (auto const& counter_value : entry.second)
+        {
+            std::cout << "," << counter_value;
+        }
+        std::cout << std::endl;
+    }
+
+    auto elapsed = t.elapsed();
+
+    // Make sure all counters are properly initialized, don't reset current counter values
```
hkaiser: Either you move this before retrieving the counter values or this can be removed.

hkaiser: I was referring to the invocation of `reinit_active_counters()` below, sorry for being insufficiently concise.

parsa: Fixed.

examples/profiling/primitive_instrumentation.cpp
parsa: Fixed.

examples/profiling/primitive_instrumentation.cpp
parsa: Fixed.

examples/profiling/primitive_instrumentation.cpp
parsa: It used `decltype(result)` instead of writing the full type. I hope that's okay.

examples/profiling/primitive_instrumentation.cpp
parsa: Fixed.

examples/profiling/primitive_instrumentation.cpp
parsa: Fixed.

examples/profiling/primitive_instrumentation.cpp
parsa: Fixed. Made other changes that need review. Also, I also added `/wd4552` to `Phylanx_SetupCompilerFlags.cmake` so that MSVC stops giving us this warning:
```
warning C4552: '*': operator has no effect; expected operator with side-effect
```

parsa: At some point a discrepancy between `function` and `call-function` was introduced and somehow all tests have been passing since but it broke my code. We need to somehow test for this.

examples/profiling/primitive_instrumentation.cpp
parsa: Done.

PR #191, opened 2018-02-03
 Unifying equal and equal_element_wise primitives

scheherzade:


hkaiser:
Thanks for preparing this PR that quickly! Here is the link to the inspect report causing the build to fail: https://934-101409977-gh.circle-artifacts.com/0/tmp/circle-artifacts.v9mSIh5/phylanx_inspect_report.html

src/execution_tree/primitives/equal.cpp
```diff
@@ -32,47 +32,114 @@ HPX_DEFINE_GET_COMPONENT_TYPE(equal_type::wrapped_type)
...
                 switch(rhs_dims)
                 {
                 case 0:
-                    return lhs.scalar() == rhs.scalar();
+                    return ir::node_data<bool>{lhs.scalar() == rhs.scalar()};
+
+                case 1:
+                    if (element_wise_)
+                        return equal0d1d(std::move(lhs), std::move(rhs));
+                    else
```
hkaiser: This `else` is not necessary as the `if` case always invokes `return`

src/execution_tree/primitives/equal.cpp
```diff
@@ -32,47 +32,114 @@ HPX_DEFINE_GET_COMPONENT_TYPE(equal_type::wrapped_type)
...
+                        HPX_THROW_EXCEPTION(hpx::bad_parameter,
+                            "equal::equal0d",
+                            "the operands have incompatible number of "
+                            "dimensions");
+
```
hkaiser: I'd suggest to add a `break;` statement in any case.

hkaiser:
We decided to merge the element-wise functionality into the existing primitives while removing the Boolean (non-element-wise) behavior. This can be closed without merging. The merging will be done in #187.

PR #200, opened 2018-02-07
Initial port of k-means algorithm to Phylanx

ct-clmsn:
Import k-means clustering algorithm ported from NumPy to Phylanx in Python. Implementation was ported from a BSD-licensed implementation found here:

* http://flothesof.github.io/k-means-numpy.html

### DIFF
--- a/examples/algorithms/kmeans.py
+++ b/examples/algorithms/kmeans.py
@@ -1,0 +1,53 @@
+#  Copyright (c) 2018 Steven R. Brandt
+#  Copyright (c) 2018 Christopher Taylor
+#
+#  Distributed under the Boost Software License, Version 1.0. (See accompanying
+#  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+#
+#  source code migrated from: http://flothesof.github.io/k-means-numpy.html
+#  original python code is BSD-licensed
+#
+import phylanx
+from phylanx.util import *
+
+@phyfun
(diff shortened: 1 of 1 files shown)


ct-clmsn:
numpy.newaxis is used to extend a type's dimensionality by a value of 1. additional supporting documentation:

https://stackoverflow.com/questions/29241056/how-does-numpy-newaxis-work-and-when-to-use-it
https://docs.scipy.org/doc/numpy/reference/arrays.indexing.html

ct-clmsn:
@hkaiser i'll start putting some effort into porting codes.

@parsa thank you for helping me format of this ticket!

ct-clmsn:
@parsa did you want me to start putting the primitives into the tracker?

hkaiser:
> 'copy': provide a copy primitive that can do the equivalent of a "Python deep copy" operation?

In PhySL this would be equivalent to creating a variable;

    define(deep_copy_of_x, x)

not sure how this should be expressed in Python in a generic way. @stevenrbrandt ?

ct-clmsn:
@hkaiser could a wrapper function called "copy" that is a short-hand python means of doing this operation be sufficient?

hkaiser:
> @hkaiser could a wrapper function called "copy" that is a short-hand python means of doing this operation be sufficient?

@ct-clmsn We could try hooking into Pythons standard shallow-copy/deep-copy mechanism (https://docs.python.org/3.8/library/copy.html). Not sure if this viable, though.

PR #201, opened 2018-02-07
Linear Matrix Primitive

rtohid:


hkaiser:
@rtohid the type `primitive_result_type` has been removed. In your code it has to be replaced by the equivalent `primitive_argument_type`.

PR #202, opened 2018-02-07
Slice1d

stevenrbrandt:
Adds support for slicing a 1D array. it also supports == comparing of Phylanx arrays with Numpy arrays.


hkaiser:
@stevenrbrandt inspect is not happy with this: https://990-101409977-gh.circle-artifacts.com/0/tmp/circle-artifacts.Xw91Dl9/phylanx_inspect_report.html

Issue #162, opened 2018-01-21
t-sne algorithm support


ct-clmsn:
decompose, analyze, and provide primitives supporting an implementation of the t-sne algorithm. a numpy implementation is provided [here](http://lvdmaaten.github.io/tsne/). This request is outside the original set of algorithms planned to study and support.

PR #204, opened 2018-02-08
Initial port of t-sne algorithm to Phylanx

ct-clmsn:
initial implementation of t-sne. issue #162 .

algorithm explained here:
https://nlml.github.io/in-raw-numpy/in-raw-numpy-t-sne/

numpy implementation here:
https://github.com/nlml/tsne_raw

### DIFF
--- a/examples/algorithms/tsne.py
+++ b/examples/algorithms/tsne.py
@@ -1,0 +1,169 @@
+#  Copyright (c) 2018 Steven R. Brandt
+#  Copyright (c) 2018 Christopher Taylor 
+#
+#  Distributed under the Boost Software License, Version 1.0. (See accompanying
+#  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+#
+#  Implementation ported from https://github.com/nlml/tsne_raw
+#  Adjoining blog post: https://nlml.github.io/in-raw-numpy/in-raw-numpy-t-sne/
+#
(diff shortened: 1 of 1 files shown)


examples/algorithms/tsne.py
```diff
@@ -0,0 +1,169 @@
...
+    return P
+
+@phyfun
+def estimate_sne(X, y, P, rng, num_iters, q_fn, grad_fn, learning_rate, momentum):
+    shape_x_arr = array(0, 2)
+    shape_x_arr[0] = shape(X, 0)
+    shape_x_arr[1] = 2
+
+    # TODO: https://docs.scipy.org/doc/numpy/reference/generated/numpy.random.RandomState.html
+    Y = random('normal',(0.0, 0.0001, shape_x_arr)
```
ct-clmsn: trying to emulate this capability: https://docs.scipy.org/doc/numpy/reference/generated/numpy.random.RandomState.html

hkaiser: FWIW, The existing `random()` primitive relies on balze's `rand()` generators which in turn rely in `std::uniform_int_distribution` and `std::uniform_real_distribution` of numbers generated by the standard Mersenne Twister (`mt19937`) generator. We will need to add our own facilities compatible/similar to those in blaze to be able to use other distributions (like 'normal'), see #207.

hkaiser: See also https://bitbucket.org/blaze-lib/blaze/issues/158/random-number-distributions.

PR #212, opened 2018-02-09
Improving interpreter

hkaiser:
This also adds the new `phylanx::ast::generate_transform_rules` API allowing to parse transform rules of the format `expr_to_match : expr_to_use_as_replacement`.


hkaiser:
The interpreter now understands a couple of command line options:
```
Usage: interpreter <physl_script> [options] [arguments...]:
  -h [ --help ]          print out program usage
  -c [ --code ] arg      Execute the PhySL code given in argument
  -p [ --print ]         Print the result of evaluation of the last PhySL
                         expression encountered in the input
  -t [ --transform ] arg file to read transformation rules from
```
The file specified to `--transform` should contain one or more transformation rules (`expr1 : expr2`) that will be applied to the input script before it is executed.

examples/interpreter/interpreter.cpp
```diff
@@ -57,38 +62,188 @@ std::vector<phylanx::execution_tree::primitive_argument_type> read_arguments(
...
+                ast, phylanx::ast::generate_transform_rules(transform_rules));
+        }
+
+        // Now compile AST into expression tree (into actual executable code)
+        phylanx::execution_tree::compiler::function_list snippets;
+        phylanx::execution_tree::compiler::environment env =
+            phylanx::execution_tree::compiler::default_environment();
+
+        phylanx::execution_tree::define_variable(
+            "sys_argv", snippets, env, std::move(args));
```
parsa: `args` is `const`. `std::move` of a `const` has no effect.

PR #217, opened 2018-02-14
Fixing Slicing

### LINKED ISSUE #188: slice(a, 0, -1, 0, -1) does not match Python semantics
slice(a, 0, -1, 0, -1) does not match Python semantics
```python
import numpy as np
from phylanx.util import *

@phyfun
def sliceme(a):
    return a[:-1,:-1]

v = np.arange(12)
v = np.reshape(v,(3,4))
print(v[:-1,:-1])
r = sliceme(v)
phy_print(r)
```

Output from the above
```
$ python3 x.py
[[0 1 2]
 [4 5 6]]
RuntimeError: col_stop/row_stop can not be negative if col_start/row_start is positive: HPX(bad_parameter)

During handling of the above exception, another exception occurred:

  File "x.py", line 11, in <module>
    r = sliceme(v)
  File "/home/sbrandt/.local/lib/python3.5/site-packages/phylanx-0.0.1-py3.5-linux-x86_64.egg/phylanx/util/__init__.py", line 322, in __call__
    return et.eval(self.__physl_src__,*nargs)
SystemError: <built-in method eval of PyCapsule object at 0x7fb1dc5558a0> returned a result with an error set
```
It should print the matrix twice (with slightly different formatting).

justwagle:
This pull request fixes the slicing primitives so that its more aligned with the python slicing.
This fixes #188


src/execution_tree/primitives/column_slicing.cpp
```diff
@@ -51,83 +52,97 @@ namespace phylanx { namespace execution_tree { namespace primitives
...
-                        "phylanx::execution_tree::primitives::"
-                            "column_slicing_operation::column_slicing_operation",
-                        "col_stop can not be negative if col_start is positive");
+                    result.push_back(i);
                 }
+            }
 
-                using storage0d_type = typename arg_type::storage0d_type;
-                using storage1d_type = typename arg_type::storage1d_type;
+            //Note: std::vector will throw Invalid element access index if
```
hkaiser: I'd suggest to explicitly throw an exception here with an appropriate error message instead of relying for the vector to diagnose out-of-bounds indexing.

PR #235, opened 2018-02-21
Update Phylanx Python Package

rtohid:
* Restructure Python package.
    - move transformation functionalities to module transformation under ast.
    - move AST utility functions to utilities under ast.

* Change the phyfun class to PhyTransformer and refactor the implementation to avoid elifs when possible.


stevenrbrandt:
I've looked over the code for a while and everything seems fine. Is this supposed to be able to generate something other than PhySL, @rtohid ? I seem to recall that you said so at the meeting.

rtohid:
Thank you @stevenrbrandt .
Now you can pass the target to the Phylanx decorator- please see the last commit I just added.

stevenrbrandt:
This code breaks the test suite in a few important ways. Still trying to figure it out.

rtohid:
Type of the decorated function is converted to the decorator if the decorator does not take any arguments, otherwise, it remains a function. We can make a change so, optionally, the PhySL code is also returned- not as an attribute. @stevenrbrandt Please let me know if you can think of a better way. Thank you.

stevenrbrandt:
I added another simple test to eval.py which, surprisingly, breaks things. Apparently, (1) we need more tests and (2) some things got lost in the refactoring.

hkaiser:
@rtohid  Is this ready to move forward now?

stevenrbrandt:
OK, I approve this pull request. @hkaiser ? What say you?

hkaiser:
@stevenrbrandt If you are happy - I'm happy too!

rtohid:
@hkaiser. If tests pass LGTM.

PR #240, opened 2018-02-22
Hdf5 linux

hapoo:
This pull request fixes build issue with HighFive on Linux systems and adds HDF5 builds and test to CircleCI tests.

### DIFF
--- a/cmake/Phylanx_HighFive.cmake
+++ b/cmake/Phylanx_HighFive.cmake
@@ -7,7 +7,7 @@
 macro(phylanx_setup_hdf5)
     if(PHYLANX_WITH_HIGHFIVE)
 
-        find_package(HDF5 NO_MODULE NO_CMAKE_PACKAGE_REGISTRY)
+        find_package(HDF5)
         if(NOT HDF5_FOUND)
             phylanx_warn("The HDF5 libraries could not be found, please set HDF5_DIR to help locating it.")
             set(PHYLANX_WITH_HIGHFIVE OFF)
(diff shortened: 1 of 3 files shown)


cmake/Phylanx_HighFive.cmake
```diff
@@ -7,7 +7,7 @@
 macro(phylanx_setup_hdf5)
     if(PHYLANX_WITH_HIGHFIVE)
 
-        find_package(HDF5 NO_MODULE NO_CMAKE_PACKAGE_REGISTRY)
+        find_package(HDF5)
```
hkaiser: Do we really have to remove the `NO_CMAKE_PACKAGE_REGISTRY` option?

hapoo: Yes, `NO_CMAKE_PACKAGE_REGISTRY` the `find_package()` fails on my machine

hkaiser: ok, thanks

tools/docker/Dockerfile
```diff
@@ -9,6 +9,8 @@ RUN apt-get update  &&
     apt-get -y install wget &&                                                   \
     apt-get -y install python3 python3-dev python3-pip &&                        \
     apt-get -y install libblas-dev liblapack-dev &&                              \
+    apt-get -y install libhdf5-dev &&                                            \
+    apt-get -y install libboost-test-dev &&                                      \
```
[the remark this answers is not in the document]
hapoo: HighFive's cmake complains about missing `boost_unit_test_framework`
https://circleci.com/gh/STEllAR-GROUP/phylanx/1161
Let me see if we can avoid that test. Since HighFive is header-only library we don't need the test to install HighFive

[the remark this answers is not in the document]
hkaiser: Ok, understood. If we end up needed this Boost library I'd suggest adding it to the base docker file to avoid Boost version mismatches.

PR #241, opened 2018-02-22
Set operation

justwagle:
This pull operation adds the ability to select a region in an input matrix/vector and set values to that region. Adds three new primitives viz. set, row_set and column_set.


src/execution_tree/primitives/column_set.cpp
```diff
@@ -0,0 +1,360 @@
...
+                auto init_list = create_list_col_set(
+                    col_start, col_stop, step, args[0].size());
+
+                auto sv = blaze::elements(args[0].vector(), init_list);
+
+                if (value_dimnum == 0)
+                {
+                    blaze::DynamicVector<double> temp(
+                        sv.size(), args[4].scalar());
+                    sv = temp;
```
hkaiser: Can this use `std::move`?

src/execution_tree/primitives/column_set.cpp
```diff
@@ -0,0 +1,360 @@
...
+                {
+                    HPX_THROW_EXCEPTION(hpx::bad_parameter,
+                        "phylanx::execution_tree::primitives::"
+                        "column_set_operation::column_set1d",
+                        generate_error_message("size mismatch, please check "
+                                               "your parameters or set vector",
+                            name_, codename_));
+                }
+
+                sv = temp;
```
hkaiser: Can this use `std::move`?

src/execution_tree/primitives/column_set.cpp
```diff
@@ -0,0 +1,360 @@
...
+                auto init_list_col = create_list_col_set(
+                    col_start, col_stop, step_col, num_matrix_cols);
+
+                auto sm = blaze::columns(args[0].matrix(), init_list_col);
+
+                if (value_dimnum == 0)
+                {
+                    blaze::DynamicMatrix<double> data(
+                        sm.rows(), sm.columns(), args[4].scalar());
+                    sm = data;
```
hkaiser: Can this use `std::move`? Also, more of this below...

src/execution_tree/primitives/row_set.cpp
```diff
@@ -0,0 +1,360 @@
...
+                if (stop < 0)
+                {
+                    actual_stop = array_length + stop;
+                }
+
+                std::vector<int> result;
+
+                if (step > 0)
+                {
+                    for (int i = actual_start; i < actual_stop; i += step)
```
hkaiser: Would it be possible to `reserve()` the appropriate amount of entries for this vector? Something like `result.reserve((actual_stop-actual_start+step)/step);`?

src/execution_tree/primitives/row_set.cpp
```diff
@@ -0,0 +1,360 @@
...
+                    {
+                        result.push_back(i);
+                    }
+                }
+
+                if (step < 0)
```
hkaiser: This could be `else if(...)`.

src/execution_tree/primitives/row_set.cpp
```diff
@@ -0,0 +1,360 @@
...
+                return primitive_argument_type{ir::node_data<double>(0)};
+            }
+
+            primitive_argument_type row_set2d(args_type&& args) const
+            {
+                auto row_start = args[1].scalar();
+                auto row_stop = args[2].scalar();
+                auto step_row = args[3].scalar();
+                auto col_start = 0;
+                auto col_stop = args[0].dimensions()[1];
```
hkaiser: `col_start` and `col_stop` will have different types, is that ok?

src/execution_tree/primitives/row_set.cpp
```diff
@@ -0,0 +1,360 @@
...
+                    blaze::DynamicMatrix<double> data(
+                        sm.rows(), sm.columns(), args[4].scalar());
+                    sm = data;
+                    return primitive_argument_type{ir::node_data<double>(0)};
+                }
+
+                if (value_dimnum == 1)
+                {
+                    auto data = blaze::trans(args[4].vector());
+                    std::size_t data_size = data.size();
```
hkaiser: Sometimes you use `auto` and sometimes a concrete type, what's the rationale?

src/execution_tree/primitives/row_set.cpp
```diff
@@ -0,0 +1,360 @@
...
+                    HPX_THROW_EXCEPTION(hpx::bad_parameter,
+                        "phylanx::execution_tree::primitives::"
+                        "row_set_operation::row_set1d",
+                        generate_error_message("size mismatch, please check "
+                                               "your parameters or set vector",
+                            name_, codename_));
+                }
+
+                sv = temp;
+                return primitive_argument_type{ir::node_data<double>(0)};
```
hkaiser: What is the primitive supposed to 'return'? Currently it always returns zero...

src/execution_tree/primitives/row_set.cpp
justwagle: Actually they were unused variables and I removed them from the code.

src/execution_tree/primitives/column_set.cpp
justwagle: As per our discussion, we decided to not use std::move in this case and leave this as is.

src/execution_tree/primitives/column_set.cpp
```diff
@@ -0,0 +1,344 @@
...
+                else //(stop < 0)
+                {
+                    actual_stop = array_length + stop;
+                }
+
+                std::vector<int> result;
+
+                if (step > 0)
+                {
+                    result.reserve((actual_stop - actual_start + step)/step);
```
hkaiser: Careful! It could be that `actual_stop < actual_start`.

src/execution_tree/primitives/row_set.cpp
```diff
@@ -0,0 +1,341 @@
...
+                    generate_error_message(
+                        "use store operation for setting value to a variable",
+                        name_, codename_));
+            }
+
+            primitive_argument_type row_set1d(args_type&& args) const
+            {
+                auto row_start = args[1].scalar();
+                auto row_stop = args[2].scalar();
+                int step = args[3].scalar();
```
hkaiser: Usually `int` can't fully represent the value range of `double`, so this might go wrong.

src/execution_tree/primitives/row_set.cpp
```diff
@@ -0,0 +1,341 @@
...
+                    "row_set_operation::row_set_0d",
+                    generate_error_message(
+                        "use store operation for setting value to a variable",
+                        name_, codename_));
+            }
+
+            primitive_argument_type row_set1d(args_type&& args) const
+            {
+                auto row_start = args[1].scalar();
+                auto row_stop = args[2].scalar();
```
hkaiser: `row_start` and `row_stop` will be of type `double` - is that what you want?

src/execution_tree/primitives/column_set.cpp
```diff
@@ -94,25 +94,30 @@ namespace phylanx { namespace execution_tree { namespace primitives
...
                 }
 
-                std::vector<int> result;
+                std::vector<std::int64_t> result;
 
                 if (step > 0)
                 {
-                    result.reserve((actual_stop - actual_start + step)/step);
-                    for (int i = actual_start; i < actual_stop; i += step)
+                    HPX_ASSERT(actual_stop > actual_start);
```
hkaiser: Good move! However, is it ensured that this pre-condition is met? Do you check anywhere (and throw an exception), if not?

justwagle: @hkaiser I was hoping the assert would be enough to let the user know that something is wrong with his parameter selection. One thing I could do is convert the assert into an exception. What would you suggest ?

hkaiser: The difference between an assert and an exception is that assert's are used to make sure pre-conditions are _met_. Exceptions are used to _report_ problems (for instance pre-condition violations). As a guideline, use exceptions in user-facing code, and assertions in code that is used internally only (i.e. not directly user-facing, assuming that the checks have been performed elsewhere). In your case I'd use an assertion in the function (as you have it) but add an explicit parameter check on some level above it.

PR #253, opened 2018-02-27
Argmin & Argmax Primitives

parsa:
This PR adds Argmin/Argmax primitives mentioned in #200.

It also includes iterators that allow iterating over `Row<T>`s and `Column<T>`s in a `Blaze::Matrix<T>`.

### DIFF
--- a/phylanx/execution_tree/primitives/argmax.hpp
+++ b/phylanx/execution_tree/primitives/argmax.hpp
@@ -18,10 +18,12 @@
 
 namespace phylanx { namespace execution_tree { namespace primitives
 {
-    /// \brief Implementation of argmax as a Phylanx primitive
-    /// This implementation is intended to behave like [NumPy implementation
-    /// of argmax]
+    /// \brief Implementation of argmax as a Phylanx primitive.
+    /// Returns the index of the largest element of argument a.
+    /// This implementation is intended to behave like [NumPy implementation of argmax]
     /// (https://docs.scipy.org/doc/numpy-1.14.0/reference/generated/numpy.argmax.html).
+    /// \param a It may be a scalar value, vector, or matrix
+    /// \param axis The dimension along which the indices of the max value(s) should be found
     class argmax
         : public primitive_component_base
         , public std::enable_shared_from_this<argmax>
@@ -40,8 +42,6 @@ namespace phylanx { namespace execution_tree { namespace primitives
 
         argmax() = default;
 
-        /// \brief Calculates argmax of a node_data
-        /// \param args A scalar value, vector, or matrix
         argmax(std::vector<primitive_argument_type>&& operands,
             std::string const& name, std::string const& codename);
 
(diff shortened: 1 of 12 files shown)


src/execution_tree/primitives/argmax.cpp
```diff
@@ -0,0 +1,323 @@
...
+
+                const matrix_row_iterator<decltype(a)> a_begin(a);
+                const matrix_row_iterator<decltype(a)> a_end(a, a.rows());
+
+                double global_max = 0.;
+                std::size_t global_index = 0ul;
+                std::size_t passed_rows = 0ul;
+                for (auto it = a_begin; it != a_end; ++it, ++passed_rows)
+                {
+                    const auto local_max = std::max_element(it->begin(), it->end());
```
hkaiser: Could it happen that `it->begin() == it->end()`? In this case the returned iterator shouldn't be dereferenced.

src/execution_tree/primitives/argmin.cpp
```diff
@@ -0,0 +1,280 @@
...
+            using args_type = std::vector<arg_type>;
+
+            primitive_argument_type argmin0d(args_type && args) const
+            {
+                return 0ul;
+            }
+
+            ///////////////////////////////////////////////////////////////////////////
+            primitive_argument_type argmin1d(args_type && args) const
+            {
```
hkaiser: Don't we need to handle the axis here (same as argmax)?

phylanx/execution_tree/primitives/argmax.hpp
```diff
@@ -0,0 +1,40 @@
...
+namespace phylanx { namespace execution_tree { namespace primitives
+{
+    class argmax : public primitive_component_base
+    {
+    public:
+        static match_pattern_type const match_data;
+
+        argmax() = default;
+
+        argmax(std::vector<primitive_argument_type>&& operands,
```
hkaiser: Could you add some documentation here, please?

src/execution_tree/primitives/argmax.cpp
parsa: No. Because the row count is checked in line 129-136

hkaiser: That check tests for whether there are rows, while I was asking about the number of elements in  row.

src/execution_tree/primitives/argmin.cpp
parsa: Done.

phylanx/execution_tree/primitives/argmax.hpp
parsa: Please see the new commit.

src/execution_tree/primitives/argmax.cpp
parsa: I don't know if that can happen. I added a bit that checks the column count for good measure.

PR #257, opened 2018-02-27
Refactor primitives to new style

parsa:
This PR refactors 10 of the primitives to reflect the structure updates.


src/execution_tree/primitives/block_operation.cpp
```diff
@@ -20,7 +20,78 @@
 ///////////////////////////////////////////////////////////////////////////////
 namespace phylanx { namespace execution_tree { namespace primitives
 {
+    struct block_operation::step : std::enable_shared_from_this<step>
```
hkaiser: This is not how it is supposed to be refactored. In the end the block_operation itself has to be derived from `std::enable_shared_from_this<>`, see your refactoring of the `add_primitive`. The whole point of this refactoring is to avoid having to copy `operands`, `args`, `codename`, and `name` for each of the evaluations on each primitive. If the refactoring is done as proposed here this goal will not be achieved.

phylanx/execution_tree/primitives/define_function.hpp
```diff
@@ -23,8 +24,15 @@ namespace phylanx { namespace execution_tree { namespace primitives
     //
     // This is a helper primitive needed for proper binding of the expression
     // value to a variable.
-    class define_function : public primitive_component_base
+    class define_function
+        : public primitive_component_base
+        , public std::enable_shared_from_this<define_function>
```
hkaiser: This is not needed as the `define_function` primitive never has to call `shared_from_this()`.

phylanx/execution_tree/primitives/define_function.hpp
```diff
@@ -23,8 +24,15 @@ namespace phylanx { namespace execution_tree { namespace primitives
...
     {
+    protected:
+        hpx::future<primitive_argument_type> eval(
```
hkaiser: This function is left undefined, is it needed?

phylanx/execution_tree/primitives/div_operation.hpp
```diff
@@ -29,6 +41,26 @@ namespace phylanx { namespace execution_tree { namespace primitives
...
+        primitive_argument_type div0d(operands_type && ops) const;
+        primitive_argument_type div1d0d(operands_type && ops) const;
+        primitive_argument_type div1d1d(operands_type && ops) const;
+        primitive_argument_type div1d2d(operands_type&& ops) const;
+        primitive_argument_type div1d(operands_type && ops) const;
+        primitive_argument_type div2d0d(operands_type && ops) const;
+        primitive_argument_type div2d1d(operands_type&& ops) const;
+        primitive_argument_type div2d2d(operands_type && ops) const;
+        primitive_argument_type div2d(operands_type && ops) const;
+
```
hkaiser: Why this whitespace?

phylanx/execution_tree/primitives/define_function.hpp
parsa: Removed.

phylanx/execution_tree/primitives/define_function.hpp
parsa: Removed.

src/execution_tree/primitives/block_operation.cpp
parsa: Oops.. It should be fixed in the new commit.

PR #259, opened 2018-02-28
Any and All primitives 

scheherzade:
And and all primitives check if any or all the elements of a vector or matrix are evaluated to true.

### DIFF
--- a/src/execution_tree/primitives/all_operation.cpp
+++ b/src/execution_tree/primitives/all_operation.cpp
@@ -1,0 +1,130 @@
+//  Copyright (c) 2018 Shahrzad Shirzad
+//
+//  Distributed under the Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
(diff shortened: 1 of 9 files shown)


src/execution_tree/primitives/all_operation.cpp
```diff
@@ -0,0 +1,130 @@
...
+                    default:
+                        HPX_THROW_EXCEPTION(hpx::bad_parameter,
+                            "all_operation::eval",
+                            execution_tree::generate_error_message(
+                                "operand has unsupported "
+                                "number of dimensions",
+                                this_->name_, this_->codename_));
+                    }
+                }),
+            detail::map_operands(operands, functional::numeric_operand{}, args,
```
hkaiser: Shouldn't we support Boolean operands (i.e. vectors and matrices in a `node_data<bool>`) here as well?

scheherzade:
@hkaiser would the new commit work for this purpose?

hkaiser:
> would the new commit work for this purpose?

It's perfect! Thanks!

PR #260, opened 2018-02-28
Symbol Introspection Facilities for the Python Front-end

rtohid:
- Return details of all symbols, including the called primitives.
- Return the actual line number in the Python file instead of the line number within the function.


python/phylanx/ast/transformation.py
```diff
@@ -423,19 +425,23 @@ def __init__(self, f):
...
+            transormation = self.targets[target]()
 
             assert len(tree.body) == 1
 
-            self.__physl_src__ = '%s(%s)\n' % (
-                full_node_name(tree.body[0], 'block'), physl.recompile(tree))
+            if target == "PhySL":
+                self.__physl_src__ = '%s(%s)\n' % (
+                    full_node_name(tree.body[0], 'block'),
+                    transormation.recompile(tree))
```
hkaiser: What should happen if `target != "PhySL"`?

python/phylanx/ast/transformation.py
```diff
@@ -423,19 +425,23 @@ def __init__(self, f):
...
             src = inspect.getsource(f)
             # Before recompiling the code, take
             # off the decorator from the source.
             src = re.sub(r'^\s*@\w+.*\n', '', src)
 
             # Create the AST
             tree = ast.parse(src)
-            physl = PhySL()
+            ast.increment_lineno(tree, actual_lineno)
+            transormation = self.targets[target]()
```
hkaiser: Typo: `transformation`

python/phylanx/ast/transformation.py
```diff
@@ -232,7 +233,8 @@ def _AugAssign(self, a):
         nn = args[1].__class__.__name__
         if nn == "Add":
             sym = "+"
-        return "store(" + args[0].id + "," + args[0].id + sym + self.recompile(
+        arg0 = '%s' % full_node_name(args[0], args[0].id)
+        return "store(" + arg0 + "," + arg0 + sym + self.recompile(
```
hkaiser: The `store()` primitive should be annotated as well.

hkaiser: There are more primitive names without annotation in this file ([`slice_column`](https://github.com/STEllAR-GROUP/phylanx/blob/21e0468e6160494f83d1c73615cdfae8ebe951dd/python/phylanx/ast/transformation.py#L120), [`slice`](https://github.com/STEllAR-GROUP/phylanx/blob/21e0468e6160494f83d1c73615cdfae8ebe951dd/python/phylanx/ast/transformation.py#L92), [`shape`](https://github.com/STEllAR-GROUP/phylanx/blob/21e0468e6160494f83d1c73615cdfae8ebe951dd/python/phylanx/ast/transformation.py#L92), etc.)

PR #263, opened 2018-03-01
Shuffle Primitive

parsa:
This PR adds the shuffle primitive to Phylanx, which is the equivalent of `np.random.shuffle`. This primitive was mentioned in #200.

Also:
* Fixes a bug in variable primitive where unwanted copies were made during the assignment of a PhySL variable from PhySL Rvalue (`DynamicMatrix`)
* Updates matrix row & column iterator to make their instances copyable

### DIFF
--- a/phylanx/util/matrix_iterators.hpp
+++ b/phylanx/util/matrix_iterators.hpp
@@ -7,6 +7,7 @@
 #if !defined(PHYLANX_MATRIX_ITERATORS)
 #define PHYLANX_MATRIX_ITERATORS
 
+#include <algorithm>
 #include <cstddef>
 #include <utility>
 
--- a/phylanx/util/matrix_iterators.hpp
+++ b/phylanx/util/matrix_iterators.hpp
@@ -50,6 +50,10 @@ namespace phylanx { namespace util
         {
             return blaze::row(data_, index_);
         }
+        std::ptrdiff_t distance_to(matrix_row_iterator const& other) const
+        {
+            return other.index_ - index_;
+        }
 
     private:
         T & data_;
@@ -93,6 +97,10 @@ namespace phylanx { namespace util
         blaze::Column<T> dereference() const
         {
             return blaze::column(data_, index_);
+        }
+        std::ptrdiff_t distance_to(matrix_column_iterator const& other) const
+        {
+            return other.index_ - index_;
         }
 
     private:
--- a/phylanx/util/matrix_iterators.hpp
+++ b/phylanx/util/matrix_iterators.hpp
@@ -1,17 +1,45 @@
-//  Copyright (c) 2018 Parsa Amini
-//  Copyright (c) 2018 Hartmut Kaiser
+// Copyright (c) 2018 Parsa Amini
+// Copyright (c) 2018 Hartmut Kaiser
 //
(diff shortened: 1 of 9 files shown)


src/execution_tree/primitives/shuffle_operation.cpp
```diff
@@ -0,0 +1,132 @@
...
+
+        return primitive_argument_type(std::move(x));
+    }
+
+    primitive_argument_type shuffle_operation::shuffle_2d(args_type&& args) const
+    {
+        auto x = args[0].matrix();
+        auto x_begin = util::matrix_row_iterator<decltype(x)>(x);
+        auto x_end = util::matrix_row_iterator<decltype(x)>(x, x.rows());
+        std::shuffle(x_begin, x_end, std::mt19937{ std::random_device{}() });
```
hkaiser: Would you see a way not to create a new generator and not calling `std::random_device` for each invocation of this primitive?

phylanx/util/matrix_iterators.hpp
```diff
@@ -7,6 +7,7 @@
 #if !defined(PHYLANX_MATRIX_ITERATORS)
 #define PHYLANX_MATRIX_ITERATORS
 
+#include <algorithm>
```
hkaiser: What do you need this #include for?

src/execution_tree/primitives/shuffle_operation.cpp
```diff
@@ -44,6 +44,8 @@ namespace phylanx { namespace execution_tree { namespace primitives
     };
 
     ///////////////////////////////////////////////////////////////////////////
+    std::mt19937 shuffle_operation::rand_machine{std::random_device{}()};
```
hkaiser: You're missing `#include <random>` for this file.

phylanx/util/matrix_iterators.hpp
```diff
@@ -1,17 +1,46 @@
...
+    //         caused by matrix_row_iterator not being a real random access
+    //         iterator. Dereferencing matrix_row_iterator does not yield a
+    //         true reference but only a temporary blaze::Row holding true
+    //         references.
+    //
+    // A real fix for this problem is proposed in PR0022R0
+    // (http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2015/p0022r0.html)
+    //
+    template <typename T>
+    HPX_FORCEINLINE
```
hkaiser: I think this should not be added to the iterators, it's merely a facility that should be visible to the shuffle primitive (i.e. should be placed in shuffle_primitive.cpp).

phylanx/util/matrix_iterators.hpp
parsa: I don't need it. Inspect was complaining.

src/execution_tree/primitives/shuffle_operation.cpp
parsa: It's in the header file.

src/execution_tree/primitives/shuffle_operation.cpp
parsa: Done.

phylanx/util/matrix_iterators.hpp
parsa: I think I got it right. Please check the updated code.

PR #280, opened 2018-03-06
Mixing of dimensions in hstack and vstack 

### LINKED ISSUE #252: Add mixing arguments of different dimensionality in hstack and vstack
Hstack and vstack needs to be extended to add support for Add mixing arguments of different dimensionality.

justwagle:
`make_vector` primitive is removed and regression tests using `make_vector` is replaced with `hstack`. This PR fixes #252.


tests/unit/python/execution_tree/primitives/make_vector.py
```diff
@@ -1,45 +0,0 @@
-#  Copyright (c) 2018 Hartmut Kaiser
```
hkaiser: Would you mind re-adding this test, just changed for `hstack()`?

justwagle: @hkaiser yes, I can do that but  I just realized if we remove make vector, we will not be able to create a vector of single element as hstack requires at least two elements. 
``` The following test cases will not pass : 
@Phylanx("PhySL")
def test_make_vector_one():
    return make_vector(42)
```

 I could update hstack so that if only one argument is passed to it, it will return a vector containing the single element. But then in python 
```
np.hstack(5)
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.6/site-packages/numpy/core/shape_base.py", line 283, in hstack
    arrs = [atleast_1d(_m) for _m in tup]
TypeError: 'int' object is not iterable
```
So we will not be doing what python does. What should I do here ? Should we keep make_vector as well in addtion to hstack ?

hkaiser: Shouldn't that be resolved once we fix #295? And yes, I think hstack/vstack with one element should return the element itself (possibly wrapped into a vector if it's a single double).

justwagle: @hkaiser As far as I can see, #295 should not have any impact on this. If we add the ability in hstack/vstack with one element to return the element itself wrapped in a vector, we should have all the cases covered. I will go ahead and modify hstack and vstack and also complete adding make_vector python test back. Thank you.

PR #282, opened 2018-03-06
Print Newick formatted trees in the interpreter

justwagle:
This pull request allows the physl interpreter to print  Newick-formatted trees along with the currently printed dot formatted tree.   Newick-formatted  tree is required by katy's visualization tool.

### DIFF
--- a/examples/interpreter/physl.cpp
+++ b/examples/interpreter/physl.cpp
@@ -237,6 +237,12 @@ int main(int argc, char* argv[])
                 << phylanx::execution_tree::dot_tree(code_source_name,
                        snippets.snippets_.back().get_expression_topology())
                 << "\n";
+
+            std::cout << "\n"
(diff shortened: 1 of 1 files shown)


examples/interpreter/physl.cpp
```diff
@@ -237,6 +237,12 @@ int main(int argc, char* argv[])
                 << phylanx::execution_tree::dot_tree(code_source_name,
                        snippets.snippets_.back().get_expression_topology())
                 << "\n";
+
+            std::cout << "\n"
+                << phylanx::execution_tree::newick_tree(code_source_name,
+                       snippets.snippets_.back().get_expression_topology())
+                << "\n\n";
+
```
hkaiser: Would you see a way to extract the expression-topology only once and reuse it for the printing operations?

PR #286, opened 2018-03-07
Fix docker images

parsa:
Our current docker image does not actually contain Phylanx, but just HPX and Phylanx's prerequisites. This PR resolves this issue by installing Phylanx on a Docker container and committing the changes before deploying the image.


.circleci/config.yml
```diff
@@ -8,7 +8,7 @@ jobs:
     build:
         machine: true
         environment:
-            IMAGE_NAME: stellargroup/phylanx:dev
+            IMAGE_NAME: stellargroup/phylanx
```
hkaiser: What's the rationale to change this name?

.circleci/config.yml
```diff
@@ -29,7 +29,7 @@ jobs:
             # CMake
             - run:
                 name: Run CMake
-                command: docker run -v $PWD:/phylanx -w /phylanx/build -e "CIRCLECI=true" ${IMAGE_NAME} cmake -DPHYLANX_WITH_GIT_COMMIT=${CIRCLE_SHA1} -DPHYLANX_WITH_TOOLS=On -DHPX_DIR=/usr/local/lib/cmake/HPX -Dpybind11_DIR=/pybind11/share/cmake/pybind11/ -Dblaze_DIR=/blaze/share/blaze/cmake -DCMAKE_BUILD_TYPE=Debug -DCMAKE_CXX_COMPILER=clang++ -DCMAKE_C_COMPILER=clang -DPHYLANX_WITH_HIGHFIVE=On -DHighFive_DIR=/highfive/share/HighFive/CMake ..
+                command: docker run -v $PWD:/phylanx -w /phylanx/build -e "CIRCLECI=true" ${IMAGE_NAME} cmake -DPHYLANX_WITH_GIT_COMMIT=${CIRCLE_SHA1} -DPHYLANX_WITH_TOOLS=On -DHPX_DIR=/usr/local/lib/cmake/HPX -Dblaze_DIR=/blaze/share/blaze/cmake -DCMAKE_BUILD_TYPE=Debug -DCMAKE_CXX_COMPILER=clang++ -DCMAKE_C_COMPILER=clang -DPHYLANX_WITH_HIGHFIVE=On ..
```
hkaiser: Why don't we need the Pybind11 and HighFive directories anymore?

parsa: Because there is no justification for using custom paths. We wouldn't even be installing from source if there were Debian package available which would install in `/usr`.

.circleci/config.yml
parsa: I separated the prerequisites image and the current working master.

PR #287, opened 2018-03-07
Fix jupyter io

stevenrbrandt:
Using Python's print routine for I/O redirection is problematic. In principle, one probably needs the GIL to do it. This branch uses linux pipes to redirect output (but only if we're running inside a notebook).

### DIFF
--- a/python/phylanx/ast/transformation.py
+++ b/python/phylanx/ast/transformation.py
@@ -450,6 +450,7 @@ def convert_to_phylanx_type(v):
         pass
     return v
 
+
 # Determine whether we are
 # running in a notebook or
 # not. If we are, use the
@@ -457,7 +458,7 @@ def convert_to_phylanx_type(v):
 try:
     get_ipython()
     iod = IORedirecter()
-except:
+except NameError:
     iod = DoNothing()
 
 
--- a/python/phylanx/ast/transformation.py
+++ b/python/phylanx/ast/transformation.py
@@ -5,38 +5,46 @@
 # Distributed under the Boost Software License, Version 1.0. (See accompanying
 # file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
 
+import ast
+import re
+import os
 import phylanx
 import inspect
-from .utils import *
+from .utils import full_name, full_node_name
+from threading import Thread
 
 et = phylanx.execution_tree
 
-import ast,re,os
-from threading import Thread
 
 def readfds(fd):
     while True:
-        result=os.read(fd,1000)
-        print(result.decode('utf-8'),end='')
+        result = os.read(fd, 1000)
+        print(result.decode('utf-8'), end='')
+
 
 class IORedirecter:
     def __init__(self):
         self.fds = os.pipe()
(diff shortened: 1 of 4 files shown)


python/src/phylanx.cpp
```diff
@@ -11,6 +11,9 @@
 #include <bindings/binding_helpers.hpp>
 
 #include <pybind11/pybind11.h>
+#include <pybind11/iostream.h>
+#include <pybind11/operators.h>
+#include <pybind11/stl.h>
```
hkaiser: Why are these #include's necessary for this file?

hkaiser:
Could you please explain why using the GIL would be problematic? While your solution might work it looks to be too complicated for a simple thing as getting output redirected to stdout.

Please provide some rationale for your design decision.

python/phylanx/ast/transformation.py
```diff
@@ -320,10 +345,23 @@ def _UnaryOp(self, a):
...
+        sep = ""
+        for arg in ast.iter_child_nodes(a):
+            if arg.__class__.__name__ == "Load":
+                break
+            s += sep
+            s += self.recompile(arg)
+            sep = ","
+        s += ")"
+        return s
+
```
hkaiser: Is this change unrelated?

python/phylanx/ast/transformation.py
```diff
@@ -451,6 +500,7 @@ def __init__(self, f):
 
         def __call__(self, *args):
             nargs = tuple(convert_to_phylanx_type(a) for a in args)
-            return et.eval(self.__physl_src__, *nargs)
+            with iod:
+                return et.eval(self.__physl_src__, *nargs)
```
hkaiser: Wouldn't this disable evaluating things anywhere except in Jupyter notebooks?

python/phylanx/ast/transformation.py
```diff
@@ -418,6 +458,15 @@ def convert_to_phylanx_type(v):
         pass
     return v
 
+# Determine whether we are
+# running in a notebook or
+# not. If we are, use the
+# IORedirector.
+try:
+    get_ipython()
```
hkaiser: Does `get_ipython` really check whether we're running in a notebook. Wouldn't this call succeed in any IPython environment?

python/src/phylanx.cpp
stevenrbrandt: Sorry, let me clean up the unrelated stuff.

python/phylanx/ast/transformation.py
hkaiser: Ahh sorry, I misread the code. Please disregard my comment.

python/phylanx/ast/transformation.py
stevenrbrandt: Yes, it will also succeed in an IPython environment, but not from the regular command line.

hkaiser: Would this be a problem?

stevenrbrandt: It means IPython will use the IO redirection, which it doesn't need to do. I suppose I could check os.environ[_] and look for the string "jupyter" and/or "ipython". The latter feels more hacky to me.

PR #296, opened 2018-03-08
Physl pretty print

stevenrbrandt:
The code generated by the Phylanx decorator can be hard to read. The physl_fmt() routine makes the code easier to look at.


python/phylanx/ast/transformation.py
```diff
@@ -7,13 +7,45 @@
...
+
+
+class DoNothing:
+    def __enter__(self):
+        pass
+
+    def __exit__(self, type, value, traceback):
+        pass
+
+
```
hkaiser: Is this unrelated?

PR #298, opened 2018-03-08
Use prior phylanx defs

stevenrbrandt:
This change makes it possible to use one Phylanx defined function inside another. This is especially important for Jupyter.


python/src/bindings/binding_helpers.hpp
```diff
@@ -117,16 +117,24 @@ namespace phylanx { namespace bindings
...
     {
         namespace et = phylanx::execution_tree;
         return hpx::threads::run_as_hpx_thread(
             [&]() -> et::primitive_argument_type
             {
                 try
                 {
-                    et::compiler::function_list eval_snippets;
-                    auto x = et::compile(xexpr, eval_snippets);
+                    static phylanx::execution_tree::compiler::environment *eval_env = nullptr;
```
hkaiser: Why pointers? Where is the allocated memory free'd?

python/phylanx/ast/transformation.py
```diff
@@ -7,13 +7,45 @@
...
+
+
+class DoNothing:
+    def __enter__(self):
+        pass
+
+    def __exit__(self, type, value, traceback):
+        pass
+
+
```
hkaiser: Is this change related?

hkaiser:
Could you please add a test for this new functionality?

hkaiser:
@stevenrbrandt are we sure that we will not need to separate several different 'translation units'? The way this is written now everything is dumped into the same 'namespace'. 

I think we should make this more modular, for instance compiling each Python module into it's own PhySL environment.

python/src/bindings/binding_helpers.hpp
stevenrbrandt: It is not freed, of course. The plan was for a single, long-running set of defs to be available globally. My experience is that static objects don't initialize properly, so I use pointers to objects. I also tried making an extern version of these two objects, but this also caused memory problems.

hkaiser: Static objects initialize just fine, starting with C++11 their initialization is thread-safe even (which your solution is not, btw).

stevenrbrandt:
I've refactored this to have a compiler_state object that can be created in Python and passed around. It works, except the following problem occurs at shutdown:
```
terminate called after throwing an instance of 'std::invalid_argument'
  what():  hpx::resource::get_partitioner() can be called only after the resource partitioner has been allowed to parse the command line options.
Aborted (core dumped)
```

hkaiser:
@stevenrbrandt how do I reproduce this error during shutdown?

hkaiser:
@stevenrbrandt this happens during destruction of the (global) `compiler_state` object during module unload. This happens after HPX has been released already. I think we'll need to add a weak reference to the phylanx extension library for each `compiler_state` instance created. Let me see what I can do.

hkaiser:
@stevenrbrandt things should be ok now.

python/phylanx/ast/transformation.py
```diff
@@ -442,15 +446,14 @@ def __init__(self, f):
             assert len(tree.body) == 1
 
             if target == "PhySL":
-                self.__physl_src__ = '%s(%s)\n' % (
-                    full_node_name(tree.body[0], 'block'),
-                    transformation.recompile(tree))
+                self.__physl_src__ = transformation.recompile(tree)
+                et.eval(self.__physl_src__, self.cs)
```
hkaiser: Should things be evaluated whenever the decorators is created?

stevenrbrandt: @hkaiser It seems to work now. Yes, when the decorator is evaluated, the function is compiled. This is an improvement over what happened before, where the function was compiled on each evaluation. When the function is called, eval is called again, but instead of the function definition, only the function name is needed.

hkaiser: Neat trick!

hkaiser: OTOH, doesn't this also mean that the function is evaluated twice? Once without arguments (which could btw fail) and once with them? It might be better to explicitly separate compilation and evaluation in the Python interface.

hkaiser:
@stevenrbrandt could you address the failing builds for us to be able to go ahead with merging, please?

PR #310, opened 2018-03-16
fixed small compile issue with -lboost_regex

ct-clmsn:
compile error on the lra example. ld couldn't find -lboost_regex. this patch fixes the issue on ubuntu (or ubuntu-windows).

### DIFF
--- a/examples/algorithms/CMakeLists.txt
+++ b/examples/algorithms/CMakeLists.txt
@@ -11,7 +11,7 @@ set(example_programs
 
 foreach(example_program ${example_programs})
 
-  set(${example_program}_FLAGS DEPENDENCIES iostreams_component)
+  set(${example_program}_FLAGS DEPENDENCIES iostreams_component ${Boost_REGEX_LIBRARY})
 
   set(sources ${example_program}.cpp)
 


examples/algorithms/CMakeLists.txt
```diff
@@ -11,7 +11,7 @@ set(example_programs
 
 foreach(example_program ${example_programs})
 
-  set(${example_program}_FLAGS DEPENDENCIES iostreams_component)
```
ct-clmsn: on master this line looks like the following:

set(${example_program}_FLAGS DEPENDENCIES iostreams_component boost_regex)

hkaiser: @ct-clmsn I'm not sure I understand what you mean by that. Care to elaborate?

ct-clmsn: Had a couple compile errors this week after doing several synchs with master. Isolated the issue down to that line in CMakeLists.txt. Made the modification and found boost_regex isn't in the master CMakeLists.txt...after submitting the PR. I'll close this PR.

PR #311, opened 2018-03-16
Sum primitive

parsa:
This PR implements sum primitive, which is needed in the K-Means example. Adapted from:

* <https://docs.scipy.org/doc/numpy/reference/generated/numpy.sum.html>

### DIFF
--- a/src/execution_tree/primitives/sum_operation.cpp
+++ b/src/execution_tree/primitives/sum_operation.cpp
@@ -1,0 +1,289 @@
+// Copyright (c) 2018 Parsa Amini
+// Copyright (c) 2018 Hartmut Kaiser
+//
+// Distributed under the Boost Software License, Version 1.0. (See accompanying
+// file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+#include <phylanx/config.hpp>
+#include <phylanx/execution_tree/primitives/sum_operation.hpp>
+#include <phylanx/ir/node_data.hpp>
+#include <phylanx/util/matrix_iterators.hpp>
+
+#include <hpx/include/lcos.hpp>
+#include <hpx/include/naming.hpp>
+#include <hpx/include/util.hpp>
+#include <hpx/throw_exception.hpp>
+#include <hpx/util/iterator_facade.hpp>
+#include <hpx/util/optional.hpp>
+
+#include <algorithm>
+#include <cstddef>
+#include <cstdint>
+#include <memory>
+#include <string>
+#include <utility>
+#include <vector>
+
+#include <blaze/Math.h>
+
+///////////////////////////////////////////////////////////////////////////////
+namespace phylanx { namespace execution_tree { namespace primitives
+{
+    ///////////////////////////////////////////////////////////////////////////
+    primitive create_sum_operation(hpx::id_type const& locality,
+        std::vector<primitive_argument_type>&& operands,
+        std::string const& name, std::string const& codename)
+    {
+        static std::string type("sum_operation");
+        return create_primitive_component(
+            locality, type, std::move(operands), name, codename);
+    }
+
+    match_pattern_type const sum_operation::match_data =
+    {
+        hpx::util::make_tuple("sum_operation",
+        std::vector<std::string>{"sum(_1, _2)"},
+        &create_sum_operation, &create_primitive<sum_operation>)
+    };
+
+    ///////////////////////////////////////////////////////////////////////////
+    sum_operation::sum_operation(
+        std::vector<primitive_argument_type>&& operands,
+        std::string const& name, std::string const& codename)
+      : primitive_component_base(std::move(operands), name, codename)
+    {}
+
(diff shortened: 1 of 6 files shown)


src/execution_tree/primitives/sum_operation.cpp
```diff
@@ -0,0 +1,290 @@
...
+        std::string const& name, std::string const& codename)
+    {
+        static std::string type("sum_operation");
+        return create_primitive_component(
+            locality, type, std::move(operands), name, codename);
+    }
+
+    match_pattern_type const sum_operation::match_data =
+    {
+        hpx::util::make_tuple("sum_operation",
```
hkaiser: I'd suggest to let PhySL refer to the primitive as `sum` not `sum_operation`.

src/execution_tree/primitives/sum_operation.cpp
```diff
@@ -0,0 +1,290 @@
...
+        std::string const& name, std::string const& codename)
+      : primitive_component_base(std::move(operands), name, codename)
+    {}
+
+    primitive_argument_type sum_operation::sum0d(arg_type&& arg,
+        hpx::util::optional<std::int64_t> axis, bool keep_dims) const
+    {
+        if (axis && axis.value() != 0 && axis.value() != -1)
+        {
+            /*HPX_THROW_EXCEPTION(hpx::bad_parameter,
```
hkaiser: If this is not needed, please remove the code.

src/execution_tree/primitives/sum_operation.cpp
```diff
@@ -0,0 +1,290 @@
...
+                    {
+                        HPX_THROW_EXCEPTION(hpx::bad_parameter,
+                            "sum_operation::eval",
+                            execution_tree::generate_error_message(
+                                "invalid number of arguments specified", this_->name_,
+                                this_->codename_));
+                    }
+                    else if (args.size() > 1)
+                    {
+                        switch (args[1].index())
```
hkaiser: `args[1]` refers to a `primitive_argument_type`. That means that `index == 0, 1, 2` mean `nil`, `node_data<bool>`, and `std::int64_t`. Is that what you need?

src/execution_tree/primitives/sum_operation.cpp
```diff
@@ -0,0 +1,290 @@
...
+
+    primitive_argument_type sum_operation::sum2d_flat(
+        arg_type&& arg, bool keep_dims) const
+    {
+        auto m = arg.matrix();
+        double result = 0.;
+        for (std::size_t i = 0; i < m.columns(); ++i)
+        {
+            auto col = blaze::column(m, i);
+            result += std::accumulate(col.begin(), col.end(), 0.);
```
hkaiser: The initial value should be `result`.

src/execution_tree/primitives/sum_operation.cpp
parsa: Yes:
* Argument `axis` may or may not be present
    * case 0 if not present
    * case 2 if present).
* Argument `keep_dims` is also optional, but has to be the 3rd argument if `axis` is also present.
   * case 1
   * nested if inside case 2.

src/execution_tree/primitives/sum_operation.cpp
parsa: I could write `result = std::accumulate(begin, end, result)` or `result += std::accumulate(begin, end, 0.)` as I did. What is wrong with it?

src/execution_tree/primitives/sum_operation.cpp
parsa: Done.

src/execution_tree/primitives/sum_operation.cpp
parsa: Fixed.

src/execution_tree/primitives/sum_operation.cpp
hkaiser: I think you should utilize one of the existing extraction functions for this instead (for instance `phylanx::execution_tree::numeric_operand` or similar).

Generally I'd like to avoid having checks for the index() of any `primitive_argument_type` all over the code, Let's try to keep that localized.

parsa: Done.

src/execution_tree/primitives/sum_operation.cpp
```diff
@@ -0,0 +1,255 @@
...
+    {
+        static std::string type("sum_operation");
+        return create_primitive_component(
+            locality, type, std::move(operands), name, codename);
+    }
+
+    match_pattern_type const sum_operation::match_data =
+    {
+        hpx::util::make_tuple("sum_operation",
+        std::vector<std::string>{"sum(_1, _2)"},
```
hkaiser: This is not correct anymore, this vector should contain the three allowed schemes: 

    "sum(_1)", "sum(_1, _2)", "sum(_1, _2, _3)"

parsa: Fixed.

PR #321, opened 2018-03-20
Gradient primitive

justwagle:
This pull request adds the gradient (similar to numpy.gradient) primitive to Phylanx.


src/execution_tree/primitives/gradient_operation.cpp
```diff
@@ -0,0 +1,265 @@
...
+    {
+        static std::string type("gradient");
+        return create_primitive_component(
+            locality, type, std::move(operands), name, codename);
+    }
+
+    match_pattern_type const gradient_operation::match_data =
+    {
+        hpx::util::make_tuple("gradient",
+            std::vector<std::string>{"gradient(__1)"},
```
hkaiser: This will accept any number of arguments. It might be better to use `gradient(_1, _2)` if you need exactly two arguments.

justwagle: I wanted the gradient primitive to take either one or two arguments, the second one being the optional axis. In order to make sure that I get either one or two arguments, I throw an exception if the size of the operands is greater than 2 inside the eval() function.

hkaiser: In case you might want to write:
```
match_pattern_type const gradient_operation::match_data = 
{ 
    hpx::util::make_tuple("gradient", 
        std::vector<std::string>{"gradient(_1, _2)", "gradient(_1)"}, 
        &create_gradient_operation, 
        &create_primitive<gradient_operation>) 
}; 
```

PR #322, opened 2018-03-20
Range Infrastructure

parsa:
This PR adds support for ranges. Ranges can be an integer range or a range of arguments.

### DIFF
--- a/phylanx/ir/ranges.hpp
+++ b/phylanx/ir/ranges.hpp
@@ -1,0 +1,94 @@
+// Copyright (c) 2018 Parsa Amini
+// Copyright (c) 2018 Hartmut Kaiser
+//
+// Distributed under the Boost Software License, Version 1.0. (See accompanying
+// file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+#if !defined(PHYLANX_IR_RANGES)
+#define PHYLANX_IR_RANGES
+
+#include <phylanx/config.hpp>
+#include <phylanx/execution_tree/primitives/base_primitive.hpp>
+#include <phylanx/util/variant.hpp>
+
+#include <hpx/include/serialization.hpp>
+#include <hpx/include/util.hpp>
+#include <hpx/throw_exception.hpp>
+
+#include <array>
+#include <cstddef>
+#include <cstdint>
+#include <iosfwd>
+#include <type_traits>
+#include <utility>
+#include <vector>
+
+#include <blaze/Math.h>
+
+namespace phylanx { namespace ir
+{
+    //////////////////////////////////////////////////////////////////////////
+    template <typename T>
+    class range_iterator
+        : public hpx::util::iterator_facade<range_iterator<T>>,
+        T,
+        std::forward_iterator_tag,
+        T const&,
+        std::ptrdiff_t,
+        T const*>
+    {
+    public:
+        friend class hpx::util::iterator_core_access;
+
+        typename base_type::reference dereference() const
+        {
+            ;
+        }
+
+        bool equal(range_iterator const& other) const
+        {
+            ;
+        }
+
+        void advance(typename base_type::difference_type n)
+        {
+            ;
+        }
+
+        void increment()
+        {
+            ;
+        }
+
+        void decrement()
+        {
+            ;
+        }
+
+        typename base_type::difference_type distance_to(
+            range_iterator const& other) const
+        {
+            ;
+        }
+
+        std::size_t index_;
+    };
+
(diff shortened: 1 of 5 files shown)


src/ir/ranges.cpp
```diff
@@ -0,0 +1,121 @@
...
+            return *(util::get<1>(it_));
+        }
+
+        HPX_THROW_EXCEPTION(hpx::invalid_status,
+            "phylanx::ir::range_iterator::dereference",
+            "range_iterator object holds unsupported data type");
+    }
+
+    bool range_iterator::equal(range_iterator const& other) const
+    {
```
hkaiser: Please make sure both iterators hold the same type (index) before comparing the values themselves.

src/ir/ranges.cpp
```diff
@@ -0,0 +1,121 @@
...
+        case 0:    // int_range_type
+        {
+            int_range_type& p = util::get<0>(it_);
+            p.first += p.second;
+            break;
+        }
+        case 1:    // arg_range_type
+            ++util::get<1>(it_);
+            break;
+        default:
```
hkaiser: This misses the case for `index() == 2`.

phylanx/ir/ranges.hpp
```diff
@@ -0,0 +1,110 @@
...
+            execution_tree::primitive_argument_type>
+    {
+    private:
+        using int_range_type = std::pair<std::int64_t, std::int64_t>;
+        using arg_range_type =
+            std::vector<execution_tree::primitive_argument_type>::iterator;
+        using iterator_type = util::variant<int_range_type, arg_range_type>;
+
+    public:
+        range_iterator(int_range_type it)
```
hkaiser: This constructor is not used anywhere, do we need it?

tests/unit/ir/range_iterator.cpp
```diff
@@ -0,0 +1,114 @@
...
+    std::vector<arg_t> v{
+        arg_t{static_cast<std::int64_t>(6)},
+        arg_t{static_cast<std::int64_t>(9)},
+        arg_t{static_cast<std::int64_t>(42)}};
+
+    phylanx::ir::range_iterator it(v.begin());
+    HPX_TEST_EQ(*it++, 6);
+    HPX_TEST_EQ(*it++, 9);
+    HPX_TEST_EQ(*it++, 42);
+}
```
hkaiser: Please verify that `it == range_iterator{v.end()}`

src/ir/ranges.cpp
parsa: Done.

phylanx/ir/ranges.hpp
parsa: Removed.

tests/unit/ir/range_iterator.cpp
parsa: Done.

phylanx/ir/ranges.hpp
```diff
@@ -0,0 +1,155 @@
...
+
+        range_iterator begin();
+        range_iterator end();
+
+        reverse_range_iterator rbegin();
+        reverse_range_iterator rend();
+
+        explicit range(
+            std::vector<execution_tree::primitive_argument_type> data)
+          : data_(data)
```
hkaiser: This should be `std::move(data)`. Alternatively you might want to provide two constructors, one for `std::vector<execution_tree::primitive_argument_type> const& data` and one for `std::vector<execution_tree::primitive_argument_type> && data`.

src/ir/ranges.cpp
```diff
@@ -0,0 +1,256 @@
...
+            std::int64_t start = hpx::util::get<0>(int_range);
+            std::int64_t step = hpx::util::get<2>(int_range);
+            return range_iterator{start, step};
+        }
+        case 1:    // args_type
+            return util::get<1>(data_).begin();
+        case 2:    // arg_pair_type
+            return util::get<2>(data_).first;
+        case 3:
+            HPX_THROW_EXCEPTION(hpx::bad_parameter,
```
hkaiser: Hmmm. A range should not know whether it is a 'reverse' range or not. Also ranges are always 'forward' ranges. This makes it always be possible to derive a reverse iterator from a forward range.

phylanx/ir/ranges.hpp
parsa: Done.

src/ir/ranges.cpp
parsa: I think I addressed it. We can't do `range x(v.rbegin(), v.rend()` anymore, though.

Issue #320, opened 2018-03-19
Store Primitive Error 


aserio:
The following code produces throws an exception: "store function should only be called for the store_primitive"

```scheme
block(
    // Newton's Method
    define(xn, 3.0),
    define(x, 0.0),
    define(error, 1.0),
    define(max_iter, 10.0),

    define(func, a, 4.0 - (a * a)),
    define(d_func, a, (2.0 * a)),
    define(newton, xn, error, max_iter, block(
        define(iter, 0.0),
        while(iter < max_iter, block(
            store(iter, iter + 1.0),
            define(fx, func(xn)),
            define(dfx, d_func(xn)),
            define(x_guess, xn + (fx/dfx)),
            store(error, xn - x_guess),
            store(xn, x_guess),
            xn
        ))
    )),
    newton(xn, error, max_iter)
)
```
Where should I look to fix this problem? Is there a phrasing that would make the error message more informative?

hkaiser:
As we found out today, this is caused by trying to `store()` a value in one of the function arguments (`error` and `xn`). This is not supported yet. As a workaround I'd suggest to create an additional local variable, initialized from the argument, and to use this variable inside the function instead. 

We will have to check whether storing a new value in one of the function arguments is safe to implement or whether it might cause other issues.

hkaiser:
...and yes, the error message is as cryptic as it can get ;-) This needs fixing.

hkaiser:
This should have been fixed by merging #642. So I'm going to close this, please re-open, if necessary.

PR #323, opened 2018-03-20
Addressing #320: improved generated error messages

hkaiser:
- flyby: add is_<foo>_operand helper functions to introspect primitive_argument_type instances


hkaiser:
@justwagle this PR has the query functions you requested.

PR #328, opened 2018-03-22
Add Range support to primitive_argument_type

parsa:
This PR is the second phase of implementing range support in Phylanx. From now on `primitive_argument_type`s cannot contain `vector<primitive_argument_type>`s but `range`s instead.


parsa:
Superseded by #355

PR #330, opened 2018-03-23
Adding more test to make sure operations do not change the original value of the operands- issue 245

scheherzade:


src/execution_tree/primitives/and_operation.cpp
```diff
@@ -120,6 +120,30 @@ namespace phylanx { namespace execution_tree { namespace primitives
                     and_.name_, and_.codename_));
         }
 
+        primitive_argument_type operator()(ir::node_data<double>&& lhs,
+            std::int64_t&& rhs) const
+        {
+            if (lhs.num_dimensions() != 0)
+            {
+                return and_.and_all(ir::node_data<bool>{std::move(lhs)},
```
hkaiser: Doesn't this create a copy of the original matrix?

src/execution_tree/primitives/div_operation.cpp
```diff
@@ -447,8 +447,16 @@ namespace phylanx { namespace execution_tree { namespace primitives
 
         if (ops.size() == 2)
         {
-            lhs.matrix() =
-                blaze::map(lhs.matrix(), rhs.matrix(), divndnd_simd());
+            if (lhs.is_ref())
```
hkaiser: Excellent catch!

scheherzade: Yes it does. What about the new commit?

PR #332, opened 2018-03-24
Removing direct eval actions from primitives

hkaiser:
- This is adding runtime adaptive switching between direct or
  asynchronous execution of `primitive::eval`.
- Removing the performance counters that were reporting data from
  direct evaluation
- Adding performance counter `/phylanx/primitives/<name>/eval_direct`

To be functional, this PR also requires for
https://github.com/STEllAR-GROUP/hpx/pull/3254 to be merged first.

### DIFF
--- a/phylanx/execution_tree/primitives/primitive_component.hpp
+++ b/phylanx/execution_tree/primitives/primitive_component.hpp
@@ -13,6 +13,7 @@
 #include <hpx/include/actions.hpp>
 #include <hpx/include/components.hpp>
 #include <hpx/include/util.hpp>
+#include <hpx/runtime/launch_policy.hpp>
 
 #include <cstdint>
 #include <memory>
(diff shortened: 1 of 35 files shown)


phylanx/execution_tree/primitives/primitive_component.hpp
```diff
@@ -13,6 +13,8 @@
 #include <hpx/include/actions.hpp>
 #include <hpx/include/components.hpp>
 #include <hpx/include/util.hpp>
+#include <hpx/runtime/launch_policy.hpp>
+#include <hpx/runtime/launch_policy.hpp>
```
justwagle: This header is included twice.

hkaiser: Good catch, will fix. Thanks!

Issue #250, opened 2018-02-25
Phylanx Jupyter integration


hkaiser:
The following things are missing from the current integration of Phylanx with Jupyter:

- [ ] `cout()` primitive (the python `print()` function) prints to the Jupyter console
- [ ] integrate performance visualization script with Jupyter

hkaiser:
@stevenrbrandt, @rtohid please add more, if needed

stevenrbrandt:
@rtohid I'm pretty sure the way to solve this is to use Pybind11's py::print() function instead of hpx::cout. However, I'm unable to figure out how to get the build system to accept this seemingly easy change.

hkaiser:
@stevenrbrandt using `pybind11::print` should be doable if we created a new Phylanx primitive. That may require some acrobatics to avoid circular link dependencies between the phylanx_component and the phylanx_py module, though. 

The simpler alternative would be to intercept `std::cout` (which is used by the `cout()` primitive):
```
execution_tree.def("eval", [](std::string const& expr, pybind11::dict dict) {
    pybind11::scoped_ostream_redirect stream(
        std::cout,                                     // std::ostream&
        pybind11::module::import("sys").attr("stdout") // Python output
    );
    expresson_compiler(expr, dict);
});
```

stevenrbrandt:
The above kind of works... sometimes we get console output, sometimes segfaults.

PR #339, opened 2018-04-05
Redirecting all std::cout output to Python's stdout

hkaiser:
This addresses part of #250

### DIFF
--- a/python/src/bindings/ast.cpp
+++ b/python/src/bindings/ast.cpp
@@ -255,7 +255,9 @@ void phylanx::bindings::bind_ast(pybind11::module m)
     ///////////////////////////////////////////////////////////////////////////
     // phylanx::ast::generate_ast()
     ast.def("generate_ast",
-        [](std::string const& code) {
+        [](std::string const& code)
+        {
+            pybind11::gil_scoped_release release;       // release GIL
             return phylanx::ast::generate_ast(code);
         },
         "generate an AST from the given expression string");
--- a/python/src/bindings/binding_helpers.hpp
+++ b/python/src/bindings/binding_helpers.hpp
@@ -39,12 +39,17 @@ namespace phylanx { namespace bindings
         phylanx::execution_tree::compiler::environment eval_env;
         phylanx::execution_tree::compiler::function_list eval_snippets;
 
-        compiler_state()
+        static pybind11::object import_phylanx()
+        {
 #if defined(_DEBUG)
-          : m(pybind11::module::import("_phylanxd"))
+            return pybind11::module::import("_phylanxd");
 #else
-          : m(pybind11::module::import("_phylanx"))
+            return pybind11::module::import("_phylanx");
 #endif
+        }
+
+        compiler_state()
+          : m(import_phylanx())
           , eval_env(phylanx::execution_tree::compiler::default_environment())
           , eval_snippets()
         {
@@ -141,6 +146,8 @@ namespace phylanx { namespace bindings
     ///////////////////////////////////////////////////////////////////////////
     inline void expression_compiler(std::string xexpr_str, compiler_state& c)
     {
+        pybind11::gil_scoped_release release;       // release GIL
+
         namespace et = phylanx::execution_tree;
         return hpx::threads::run_as_hpx_thread(
             [&]() -> void
@@ -167,10 +174,13 @@ namespace phylanx { namespace bindings
     expression_evaluator(
         std::string xexpr_str, compiler_state& c, pybind11::args args)
     {
+        pybind11::gil_scoped_release release;       // release GIL
+
         namespace et = phylanx::execution_tree;
         return hpx::threads::run_as_hpx_thread(
             [&]() -> et::primitive_argument_type
             {
+                pybind11::gil_scoped_acquire acquire;
                 try
                 {
                     auto xexpr = phylanx::ast::generate_ast(xexpr_str);
--- a/python/src/bindings/binding_helpers.hpp
+++ b/python/src/bindings/binding_helpers.hpp
@@ -63,6 +63,8 @@ namespace phylanx { namespace bindings
         template <typename Ast>
         bool on_enter(Ast const& ast) const
         {
+            pybind11::gil_scoped_acquire acquire;       // acquire GIL
+
             auto d =
                 func_.attr("__class__").attr("__dict__").cast<pybind11::dict>();
             if (d.contains("on_enter"))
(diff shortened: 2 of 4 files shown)


hkaiser:
@stevenrbrandt this should fix the Jupyter printing issues. Could you please verify?

hkaiser:
@stevenrbrandt for the first issue I'm seeing this exception (in Jupyter) and something similar standalone:
```
---------------------------------------------------------------------------
Exception                                 Traceback (most recent call last)
<ipython-input-1-a9db418a1ecd> in <module>()
      2 from phylanx.ast import *
      3 
----> 4 @Phylanx()
      5 def hello():
      6     print("Hello")

D:\Devel\phylanx\build\All.vcpkg\python\build\lib.win-amd64-3.6\phylanx\ast\transformation.py in __init__(self, f)
    445             ast.increment_lineno(tree, actual_lineno)
    446             assert len(tree.body) == 1
--> 447             self.transformation = self.targets[target](tree, kwargs)
    448             self.__src__ = self.transformation.__src__
    449 

D:\Devel\phylanx\build\All.vcpkg\python\build\lib.win-amd64-3.6\phylanx\ast\transformation.py in __init__(self, tree, kwargs)
     52         self.priority = 0
     53         self.groupAggressively = True
---> 54         self.__src__ = self.recompile(tree)
     55 
     56     def _Num(self, a):

D:\Devel\phylanx\build\All.vcpkg\python\build\lib.win-amd64-3.6\phylanx\ast\transformation.py in recompile(self, a, allowreturn)
    386                 return self.nodes[nm](self, a) #, allowreturn)
    387             else:
--> 388                 return self.nodes[nm](self, a)
    389         except NotImplementedError:
    390             raise Exception('unsupported AST node type: %s' % nm)

D:\Devel\phylanx\build\All.vcpkg\python\build\lib.win-amd64-3.6\phylanx\ast\transformation.py in _Module(self, a)
    207     def _Module(self, a):
    208         args = [arg for arg in ast.iter_child_nodes(a)]
--> 209         return self.recompile(args[0])
    210 
    211     def _Return(self, a, allowreturn=False):

D:\Devel\phylanx\build\All.vcpkg\python\build\lib.win-amd64-3.6\phylanx\ast\transformation.py in recompile(self, a, allowreturn)
    386                 return self.nodes[nm](self, a) #, allowreturn)
    387             else:
--> 388                 return self.nodes[nm](self, a)
    389         except NotImplementedError:
    390             raise Exception('unsupported AST node type: %s' % nm)

D:\Devel\phylanx\build\All.vcpkg\python\build\lib.win-amd64-3.6\phylanx\ast\transformation.py in _FunctionDef(self, a)
    151                 aa = sargs[i]
    152                 if i + 1 == len(sargs):
--> 153                     s += self.recompile(aa, True)
    154                 else:
    155                     s += self.recompile(aa, False)

D:\Devel\phylanx\build\All.vcpkg\python\build\lib.win-amd64-3.6\phylanx\ast\transformation.py in recompile(self, a, allowreturn)
    384         try:
    385             if allowreturn:
--> 386                 return self.nodes[nm](self, a) #, allowreturn)
    387             else:
    388                 return self.nodes[nm](self, a)

D:\Devel\phylanx\build\All.vcpkg\python\build\lib.win-amd64-3.6\phylanx\ast\transformation.py in _Return(self, a, allowreturn)
    212         if not allowreturn:
    213             raise Exception(
--> 214                 "Return only allowed at end of function: line=%d\n" % a.lineno)
    215         args = [arg for arg in ast.iter_child_nodes(a)]
    216         return " " + self.recompile(args[0])

Exception: Return only allowed at end of function: line=7
```

For the second issue: yes, this is known, but unrelated, please see #340 for the cause.

stevenrbrandt:
@hkaiser you are reporting a different, unrelated issue to the two I am reporting. At present, you need to have a return in a @Phylanx() function. I don't understand what #340 is about. I couldn't make sense of the description and there's no example.

Anyway, I suppose, I made another discovery that clarifies my initial report. If I run this code in the notebook, it print's "Hello," even though the hello() function is never invoked.
```
import phylanx
from phylanx.ast import *

@Phylanx()
def hello():
    print("Hello")
    return 1
```
The interesting thing is, when print() has a constant argument, it is called at compile time rather than run time. Alternatively, if I simply pass a variable in along with the constant, then print() is called at runtime like it should be.
```
@Phylanx()
def hello3(n):
    print("Hello",n)
    return 1
```
The other observation is, calling print() in the notebook works now (yay!), even if it is sometimes called during compile time. The bad news is, whenever you call print() from the command line (i.e. not in the notebook), the code hangs.

hkaiser:
> Anyway, I suppose, I made another discovery that clarifies my initial report. If I run this code in the notebook, it print's "Hello," even though the hello() function is never invoked.
> ...
> The interesting thing is, when print() has a constant argument, it is called at compile time rather than run time. Alternatively, if I simply pass a variable in along with the constant, then print() is called at runtime like it should be.

Yes, this is what #340 is about. I will add some comments there.

> The other observation is, calling print() in the notebook works now (yay!), even if it is sometimes called during compile time. The bad news is, whenever you call print() from the command line (i.e. not in the notebook), the code hangs.

This just works fine for me. But I guess we could change the code such that the redirection is set up only if it runs in ipython...

hkaiser:
> @hkaiser you are reporting a different, unrelated issue to the two I am reporting

I'm not able to run the example code you gave me as it triggers the exception I listed. Could you please fix this? Should I create a separate ticket?

stevenrbrandt:
@hkaiser the reason my sample code doesn't work for you is you left off the last line: "return 1".

hkaiser:
> @hkaiser the reason my sample code doesn't work for you is you left off the last line: "return 1".

I did not leave off the last line.

hkaiser:
@stevenrbrandt please try again, I have added acquiring the GIL in more places.

rtohid:
@hkaiser #342 should fix the issue with the exception you encountered.

rtohid:
@hkaiser The latest change fixed the issue for me- I was experiencing the same hanging issue as @stevenrbrandt .

hkaiser:
Fixing this requires more work integrating the threads created by HPX with the Python runtime, see https://docs.python.org/3/c-api/init.html for more details.

stevenrbrandt:
This branch seems to work now. We should merge it.

hkaiser:
@stevenrbrandt Thanks for checking! Let's wait for the testing infrastructure to normalize before merging, though...

PR #360, opened 2018-04-18
Always use make_list to create a list in PhySL.

rtohid:


hkaiser:
This is superseded by #362.

PR #361, opened 2018-04-18
 Add support for YCM. 

rtohid:
Add CMake option PHYLANX_WITH_VIM_YCM to export compile commands and
generate appropriate YCM config file.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -111,6 +111,11 @@ phylanx_option(
   "Enable or disable building HDF5 Support"
   OFF ADVANCED CATEGORY "Build")
 
+phylanx_option(
+  PHYLANX_WITH_VIM_YCM BOOL
+  "Enable or disable YouCompleteMe configuration support for VIM"
+  OFF ADVANCED CATEGORY "Build")
+
 if(MSVC)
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -275,6 +285,19 @@ install(
...
+  configure_file(${CMAKE_SOURCE_DIR}/tools/vim/.ycm_extra_conf.py
+               ${build_dir_file} @ONLY)
+  add_custom_target(
+    configure_ycm
+    COMMAND ${CMAKE_COMMAND} -E copy ${build_dir_file} ${source_dir_file}
+    COMMENT "Copying YCM config file to source directory"
+    VERBATIM
+  )
+  phylanx_info("VIM YouCompleteMe: run 'make configure_ycm' to copy config file to source directory and enable support in YCM. To enable automatic loading of configure file, add to your .vimrc option: \"let g:ycm_extra_conf_globlist = ['${CMAKE_SOURCE_DIR}/*']\"")
+endif()
```
hkaiser: May I suggest to move this as a macro into a separate file and include it here? That might help keeping the main build system file a bit cleaner.

PR #362, opened 2018-04-18
1) Fix indexing for slicing so that it uses make_list() and commas ar…

stevenrbrandt:
1) Fix indexing for slicing so that it uses make_list() and commas are correctly inserted
2) Allow lists to be generated from Python lists
3) Add debug capability to @Phylanx() allowing the source to be pretty-printed before it is compiled


hkaiser:
Is this related to #360?

stevenrbrandt:
Yes, the changes in #360 are a subset of these changes.

PR #366, opened 2018-04-24
Adding the option to generate an output of type node_data<double> to comparison primitives

scheherzade:


src/plugins/booleans/equal.cpp
```diff
@@ -27,7 +27,7 @@ namespace phylanx { namespace execution_tree { namespace primitives
     match_pattern_type const equal::match_data =
     {
         hpx::util::make_tuple("__eq",
-            std::vector<std::string>{"_1 == _2", "__eq(_1, _2)"},
+            std::vector<std::string>{"_1 == _2", "__eq(_1, _2, _3)"},
```
hkaiser: We now probably should have all three patterns defined:

    std::vector<std::string>{"_1 == _2", "__eq(_1, _2)", "__eq(_1, _2, _3)"},

PR #367, opened 2018-04-25
Add AST load/dump support to PhySL interpreter

parsa:
Providing a `dump-ast=<path.ast>` flag to the interpreter creates an AST dump file from PhySL code.

e.g. `physl code.physl 10 --dump-ast=code.ast`

An AST dump file can be used instead of PhySL code by providing a `load-ast=<path.ast>` flag.
    * e.g. `physl --load-ast=code.ast`

Extras:
* Added a templated overload for `unserialize()` for AST expressions.
* Allow PhySL interpreter users to provide no positional arguments
    * e.g. `physl code.physl`
    * e.g. `physl --load-ast=file.ast`

### DIFF
--- a/examples/interpreter/physl.cpp
+++ b/examples/interpreter/physl.cpp
@@ -1,8 +1,8 @@
-//  Copyright (c) 2018 Parsa Amini
-//  Copyright (c) 2018 Hartmut Kaiser
+// Copyright (c) 2018 Parsa Amini
+// Copyright (c) 2018 Hartmut Kaiser
 //
-//  Distributed under the Boost Software License, Version 1.0. (See accompanying
-//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+// Distributed under the Boost Software License, Version 1.0. (See accompanying
+// file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
 
 #include <phylanx/phylanx.hpp>
 
@@ -44,6 +44,23 @@ std::string read_user_code(std::string const& arg)
         std::istreambuf_iterator<char>());
 
     return code;
+}
+
+void dump_ast(std::vector<phylanx::ast::expression> ast, std::string path)
+{
+    std::ofstream ast_stream(path, std::ios::binary);
+    if (!ast_stream.good())
+    {
+        HPX_THROW_EXCEPTION(hpx::filesystem_error,
+            "dump_ast",
+            "Failed to open the specified file: " + path);
+    }
+
+    // Serialize the AST into a byte array
+    std::vector<char> bytes = phylanx::util::serialize(ast);
+
+    // char is 1 byte
+    ast_stream.write(bytes.data(), bytes.size());
 }
 
 std::vector<phylanx::execution_tree::primitive_argument_type>
@@ -117,6 +134,8 @@ int handle_command_line(int argc, char* argv[], po::variables_map& vm)
                 "tree and the corresponding performance counter results")
             ("transform,t", po::value<std::string>(),
                 "file to read transformation rules from")
+            ("dump-ast,d", po::value<std::string>(),
+                "file to dump AST to")
         ;
 
         po::positional_options_description pd;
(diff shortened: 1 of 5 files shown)


examples/interpreter/physl.cpp
```diff
@@ -21,36 +22,97 @@
...
     // Read the file
-    code.assign(std::istreambuf_iterator<char>(code_stream),
-        std::istreambuf_iterator<char>());
+    std::ostringstream str_stream;
+    if (!(str_stream << code_stream.rdbuf()))
+    {
+        HPX_THROW_EXCEPTION(hpx::filesystem_error,
+            "read_user_code",
+            "Failed to read code from the specified file: " + path);
+    }
```
hkaiser: What is the rationale of this change?

examples/interpreter/physl.cpp
```diff
@@ -21,36 +22,97 @@
...
+    std::ifstream str_stream(path, std::ios::binary);
+    auto const start_pos = str_stream.tellg();
+    if (std::streamsize(-1) == start_pos)
+    {
+        HPX_THROW_EXCEPTION(hpx::filesystem_error,
+            "load_ast",
+            "Failed to open the specified file: " + path);
+    }
+
+    if (!str_stream.ignore((std::numeric_limits<std::streamsize>::max)()))
```
hkaiser: This will stop loading if `-1` (`EOF`) is found. What is the rationale of using `.ignore()`?

examples/interpreter/physl.cpp
```diff
@@ -173,43 +234,78 @@ int main(int argc, char* argv[])
...
+    std::vector<phylanx::ast::expression> ast;
+
+    if (vm.count("load-ast") != 0)
     {
-        // Execute code as given directly on the command line
-        user_code = vm["code"].as<std::string>();
-        code_source_name = "<command_line>";
+        auto path = vm["load-ast"].as<std::string>();
+        ast = load_ast(path);
+        code_source_name = "<ast_dump>";
```
hkaiser: we have the file name where we've loaded the AST from. Should we use this here?

examples/interpreter/physl.cpp
```diff
@@ -173,43 +234,78 @@ int main(int argc, char* argv[])
...
+                    read_user_code(vm["transform"].as<std::string>());
+
+                ast = phylanx::ast::transform_ast(
+                    ast, phylanx::ast::generate_transform_rules(transform_rules));
+            }
+
+            // Dump the AST to a file, if requested
+            if (vm.count("dump-ast") != 0)
+            {
+                dump_ast(ast, vm["dump-ast"].as<std::string>());
```
hkaiser: If `--dump-ast` is used without any additional parameters (no file name) we could use the base file name of the input file, just replacing the extension to `.physl-ast` or just `.ast` (or similar).

examples/interpreter/physl.cpp
```diff
@@ -173,43 +219,112 @@ int main(int argc, char* argv[])
...
+                        std::cerr << "physl: command line handling: exception "
+                                     "caught: the required path argument for "
+                                     "option '--dump-ast' is missing\n";
+                        return -1;
+                    }
+                    dump_file = fs::path(positional_args[0])
+                                    .replace_extension("ast")
+                                    .string();
+                }
+                dump_ast(ast, dump_file);
```
hkaiser: You might want to move the ast (as the `dump_ast` function takes the vector by value).

hkaiser: Alternatively, change `dump_ast` to take the vector by `const&`.

examples/interpreter/physl.cpp
```diff
@@ -173,43 +219,112 @@ int main(int argc, char* argv[])
...
+            }
+        }
+        catch (std::exception const& e)
+        {
+            std::cout << "physl: exception caught:\n" << e.what() << "\n";
+            return -1;
+        }
```
hkaiser: Also, I think the whole code block responsible for creating the AST should go into a separate function.

hkaiser: The interpreter can apply an AST transformation. Would it be possible to move the writing of the generated AST further down to enable writing the transformed AST instead of the originally generated one?

parsa: Its value is needed. It can't be move. Made it `const&`

examples/interpreter/physl.cpp
parsa: Its value is needed. It can't be move. Made it `const&`

PR #379, opened 2018-04-28
Adding generic unary arithmetic function plugins

tianyi93:
Adding generic unary arithmetic function primitive
existing exponential function removed


tianyi93:
Dr.Kaiser, 
the problem this afternoon on Linux was the code issue and it has been fixed.

PR #381, opened 2018-04-29
__add() now knows how to concatenate lists

### LINKED ISSUE #377: Using the + operator on lists in Python must concatenate lists
This example shows that behavior of the `+` operator inside Python functions decorated with `@Phylanx` is different from normal Python code.

### Regular Python Code
```python
a = []
b = a
b += [42]
print(a, b)
```
Output:
```
[42, 42] [42, 42]
```

### PhySL code
```python
@Phylanx()
def fx1():
    a = [42]
    b = a
    b += a
    print(a, b)
fx1()
```
Output:
```
[42] [84]
<nil>
```

hkaiser:
This fixes #377


hkaiser:
> The operation is concatenation, not addition. The names may need to reflect this

You were asking for concatenation using `+`.

> I don't know if using the same syntax for appending an element to a list and concatenating two lists in PhySL easily translates to Python.

Fair enough, we can remove this.

parsa:
Should we move it to its own primitive? We'd need to have a way to append an element to a list.

PR #389, opened 2018-04-30
Transforming Python for loops to Phylanx map.

rtohid:
Python `for` loops are now transformed into Phylanx `map` using lambda functions and the range primitive. Please note that the support for xrange has been dropped to follow Python3 syntax.

### DIFF
--- a/tests/unit/python/execution_tree/for.py
+++ b/tests/unit/python/execution_tree/for.py
@@ -44,4 +44,3 @@ def sumn3():
 
 
 assert sumn3() == 9
-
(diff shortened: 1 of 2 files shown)


tests/unit/python/execution_tree/for.py
```diff
@@ -44,18 +44,3 @@ def sumn3():
...
 assert sumn3() == 9
-
-
-@Phylanx("PhySL")
-def sumn4():
-    n = 0
-    sum = 0
-    c = 0
-    ss = -1
-    for n in xrange(3, 0, ss):
```
hkaiser: Should we map `xrange` onto `range` as well?

PR #392, opened 2018-05-03
Fixing annotation of symbols for certain function invocations

hkaiser:


hkaiser:
@scheherzade please verify whether this resolves the problem with the instrumentation printout for ALS you reported.

scheherzade:
@hkaiser yes it resolves the problem, thank you.

PR #399, opened 2018-05-05
Use the base Docker image instead of building it

parsa:
Updates:

* Use `stellargroup/phylanx_base:prerequisistes` as base image instead of `stellargroup/hpx:dev`
    * `stellargroup/phylanx_base:prerequisistes` is rebuilt every time any tag in `stellargroup/phylanx_base` is updated
* Test summaries are added to the CircleCI panel.
* `flake8` or `inspect` failures won't stop the build.
* Post-installation testing added
    * `physl` is run after installation.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,61 +3,327 @@
 #
 # Distributed under the Boost Software License, Version 1.0. (See accompanying
 # file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+################################################################################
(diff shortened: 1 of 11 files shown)


.circleci/config.yml
```diff
@@ -1,63 +1,328 @@
-# Copyright (c) 2017 R. Tohid
 # Copyright (c) 2018 Parsa Amini
```
hkaiser: I think you should leave Rod's copyright in place here

PR #415, opened 2018-05-13
Broadcasting rules

### LINKED ISSUE #410: Different results from numpy when performing arithmetic operations between a vector and a matrix with only one column
The result of performing `-`, `+`, `/`, and `*` operations between a vector and a matrix with only one column in Phylanx is different from NumPy.  
Generally to be able to perform this kind of operations between a vector and a matrix in NumPy, size of the vector should be equal to number of columns of the matrix, which is the way it is implemented in Phylanx. But in NumPy for cases where the matrix has only one column, you could still perform the operations, which is missing from Phylanx. 
e.g.
```python
np.array([1, 2]) + np.array([[3],[4]])
```
returns
```
array([[4, 6],
       [6, 8]])
```

while in PhySL: 
```
[1, 2] + [[3], [4]]
```
returns:
```
__add$0/0$1$1:: vector size does not match number of matrix columns: HPX(bad_parameter)
```

parsa:
This PR:
* Implements broadcasting rules of NumPy arrays for addition, subtraction, multiplication, and division primitives.
* Fixes #410

### DIFF
--- a/src/plugins/arithmetics/add_operation.cpp
+++ b/src/plugins/arithmetics/add_operation.cpp
@@ -37,7 +37,7 @@ namespace phylanx { namespace execution_tree { namespace primitives
 
             template <typename T>
             BLAZE_ALWAYS_INLINE auto operator()(T const& a) const
-                ->  decltype(a + std::declval<double>())
+                -> decltype(a + std::declval<double>())
             {
                 return a + scalar_;
             }
@@ -144,7 +144,9 @@ namespace phylanx { namespace execution_tree { namespace primitives
         default:
             HPX_THROW_EXCEPTION(hpx::bad_parameter,
                 "add_operation::add0d",
-                "the operands have incompatible number of dimensions");
+                execution_tree::generate_error_message(
+                    "the operands have incompatible number of dimensions",
+                    name_, codename_));
         }
     }
 
@@ -187,46 +189,8 @@ namespace phylanx { namespace execution_tree { namespace primitives
         std::size_t lhs_size = lhs.dimension(0);
         std::size_t rhs_size = rhs.dimension(0);
 
-        if (lhs_size != rhs_size)
-        {
-            // Propagation rule 2
-            if (lhs_size == 1)
-            {
-                if (rhs.is_ref())
-                {
-                    rhs = blaze::map(
-                        rhs.vector(), detail::add_simd(lhs.vector()[0]));
-                }
-                else
-                {
-                    rhs.vector() += blaze::map(
-                        rhs.vector(), detail::add_simd(lhs.vector()[0]));
-                }
-            }
-            else if (rhs_size == 1)
-            {
-                if (lhs.is_ref())
-                {
-                    lhs = blaze::map(
-                        lhs.vector(), detail::add_simd(rhs.vector()[0]));
-                }
-                else
-                {
-                    lhs.vector() += blaze::map(
-                        lhs.vector(), detail::add_simd(rhs.vector()[0]));
-                }
-            }
-            else
-            {
-                HPX_THROW_EXCEPTION(hpx::bad_parameter,
-                    "add_operation::add1d1d",
-                    execution_tree::generate_error_message(
-                        "the dimensions of the operands do not match", name_,
-                        codename_));
-            }
-        }
-        // Dimensions are identical
-        else
+        // Broadcasting rule 1: Dimensions are identical
+        if (lhs_size == rhs_size)
         {
             // Avoid overwriting references, avoid memory reallocation when possible
             if (lhs.is_ref())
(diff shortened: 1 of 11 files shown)


.circleci/config.yml
```diff
@@ -98,6 +98,7 @@ jobs:
                 name: Run CMake
                 command: cmake -DPHYLANX_WITH_GIT_COMMIT=${CIRCLE_SHA1} -DPHYLANX_WITH_TOOLS=On -DHPX_DIR=/usr/local/lib/cmake/HPX -Dblaze_DIR=/blaze/share/blaze/cmake -DCMAKE_BUILD_TYPE=Debug -DCMAKE_CXX_COMPILER=clang++ -DCMAKE_C_COMPILER=clang -DPHYLANX_WITH_HIGHFIVE=On ..
                 working_directory: /phylanx/build
+                when: always
```
hkaiser: These changes are unrelated, could you separate them, please?

src/plugins/matrixops/sum_operation.cpp
```diff
@@ -32,7 +32,7 @@ namespace phylanx { namespace execution_tree { namespace primitives
     ///////////////////////////////////////////////////////////////////////////
     match_pattern_type const sum_operation::match_data =
     {
-        hpx::util::make_tuple("sum_operation",
+        hpx::util::make_tuple("sum",
```
hkaiser: This requires changing the name here as well: https://github.com/STEllAR-GROUP/phylanx/blob/master/phylanx/plugins/matrixops/sum_operation.hpp#L77

src/plugins/matrixops/shuffle_operation.cpp
```diff
@@ -68,8 +68,8 @@ namespace phylanx { namespace execution_tree { namespace primitives
     ///////////////////////////////////////////////////////////////////////////
     match_pattern_type const shuffle_operation::match_data =
     {
-        hpx::util::make_tuple("shuffle_operation",
-            std::vector<std::string>{"shuffle_operation(_1)"},
+        hpx::util::make_tuple("shuffle",
```
hkaiser: This requires changing the name here as well: https://github.com/STEllAR-GROUP/phylanx/blob/master/phylanx/plugins/matrixops/shuffle_operation.hpp#L69

tools/inspect/inspect_to_junit.py
```diff
@@ -30,14 +30,16 @@
 def parse_inspect8_log(fh):
     line_pattern = re.compile('(.+):\ (\*.+\*.+)')
     split_pattern = re.compile(',\ (?=\*)')
+    stripper_pattern = re.compile('<\/?\w+[^>]*>')
```
hkaiser: These changes are unrelated, could you please separate them?

src/plugins/matrixops/sum_operation.cpp
parsa: This comment doesn't make sense. I don't have the old name anywhere.

hkaiser: Not anymore, yes.

src/plugins/arithmetics/add_operation.cpp
```diff
@@ -190,40 +189,89 @@ namespace phylanx { namespace execution_tree { namespace primitives
...
         else
         {
-            lhs.vector() += rhs.vector();
+            // Avoid overwriting references, avoid memory reallocation when possible
+            if (lhs.is_ref())
+            {
+                // Cannot reuse the memory if an operand is a reference
+                if (rhs.is_ref())
+                {
+                    lhs = lhs.vector() + rhs.vector();
```
hkaiser: Should this be `rhs = ...`?

src/plugins/arithmetics/add_operation.cpp
```diff
@@ -342,32 +389,302 @@ namespace phylanx { namespace execution_tree { namespace primitives
...
+            // Cannot reuse the memory if an operand is a reference
+            if (rhs.is_ref())
             {
-                blaze::row(m, i) = blaze::row(cm, i) + blaze::trans(cv);
+                lhs = lhs.matrix() + rhs.matrix();
+            }
+            // Reuse the memory from rhs operand
+            else
+            {
+                rhs.matrix() = lhs.matrix() + rhs.matrix();
```
hkaiser: Assigning things to `rhs` will not be propagated to the result returned by this function.

src/plugins/arithmetics/add_operation.cpp
```diff
@@ -342,32 +389,302 @@ namespace phylanx { namespace execution_tree { namespace primitives
...
+            if (rhs.is_ref())
+            {
+                blaze::DynamicMatrix<double> result(
+                    rhs.dimension(0), rhs.dimension(1));
+
+                for (std::size_t i = 0; i != result.rows(); ++i)
+                {
+                    blaze::row(result, i) =
+                        blaze::row(lhs.matrix(), 0) +
+                        blaze::row(rhs.matrix(), i);
```
hkaiser: You might want to avoid calling `.matrix()` for each row/column. Same applies below.

src/plugins/arithmetics/add_operation.cpp
```diff
@@ -342,32 +389,302 @@ namespace phylanx { namespace execution_tree { namespace primitives
...
+                return 1;
+            }
+            // rhs x dimensions must be stretched
+            else if (rhs == 1)
+            {
+                return 2;
+            }
         }
+        return 0;
+    }
```
hkaiser: I'd suggest using an enumerator type instead of integer values for distinguishing the necessary operations.

src/plugins/arithmetics/add_operation.cpp
```diff
@@ -342,32 +389,302 @@ namespace phylanx { namespace execution_tree { namespace primitives
...
+                "the dimensions of the operands do not match", name_,
+                codename_));
+    }
+
+    primitive_argument_type add_operation::add2d2d_stretch_xy(
+        arg_type&& lhs, arg_type&& rhs, int stretch_x, int stretch_y) const
+    {
+        blaze::DynamicMatrix<double> result(
+            (std::max)({lhs.dimension(0), rhs.dimension(0)}),
+            (std::max)({lhs.dimension(1), rhs.dimension(1)}));
```
hkaiser: Why do you call the `std::max` algorithm (usable for arbitrary number of arguments) instead of the simple `std::max` function (for 2 arguments)?

src/plugins/arithmetics/add_operation.cpp
```diff
@@ -342,32 +389,302 @@ namespace phylanx { namespace execution_tree { namespace primitives
...
+        {
+            if (stretch_y == 1)
+            {
+                for (std::size_t i = 0; i < result.rows(); ++i)
+                {
+                    for (auto& j : blaze::row(result, i))
+                    {
+                        j = lhs.matrix()(0, 0);
+                    }
+                }
```
hkaiser: Couldn't this just be a `result = DynamicMatrix<> m(rows, columns, value)` instead of the explicit iteration?

src/plugins/arithmetics/add_operation.cpp
```diff
@@ -342,32 +389,302 @@ namespace phylanx { namespace execution_tree { namespace primitives
...
+            if (stretch_y == 2)
+            {
+                for (std::size_t i = 0; i < result.rows(); ++i)
+                {
+                    for (auto& j : blaze::row(result, i))
+                    {
+                        j += rhs.matrix()(0, 0);
+                    }
+                }
+            }
```
hkaiser: Isn't this just a add2d0d instead?

PR #418, opened 2018-05-14
Fixing how 2nd argument to shape() is being handled

### LINKED ISSUE #406: shape() accepts anything for its second argument
`shape([1], [1])` returns `1`. It should only:
* Only accept an integer for its second argument..
* Either use it or raise an error when the value is invalid.

hkaiser:
This fixes #406 

@parsa, please verify that this fixes the problem you reported.


parsa:
My PhySL code worked, but I didn't run the `extract_shape` test. It seems to be broken now.

PR #421, opened 2018-05-14
Adding 'true' and 'false' to the compiler's list of predefined names

hkaiser:


hkaiser:
@rtohid as promised... the compiler now understands 'true' and 'false' as Boolean literal constants.

PR #422, opened 2018-05-15
Fix 411

rtohid:
Add support for constant names, namely, True, False and None.

### DIFF
--- a/tests/regressions/python/name_constants_411.py
+++ b/tests/regressions/python/name_constants_411.py
@@ -1,3 +1,8 @@
+#  Copyright (c) 2018 R. Tohid
+#
+#  Distributed under the Boost Software License, Version 1.0. (See accompanying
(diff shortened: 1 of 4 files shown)


tests/regressions/python/name_constants_411.py
```diff
@@ -0,0 +1,33 @@
...
+# assert f.__src__ == \
+#     "define$13$0(f$13$0, block$13$0(define$14$4(a$14$4, true$14$8), a$15$11))"
+
+# @Phylanx(debug=True)
+# def f():
+#     a = False
+#     return a
+
+# assert f.__src__ == \
+#     "define$13$0(f$21$0, block$21$0(define$22$4(a$22$4, false$22$8), a$23$11))"
```
hkaiser: Should these tests be uncommented?

PR #424, opened 2018-05-16
Add `true` and `false` to regression test.

rtohid:


parsa:
@hkaiser @rtohid The regression tests for this branch are broken and it's merged into master. It should be reverted or fixed ASAP.

PR #427, opened 2018-05-16
Change argmin and argmax to return node_data

parsa:
This PR:
* Modifies: argmin and argmax primitives.
* Rationale: NumPy's argmin and argmax return a NumPy array when an axis is specified. Phylanx's argmin and argmax was returning a list instead.


src/plugins/matrixops/argmax.cpp
```diff
@@ -157,12 +157,12 @@ namespace phylanx { namespace execution_tree { namespace primitives
         const matrix_row_iterator<decltype(a)> a_begin(a);
         const matrix_row_iterator<decltype(a)> a_end(a, a.rows());
 
-        std::vector<primitive_argument_type> result;
-        for (auto it = a_begin; it != a_end; ++it)
+        blaze::DynamicVector<double> result(a.rows());
```
hkaiser: Shouldn't this be a `blaze::DynamicVector<std:int64_t>` after all? Looks like this is another reason to get `node_data<std::int64_t>` under control asap... @tianyi93 ?

PR #433, opened 2018-05-16
Allow returning nil from PhySL blocks

### LINKED ISSUE #432: Cannot return nil in PhySL
We should be able to return `nil` just like any other value in PhySL. However, it does not work.
e.g.
```scheme
define(f, block(
    nil
))
cout(f())
```
Output:
```
physl: exception caught:
abandoning not ready shared state: HPX(broken_promise)
```

parsa:
Fixes #432.

### DIFF
--- a/tests/regressions/execution_tree/allow_return_nil_432.cpp
+++ b/tests/regressions/execution_tree/allow_return_nil_432.cpp
@@ -15,17 +15,45 @@
 #include <string>
 #include <sstream>
 
-std::string const code = R"(
-    define(f, block(
-        nil
-    ))
-    cout(f())
-)";
-
-int hpx_main(int argc, char* argv[])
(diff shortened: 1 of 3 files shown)


tests/regressions/execution_tree/allow_return_nil_432.cpp
```diff
@@ -0,0 +1,41 @@
...
+#include <string>
+#include <sstream>
+
+std::string const code = R"(
+    define(f, block(
+        nil
+    ))
+    cout(f())
+)";
+
```
hkaiser: Could you please add another test where nil is not the only statement inside the block, please?

parsa: Done

PR #437, opened 2018-05-18
Make sure map() and parallel_map() accept only lists (ranges) as their arguments.

hkaiser:
- flyby: change `'()` to `make_list()` in tests

### DIFF
--- a/phylanx/execution_tree/primitives/base_primitive.hpp
+++ b/phylanx/execution_tree/primitives/base_primitive.hpp
@@ -890,6 +890,43 @@ namespace phylanx { namespace execution_tree
         std::vector<primitive_argument_type> const& args,
         std::string const& name = "",
         std::string const& codename = "<unknown>");
+
+    // Extract a list from a primitive_argument_type (that
+    // could be a primitive or a literal value).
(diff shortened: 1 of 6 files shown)


hkaiser:
@stevenrbrandt this should at least prevent map() (and parallel_map()) from accepting anything but lists. Please verify.

stevenrbrandt:
@hkaiser is that the right solution? Do we want the code below to not work when @Phylanx() is added?
```
import numpy as np
vec = np.array([1,2,3,4])

def iter_vec(v):
    for i in v:
        print(i)

iter_vec(vec)
```

hkaiser:
@stevenrbrandt I'm not sure whether iteration over numpy arrays should work, I'm open to discussion here. This PR solves the orthogonal issue that map (and parallel_map) were accepting any value as iterables.

stevenrbrandt:
@hkaiser my concern is whether we can write simple kernels that can be optimized by polyhedral compilers. It may be that we prefer loops of the form:
```
for i in range(1,len(x)-1):
   y[i] = x[i+1]+x[i-1]
```

Issue #440, opened 2018-05-18
python _Assign functionality for subscripted matrices


ct-clmsn:
Currently expressions such as:

```python
from phylanx.ast import Phylanx

@Phylanx
def update_values(mat, N):
    for i in range(N):
        for j in range(N):
            mat[i, j] = 0.0
```
break in the following file/line `phylanx/python/phylanx/ast/transformation.py:287 (the Python ast.Subscript type does not have an 'id' member variable)`

justwagle:
@ct-clmsn You should be able to use the set primitive. 
set (input, row_start, row_stop, row_steps  , col_start, col_stop, col_steps , value_to_set )
If you want to extract an element, slice should support that.

ct-clmsn:
@justwagle thanks! should i start off by doing a PhySL implementation before going straight to python?

justwagle:
@ct-clmsn I think python version would be better as @stevenrbrandt is already working on the ticket to add the missing functionality in python.

ct-clmsn:
@stevenrbrandt go for it! @justwagle will continue to work on the python version to identify use-cases that need additional support.

stevenrbrandt:
For reference, the sample code above should be rewritten like this:
```
from phylanx.ast import *
import numpy as np

N = 3
m = np.ones ((N,N))

@Phylanx("PhySL")
def update_values(mat, N):
   for i in range(N):
      for j in range(N):
        mat[i,j] = 0.0

m2 = update_values(m,N)
print(m2)
```

stevenrbrandt:
This is now fixed in the fixing_440 branch

stevenrbrandt:
This has been re-fixed in branch fixing_440

hkaiser:
I believe we can close this now. Please re-open, if necessary.

PR #444, opened 2018-05-21
Fixing 440

stevenrbrandt:
This fixes ticket #440 and allows the code below to run:
```
from phylanx.ast import *
import numpy as np

N = 3
m = np.ones ((N,N))

@Phylanx("PhySL")
def update_values(mat, N):
   for i in range(N):
      for j in range(N):
        mat[i,j] = 0.0
   return mat

m2 = update_values(m,N)
print(m2)
print(m2.shape)

```

### DIFF
--- a/python/phylanx/ast/transformation.py
+++ b/python/phylanx/ast/transformation.py
@@ -10,16 +10,17 @@ import re
 import phylanx
 import inspect
 from .oscop import OpenSCoP
-from .utils import full_name, full_node_name, physl_fmt, dump_info
+from .utils import full_name, full_node_name, physl_fmt
 
 et = phylanx.execution_tree
 
 
-def is_node(node,name):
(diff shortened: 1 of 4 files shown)


python/phylanx/ast/transformation.py
```diff
@@ -296,13 +303,95 @@ def _Assign(self, a, allowreturn=False):
...
+                s += ")"
+            else:
+                raise Exception()
+            return s
         else:
-            s += "define" + symbol_info
-            self.defs[args[0].id] = 1
-        a = '%s' % full_node_name(args[0], args[0].id)
-        s += "(" + a + ", " + self.recompile(args[1]) + ")"
+            raise Exception()
```
hkaiser: Could this Exception have some error text associated with it, please?

PR #456, opened 2018-05-23
Added assert primitive

parsa:
This PR adds primitive `assert` that can be used for testing in PhySL code.

### DIFF
--- a/src/execution_tree/primitives/assert_condition.cpp
+++ b/src/execution_tree/primitives/assert_condition.cpp
@@ -1,0 +1,91 @@
+// Copyright (c) 2018 Parsa Amini
+// Copyright (c) 2018 Hartmut Kaiser
+//
+// Distributed under the Boost Software License, Version 1.0. (See accompanying
+// file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+#include <phylanx/config.hpp>
+#include <phylanx/execution_tree/primitives/assert_condition.hpp>
+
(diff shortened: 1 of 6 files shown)


src/execution_tree/primitives/assert_condition.cpp
```diff
@@ -0,0 +1,91 @@
...
+    {
+        static std::string type("assert");
+        return create_primitive_component(
+            locality, type, std::move(operands), name, codename);
+    }
+
+    match_pattern_type const assert_condition::match_data =
+    {
+        hpx::util::make_tuple("assert",
+            std::vector<std::string>{"assert(__1)"},
```
hkaiser: Shouldn't this be `"assert(_1)"`? I wouldn't expect for assert to accept more than one argument.

tests/unit/execution_tree/primitives/assert_condition.cpp
```diff
@@ -0,0 +1,66 @@
...
+
+    hpx::future<phylanx::execution_tree::primitive_argument_type> f =
+        assert_cond.eval();
+
+    bool exception_thrown = false;
+    try
+    {
+        // Must throw an exception
+        f.get();
+    }
```
hkaiser: You might want to insert a `HPX_TEST(false);` after accessing the future.

hkaiser: You're right, in this case there wouldn't be any benefit.

PR #459, opened 2018-05-24
changing int64_t to node_data<int64_t>

tianyi93:
changing int64_t to node_data<int64_t>

### DIFF
--- a/src/execution_tree/primitives/base_primitive.cpp
+++ b/src/execution_tree/primitives/base_primitive.cpp
@@ -1,4 +1,5 @@
 // Copyright (c) 2017-2018 Hartmut Kaiser
+// Copyright (c) 2018 Tianyi Zhang
 //
 // Distributed under the Boost Software License, Version 1.0. (See accompanying
 // file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
@@ -40,7 +41,7 @@ namespace phylanx { namespace execution_tree
         char const* const primitive_argument_type_names[] = {
             "phylanx::ast::nil",
             "phylanx::ir::node_data<std::uint8_t>",
-            "std::int64_t",
+            "phylanx::ir::node_data<std::int64_t>",
             "std::string",
             "phylanx::ir::node_data<double>",
             "phylanx::execution_tree::primitive",
@@ -331,7 +332,7 @@ namespace phylanx { namespace execution_tree
         {
         case 0: HPX_FALLTHROUGH;    // nil
         case 1: HPX_FALLTHROUGH;    // phylanx::ir::node_data<std::uint8_t>
-        case 2: HPX_FALLTHROUGH;    // std::uint64_t
+        case 2: HPX_FALLTHROUGH;    // ir::node_data<std::int64_t>
         case 3: HPX_FALLTHROUGH;    // std::string
         case 4: HPX_FALLTHROUGH;    // phylanx::ir::node_data<double>
         case 5: HPX_FALLTHROUGH;    // primitive
@@ -358,7 +359,7 @@ namespace phylanx { namespace execution_tree
         switch (val.index())
         {
         case 0: HPX_FALLTHROUGH;    // nil
-        case 2: HPX_FALLTHROUGH;    // std::uint64_t
+        case 2: HPX_FALLTHROUGH;    // ir::node_data<std::int64_t>
         case 3: HPX_FALLTHROUGH;    // std::string
         case 5: HPX_FALLTHROUGH;    // primitive
         case 6: HPX_FALLTHROUGH;    // std::vector<ast::expression>
@@ -406,7 +407,6 @@ namespace phylanx { namespace execution_tree
         switch (val.index())
         {
         case 0: HPX_FALLTHROUGH;    // nil
-        case 2: HPX_FALLTHROUGH;    // std::uint64_t
         case 3: HPX_FALLTHROUGH;    // std::string
         case 5: HPX_FALLTHROUGH;    // primitive
         case 6: HPX_FALLTHROUGH;    // std::vector<ast::expression>
@@ -424,7 +424,18 @@ namespace phylanx { namespace execution_tree
             }
             break;
 
-        case 4:     // phylanx::ir::node_data<double>
+        case 2:    // phylanx::ir::node_data<std::int64_t>
+            {
+                auto const& v = util::get<2>(val);
+                if (v.is_ref())
+                {
+                    return primitive_argument_type{v};
+                }
+                return primitive_argument_type{v.ref()};
+            }
+            break;
+
+            case 4:    // phylanx::ir::node_data<double>
             {
                 auto const& v = util::get<4>(val);
                 if (v.is_ref())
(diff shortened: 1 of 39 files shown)


phylanx/plugins/matrixops/extract_shape.hpp
```diff
@@ -48,9 +48,12 @@ namespace phylanx { namespace execution_tree { namespace primitives
...
-        primitive_argument_type shape0d(arg_type&& arg, std::int64_t index) const;
-        primitive_argument_type shape1d(arg_type&& arg, std::int64_t index) const;
-        primitive_argument_type shape2d(arg_type&& arg, std::int64_t index) const;
+        primitive_argument_type shape0d(
+            arg_type&& arg, ir::node_data<std::int64_t> index) const;
+        primitive_argument_type shape1d(
+            arg_type&& arg, ir::node_data<std::int64_t> index) const;
+        primitive_argument_type shape2d(
+            arg_type&& arg, ir::node_data<std::int64_t> index) const;
     };
```
hkaiser: These functions are to be called with a single integer as their argument. There is no point for them to be called with a vector or matrix of integers...

src/execution_tree/primitives/base_primitive.cpp
```diff
@@ -656,7 +687,8 @@ namespace phylanx { namespace execution_tree
         {
         case 0: HPX_FALLTHROUGH;    // nil
         case 1: HPX_FALLTHROUGH;    // phylanx::ir::node_data<std::uint8_t>
-        case 2: HPX_FALLTHROUGH;    // std::uint64_t
+        case 2:                     // ir::node_data<std::int64_t>
+            return std::move(val);
```
hkaiser: This change would make `case 0` and `case 1` fall through to `case 2` which is not what we want.

src/execution_tree/primitives/base_primitive.cpp
```diff
@@ -1053,34 +1085,41 @@ namespace phylanx { namespace execution_tree
...
+            HPX_FALLTHROUGH;    // phylanx::ir::node_data<std::uint8_t>
+        case 3:
+            HPX_FALLTHROUGH;    // string
+        case 4:
+            HPX_FALLTHROUGH;    // phylanx::ir::node_data<double>
+        case 5:
+            HPX_FALLTHROUGH;    // primitive
+        case 7:
+            HPX_FALLTHROUGH;    // phylanx::ir::range
         default:
```
hkaiser: Please try to avoid reformatting code you have not touched.

src/execution_tree/primitives/base_primitive.cpp
```diff
@@ -2067,8 +2107,7 @@ namespace phylanx { namespace execution_tree
             }
 
             return f.then(hpx::launch::sync,
-                [&](hpx::future<primitive_argument_type> && f)
-                {
+                [&](hpx::future<primitive_argument_type>&& f) {
```
hkaiser: Please try to avoid reformatting code you have not edited.

src/execution_tree/primitives/base_primitive.cpp
```diff
@@ -2093,8 +2133,7 @@ namespace phylanx { namespace execution_tree
...
```
hkaiser: Same here and more places below: please don't reformat untouched code.

src/ir/node_data.cpp
```diff
@@ -1305,6 +1338,44 @@ namespace phylanx { namespace ir
...
+                    detail::print_array<std::int64_t>(
+                            out, blaze::row(data, row), data.columns());
+                }
+                out << "]";
+            }
+                break;
+
+            default:
+                HPX_THROW_EXCEPTION(hpx::invalid_status,
+                                    "node_data<double>::operator<<()",
```
hkaiser: Please fix the error message.

src/plugins/algorithms/lra.cpp
```diff
@@ -157,7 +157,7 @@ namespace phylanx { namespace execution_tree { namespace primitives
         bool enable_output = false;
         if (args.size() == 5 && valid(args[4]))
         {
-            enable_output = extract_integer_value(args[4], name_, codename_) != 0;
+            enable_output = extract_integer_value(args[4], name_, codename_)[0] != 0;
```
hkaiser: We should change this to call `extract_boolean_value` instead.

src/plugins/booleans/and_operation.cpp
```diff
@@ -117,27 +118,27 @@ namespace phylanx { namespace execution_tree { namespace primitives
...
 
         primitive_argument_type operator()(ir::node_data<double>&& lhs,
-            std::int64_t&& rhs) const
+            ir::node_data<std::int64_t>&& rhs) const
         {
             if (lhs.num_dimensions() != 0)
             {
                 return and_.and_all(
-                    std::move(lhs), ir::node_data<double>{rhs !=0 ? 1.0 : 0.0});
+                    std::move(lhs), ir::node_data<double>{rhs[0] !=0 ? 1.0 : 0.0});
```
hkaiser: This code needs to be changed now as the comparison sematics have changed. Before we compared a `node_data<double>` with a single value, now we compare against a `node_data<int>`. Possible scenarios are now `Matrix<double> && Matrix<int>` etc. The same is probably true for all boolean operations...

src/plugins/matrixops/constant.cpp
```diff
@@ -39,11 +39,11 @@ namespace phylanx { namespace execution_tree { namespace primitives
             ir::range const& shape)
         {
             std::array<std::size_t, 2> result = {0, 0};
-            result[0] = extract_integer_value(*shape.begin());
+            result[0] = extract_integer_value(*shape.begin())[0];
```
hkaiser: We should probably throw an exception if `extract_integer_value` returns a `node_data<int64_t>` that holds anything but a scalar value (matrices and vectors don't make any sense). This probably applies for all uses of this function where the result is 'converted' to a single value.

hkaiser: Better yet, we should introduce a new function `extract_scalar_integer_value` that does the proper handling and use that everywhere we currently use `[0]` on the result.

src/execution_tree/primitives/base_primitive.cpp
```diff
@@ -990,12 +1023,118 @@ namespace phylanx { namespace execution_tree
...
+        case 3:HPX_FALLTHROUGH;    // string
+        case 5:HPX_FALLTHROUGH;    // primitive
+        case 7:HPX_FALLTHROUGH;    // phylanx::ir::range
+        default:
+            break;
+        }
+
+        std::string type(detail::get_primitive_argument_type_name(val.index()));
+        HPX_THROW_EXCEPTION(hpx::bad_parameter,
+            "phylanx::execution_tree::extract_integer_value",
```
hkaiser: Please fix the function name in this error message

src/execution_tree/primitives/base_primitive.cpp
```diff
@@ -990,12 +1023,118 @@ namespace phylanx { namespace execution_tree
...
```
hkaiser: Same here, please fix the function name.

src/plugins/algorithms/als.cpp
```diff
@@ -63,9 +63,9 @@ namespace phylanx { namespace execution_tree { namespace primitives
         }
         auto regularization = arg2.scalar();
 
-        auto num_factors = extract_integer_value(args[2], name_, codename_);
+        auto num_factors = extract_integer_value(args[2], name_, codename_)[0];
```
hkaiser: This would eb acandidate for the new function `extract_integer_scalar_value`, I think (same is true at several spots below).

tianyi93:
@hkaiser I have modified the codes accordingly. Thanks!

PR #460, opened 2018-05-24
Revert "Support Python indexing for assignment to matrices"

parsa:
Reverts #444 since it breaks master


hkaiser:
@stevenrbrandt please have another look at your original patchset. It broke things, thus it was reverted.

PR #461, opened 2018-05-24
parallel range, or 'prange' support

ct-clmsn:
adding 'prange' functionality to phylanx. prange allows users to explicitly define a 'parallel loop'. prange should map into PhySL's 'parallel_map' primitive. this PR is associated with #437 #2

[numba](http://numba.pydata.org/numba-doc/0.11/prange.html) provides prange support.  [hpat](https://intellabs.github.io/hpat/_build/html/source/supported.html#explicit-parallel-loops) provides a similar prange feature.


python/phylanx/ast/transformation.py
```diff
@@ -279,6 +289,11 @@ def _Call(self, a, allowreturn=False):
             args = [arg for arg in ast.iter_child_nodes(a)]
             if args[0].id == "print":
                 args[0].id = "cout"
+            if args[0].id == "xrange":
+                args[0].id = "range"
```
hkaiser: We have removed support for `xrange` as it's not a thing in Python V3 any more.

python/phylanx/util/prange.py
```diff
@@ -0,0 +1,39 @@
...
+# unit test
+#
+if __name__ == "__main__":
+    x = prange(0, 10)
+    for i in x:
+        print(i)
+
+    for j in prange(10,20):
+        print(j)
+'''
```
hkaiser: What's the purpose of this test? Doesn't this PR add a `prange` implementation on the PhySL side of things?

python/phylanx/ast/transformation.py
ct-clmsn: Will remove from commit. Thanks!

python/phylanx/util/prange.py
ct-clmsn: Will remove that too. Shouldn't have been part of the commit.

rtohid:
@ct-clmsn I doubt it is a `prange` problem, it most probably is a bug in `print`. I tested your example with both `range` and `prange` and both caused the same error on my machine:

```
0
0
1
Stack overflow in coroutine at address Segmentation fault
``` 

Meanwhile, the following finishes with no errors:
```py
@Phylanx                                                                                  
def test_prange():                                                                        
    a = 0                                                                                 
    for x in prange(0,10):                                                                 
        a = x                                                                             
    return a
```

However, it always return 9, so I am not sure if it is actually run in parallel or not.

python/phylanx/ast/transformation.py
```diff
@@ -527,8 +555,11 @@ def recompile(self, a, allowreturn=False):
         "shape": "shape",
         "sqrt": "square_root",
         "transpose": "transpose",
-        "vstack": "vstack"
+        "vstack": "vstack",
         # slicing operations
+        # ranges
+        "range" : "range",
```
rtohid: This is not necessary as `range` and `xrange` are not NumPy methods.

hkaiser:
> However, it always return 9, so I am not sure if it is actually run in parallel or not.

@rtohid our python frontend initializes HPX to run on one core only, currently. This needs to be made configurable somehow...

python/phylanx/ast/transformation.py
ct-clmsn: thanks - will remove.

ct-clmsn:
@hkaiser @rtohid @stevenrbrandt could a variable be added to the Phylanx annotation that controls the HPX runtime configuration options? Something maybe named hpx_config (a python dictionary)?

hkaiser:
@ct-clmsn yes, something should be done about this. Currently HPX is implicitly initialized whenever `import phylanx` is executed, and it is initialized with fixed parameters (`hpx.os_threads=1`).

stevenrbrandt:
Note that my ploop() code test above does work when I take the generated PhySL code and run it through the interpreter. The problem looks like a Python cleanup issue, something revealed by this pull request rather than something caused by it.

stevenrbrandt:
I attempted another test, one that tried to avoid prints within the loop (which I guessed was the problem above).
``from phylanx.ast import Phylanx

@Phylanx(debug=True)
def ploop():
    a = 0
    for i in prange(10):
        a += 1
    print(a)

ploop()
```
This test prints "1" every time it is run for me, not 10. I'm not sure whether we expect that behavior or not. Maybe reductions should be carried out using a parallel algorithm of some sort?

**Comment by ct-clmsn** [comment]
@stevenrbrandt - the python built-in function [reduce](https://docs.python.org/2/library/functions.html#reduce) could be overloaded.

**Comment by ct-clmsn** [comment]
@rtohid @stevenrbrandt @hkaiser do ya'll have advice on the the proper PhySL for doing a store operation into a slice?

**Comment by rtohid** [comment]
>could a variable be added to the Phylanx annotation that controls the HPX runtime configuration options? Maybe a python dictionary named hpx_config?

@ct-clmsn I created a ticket for this #469, we have a couple of ideas on how to make this possible in the python script.

**Comment by rtohid** [comment]
> do ya'll have advice on the proper PhySL for doing a store operation into a slice? i'm not 100% sure the following is correct.

@ct-clmsn I think you can do this through `set_operation` (PR #467). @stevenrbrandt could you please confirm?

**Comment by rtohid** [comment]
@stevenrbrandt You'd get different results if you run the example on a larger range, so I don't think there are any issues with `parallel_map`- I tried the following:

```
define(                                                                                   
    ploop,                                                                                
    block(                                                                                
        define(a, 0),                                                                     
        parallel_map(                                                                     
            lambda(                                                                       
                i,                                                                        
                block(                                                                    
                    store(a, a+1)                                                         
                )                                                                         
            ),                                                                            
            range(1000)                                                                                                                                                                       
        ),                                                                                
        (                                                                                 
            cout(a)                                                                       
        )                                                                                 
    )                                                                                     
) 
```

And you are right, we definitely need a reduction primitive.

**Comment by ct-clmsn** [comment]
@rtohid cool thanks for tracking the config ticket w/ #469 ! i'm working on testing out the set_operation primitive.

**Comment by stevenrbrandt** [comment]
My pull request (fixing_440) would make your test file work. Failing that, Rod had an idea:
```
from phylanx.ast import Phylanx

@Phylanx()
def ptest():
  v1 = 0
  v2 = 0
  v3 = 0
  for i in prange(3):
    if i==0:
      v1 = 1
    if i==1:
      v2 = 1
    if i==2:
      v3 = 1
  return v1+v2+v3

assert ptest() == 3
```

**Comment by stevenrbrandt** [comment]
@ct-clmsn It looks like you committed a few debug print statements.

**Comment by ct-clmsn** [comment]
@stevenrbrandt my bad; will clean up the mess and push updates.

File: python/phylanx/ast/transformation.py [hunk 4]
```diff
@@ -241,13 +254,20 @@ def _Assign(self, a, allowreturn=False):
         symbol_info = full_node_name(a)
         args = [arg for arg in ast.iter_child_nodes(a)]
         s = ""
-        if args[0].id in self.defs:
-            s += "store" + symbol_info
+
+        if is_node(args[0], 'Subscript'):
+            #print('Assign1', args[0], args[1])
+            s += "set" + symbol_info + "("
+            s += self.recompile(args[0]) + ", 0, 0, 0, 0, 1, 1, " + self.recompile(args[1]) + ")"
```
hkaiser: What is the rationale to switch from using store() to using set()?

python/phylanx/ast/transformation.py
```diff
@@ -392,10 +412,23 @@ def _UnaryOp(self, a, allowreturn=False):
...
+        ret = ""
+
+        if is_node(a.iter, 'Call'):
+            func_nom = get_call_func_name(a.iter)
+            if func_nom == 'prange':
+                ret = "parallel_map%s(lambda(" % symbol_info
+            else:
+                ret = "map%s(lambda(" % symbol_info
+        else:          
+            ret = "map%s(lambda(" % symbol_info
```
hkaiser: I know this was missing before, but should we annotate the lambda() primitive with symbol information as well?

python/phylanx/ast/transformation.py
ct-clmsn: @rtohid suggested testing it out. Is there a preference? I might try to start putting together some small documentation about PhySL primitives.

python/phylanx/ast/transformation.py
ct-clmsn: Shouldn't be too hard to add. Will get that added in today.

python/phylanx/ast/transformation.py
stevenrbrandt: I think this overlaps with what I did in fixing_440.

ct-clmsn: @hkaiser store fails on execution when a slice is the target of the storage.

stevenrbrandt: @ct-clmsn If you wait for fixing_440 to be merged to master, then merge master with this branch, you can set array elements the way you want.

ct-clmsn: @stevenrbrandt will hold out for fixing_440. thanks!

PR #467, opened 2018-05-25
Rewrite set_operation.py

stevenrbrandt:
This is my second attempt to get the set_operation to work. I rewrote the stencil example so that it does not trip the same bugs as the previous version.

### DIFF
--- a/python/phylanx/ast/transformation.py
+++ b/python/phylanx/ast/transformation.py
@@ -16,6 +16,13 @@ from .utils import full_name, full_node_name, physl_fmt
 et = phylanx.execution_tree
 
 
+def is_node(node, name):
+    """Return the node name"""
+    if node is None:
+        return False
+    return node.__class__.__name__ == name
+
+
 def get_node(node, **kwargs):
     if node is None:
         return None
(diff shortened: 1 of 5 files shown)


python/phylanx/ast/transformation.py
```diff
@@ -394,15 +483,14 @@ def _If(self, a, allowreturn=False):
         return s
 
     def _List(self, a, allowreturn=False):
-        symbol_info = full_node_name(a)
-        ret = "make_list%s(" % symbol_info
+        ret = "["
```
hkaiser: Please use `make_list()` in the generated PhySL code. `[...]` will be interpreted as a (numeric) array (see: https://github.com/STEllAR-GROUP/phylanx/blob/master/phylanx/ast/parser/expression_def.hpp#L129-L131).

tests/unit/python/execution_tree/set_operation.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+# This function is never called, it just makes Flake 8 quiet
+def shape(a, b):
+    pass
+
+
+@Phylanx()
+def kernel(a, b):
+    for i in range(1, shape(a, 0) - 1):
```
hkaiser: I think you should be able to use `np.shape(a)[0]` here. This might be closer to what we want to achieve.

PR #471, opened 2018-05-28
Fix assert primitive test

parsa:
Assert primitive test are failing. This PR resolves the issue.


tests/unit/python/execution_tree/eval.py
```diff
@@ -194,8 +194,8 @@ def f7(b):
 assert np.all(f7(np_b) == np.array(b[0, 0:2]))
 
 
-@Phylanx(debug=True)
-def f8():
-    a = [1, 2]
-    a[0] = 1
-    return a[0]
+#@Phylanx(debug=True)
```
hkaiser: This will make flake bark at you.

parsa: Took care of it.

PR #474, opened 2018-05-30
Enable configuration of HPX in Python.

### LINKED ISSUE #469: Handling HPX configurations in Python
This is based on @ct-clmsn suggestion in PR #461.

HPX run-time should be made configurable in Python. Currently, as @hkaiser explains in the same PR, HPX is initialized with parameters set during the compilation of Phylanx when `phylanx` module is loaded.

rtohid:
Fixes #469.
Currently hpx initialized with configurations set at compile time; This commit allows dynamic initialization of HPX in Python through ::class::InitHPX.

### DIFF
--- a/python/phylanx/ast/transformation.py
+++ b/python/phylanx/ast/transformation.py
@@ -10,9 +10,12 @@ import re
 import phylanx
 import inspect
 from phylanx.exceptions import InvalidDecoratorArgumentError
-from phylanx.util import prange
 from .oscop import OpenSCoP
 from .utils import full_name, full_node_name, physl_fmt
+try:
+    from _phylanx import init_hpx_runtime
+except:
+    from _phylanxd import init_hpx_runtime
 
 et = phylanx.execution_tree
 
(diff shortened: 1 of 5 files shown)


python/src/init_hpx.cpp
```diff
@@ -145,8 +145,8 @@ get_command_line const& init_command_line()
 // This class initializes a console instance of HPX (locality 0).
 struct manage_global_runtime
 {
-    manage_global_runtime()
-      : running_(false), rts_(nullptr)
+    manage_global_runtime(std::vector<std::string> const config)
```
hkaiser: Did you mean for this to be a `std::vector<std::string> const&` instead?

python/phylanx/ast/transformation.py
```diff
@@ -10,9 +10,12 @@
 import phylanx
 import inspect
 from phylanx.exceptions import InvalidDecoratorArgumentError
-from phylanx.util import prange
 from .oscop import OpenSCoP
 from .utils import full_name, full_node_name, physl_fmt
+try:
+    from _phylanx import init_hpx_runtime
+except:
+    from _phylanxd import init_hpx_runtime
```
hkaiser: It might be a good idea to separate the initialization of HPX from the transformation code. Should this go into a separate module instead?

python/phylanx/ast/transformation.py
```diff
@@ -516,6 +519,36 @@ def convert_to_phylanx_type(v):
...
+    # run one thread only (for now)
+    "hpx.os_threads!=1",
+    # don't print diagnostics during forced terminate
+    "hpx.diagnostics_on_terminate!=0",
+    # disable the TCP parcelport
+    "hpx.parcel.tcp.enable!=0"
+    ]
+    hpx_init = False
+
+    def __init__(self, config=cfg):
```
hkaiser: I think that most of the settings above are mandatory for the functioning of HPX in the context of the Phylanx python frontend. The configuration settings passed to HPX should be the result of merging the user settings with the ones provided as a default.

stevenrbrandt: So I see that this sets the flags at runtime instead of compile time, but if this class is initialized in transformation.py, how does the programmer change the settings?

hkaiser:
@rtohid I think this can be closed now as these changes were part of #566? Is that correct?

rtohid:
@hkaiser, I have not included this in the pull request as it is not reliable. It works fine when number of threads is 1 but sometimes tests fail randomly, with "Child aborted", in other cases.

rtohid:
Dropped in favor of PR #675.

PR #475, opened 2018-05-30
Check that functions, e.g. "shape" are in really in numpy, however it's imported.

stevenrbrandt:
Check that functions, e.g. "shape" are in really in numpy, however it's imported, don't just guess the module name is "np" or "numpy"

### DIFF
--- a/python/phylanx/ast/transformation.py
+++ b/python/phylanx/ast/transformation.py
@@ -370,13 +370,13 @@ class PhySL:
             if a.value.id in self.defs:
                 s += a.value.id + full_node_name(a.value)
                 return s
-            elif a.value.id in fglobals and hasattr(fglobals[a.value.id],a.attr):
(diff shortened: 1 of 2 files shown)


python/phylanx/ast/transformation.py
```diff
@@ -16,6 +16,9 @@
 
 et = phylanx.execution_tree
 
+# globals used during the compilation
+fglobals = {}
```
hkaiser: What is the rational of making this 'global'?

stevenrbrandt: @hkaiser it is no longer global

python/phylanx/ast/transformation.py
```diff
@@ -602,6 +608,8 @@ class PhyTransformer(object):
         targets = {"PhySL": PhySL, "OpenSCoP": OpenSCoP}
 
         def __init__(self, f):
+            global fglobals
```
hkaiser: Looks like to be still `global` to me ...

PR #476, opened 2018-05-30
Allow for iteration over arrays using map

hkaiser:
- flyby: add unary Not to translation layer

### DIFF
--- a/phylanx/execution_tree/primitives/base_primitive.hpp
+++ b/phylanx/execution_tree/primitives/base_primitive.hpp
@@ -17,6 +17,8 @@
 #include <hpx/include/util.hpp>
 #include <hpx/include/serialization.hpp>
 
+#include <array>
+#include <cstddef>
 #include <cstdint>
 #include <initializer_list>
 #include <iosfwd>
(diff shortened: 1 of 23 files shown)


hkaiser:
@parsa, @stevenrbrandt is this what you need? Does it work for your use cases?

stevenrbrandt:
Interesting observation... the following code works perfectly:
```
import numpy as np
from phylanx.ast import Phylanx

@Phylanx()
def iter():
    a = hstack(1,2,3,4)
    for i in a:
        print(i)

iter()
```
But if I leave off the call to iter, I get a core dump at shutdown
```
1
2
3
4
terminate called after throwing an instance of 'std::runtime_error'
  what():  scoped_acquire::dec_ref(): thread state must be current!
Aborted (core dumped)
```

stevenrbrandt:
After more testing, I think the dec_ref() thing is not related to leaving off the iter().

stevenrbrandt:
I see no direct tests of the new functionality (though there is some new testing of lambda's). Maybe something like this should be added? (Note: it passes)
```
import numpy as np

a = np.linspace(0,5,6)
b = np.zeros(a.shape)

@Phylanx
def copyme(a,b):
    n = 0
    for i in a:
        b[n] = i
        n += 1
    return b

b = copyme(a,b)

assert np.all(a == b)
```

hkaiser:
Yes, the dec_ref problem is related to a bug in how we handle the GIL.

PR #481, opened 2018-06-04
Add --docs flag to physl interpreter

stevenrbrandt:
With this pull request, launching the physl command line tool with --docs causes physl to print all primitives/plugins and and exit.


parsa:
@stevenrbrandt Nice idea! Why just add this flag to the interpreter though?

stevenrbrandt:
@parsa I did just add the flag to the interpreter... not sure what you're asking.

parsa:
@stevenrbrandt As in why not add it as a general flag to all Phylanx applications

PR #483, opened 2018-06-07
Attempting to fix GIL issues while printing from PhySL

hkaiser:
@ct-clmsn, @stevenrbrandt please check whether this fixes the GIL related problems we see when using printing in PhySL.


hkaiser:
@stevenrbrandt: this should be fine now.

stevenrbrandt:
Works for me. I wish there was a way to add a test for this.

Issue #442, opened 2018-05-19
Creating new empty arrays


parsa:
There needs to be a mechanism that does something equivalent to `v = np.array((10, 10))` inside `@Phylanx` decorated functions.

hkaiser:
@parsa: we could modify `constant()` (it currently expects one or two arguments - a scalar value to use for initialization and an optional shape for the object returned) to accept just a shape argument in which case the returned object would stay uninitialized. 

@stevenrbrandt, @rtohid: do you think this could be implemented in the Python transformation?

stevenrbrandt:
@parsa It would take a little more than that since constant() can only create 1d arrays.

hkaiser:
constant() can create 0d, 1d, and 2d numeric values

stevenrbrandt:
OK, I see how it works now... the second arg can be an array.

hkaiser:
> OK, I see how it works now... the second arg can be an array.

IIRC, the first argument to `constant()` is the value to use to initialize the generated result with and the second argument specifies the requested size of the result. I think the second argument can be either a list of 0, 1, or 2 dimension sizes or another array which will be used to extract the desired shape.

PR #485, opened 2018-06-08
Resolve ticket 442, implementing np.array and np.zeros

stevenrbrandt:
Make np.array and np.zeros work


hkaiser:
This fixes #442

PR #487, opened 2018-06-19
Making sure 1d integer arrays get properly converted to node_data<int64_t>

### LINKED ISSUE #486: Array shape (n*1) error
The array shapes (n*1) between Phyton and Phylanx are not consistent:
```
from phylanx.ast import *
import numpy as np


@Phylanx(debug=True)
def phlyanx_array_test(a):
    print("Shape of array a: \n", np.shape(a))
    print("This is array a: \n", a)


y = np.array([[1], [1], [0]])
print("Shape of array y: \n", np.shape(y))
print("This is array y: \n", y)

phlyanx_array_test(y)
```
My expected two outputs should be the same but the print result is:

```
define(
    phlyanx_array_test,
    a,
    block(
        (
            cout(
                "Shape of array a: 
",
                shape(a)
            )
        ),
        (
            cout("This is array a: 
", a)
        )
    )
)
Shape of array y: 
 (3, 1)
This is array y: 
 [[1]
 [1]
 [0]]
Shape of array a: 
 '(3)
This is array a: 
 [1, 1, 0]

Process finished with exit code 0
```

But if the input array is m*n shape, the translation process from Python to Phylanx works fine. 
i.e. if you change value y into:
`y = np.array([[1, 2], [1, 2], [0,2]])`

Can you take a look at? Thanks.

hkaiser:
- flyby: don't package numpy arrays into Phylanx variables during Python -> C++ conversion

This fixes #486

### DIFF
--- a/phylanx/execution_tree/compiler/actors.hpp
+++ b/phylanx/execution_tree/compiler/actors.hpp
@@ -130,12 +130,14 @@ namespace phylanx { namespace execution_tree { namespace compiler
             primitive const* p = util::get_if<primitive>(&arg_);
             if (p != nullptr)
             {
+                arguments_type keep_alive(std::move(args));
+
                 // construct argument-pack to use for actual call
                 arguments_type params;
-                params.reserve(args.size());
-                for (auto && arg : args)
+                params.reserve(keep_alive.size());
+                for (auto const& arg : keep_alive)
                 {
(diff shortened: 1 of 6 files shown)


hkaiser:
@weilewei: please verify if this resolves your issue.

hkaiser:
@weilewei those failing tests should pass now.

PR #500, opened 2018-06-25
Set the cout precision for float numbers to 8 by default.

weilewei:
This pull request makes sure that Phylanx `cout` prints doubles with the same precision as Phython `print`.

### DIFF
--- a/src/ir/node_data.cpp
+++ b/src/ir/node_data.cpp
@@ -1293,7 +1293,7 @@ namespace phylanx { namespace ir
                 {
                     out << ", ";
                 }
-                out << T(r[i]);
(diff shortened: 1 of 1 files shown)


weilewei:
Thanks for your detailed suggestions! I also felt my code lacks flexibility when I was revising it, as the `std::setprecision` does not support different precisions for the different incoming values. I will be continuously working on it.

hkaiser:
We will abandon this PR as it is right now. The proper solution is to create a `format()` primitive that would allow to apply formatting strings to generated output, similar to Python's `"format to %8.5f" % d`.

PR #506, opened 2018-06-29
added summary stat performance test

ct-clmsn:
performance test added - accessing an array that stores indices needs performance tuning in phylanx. this test provides a python/numpy version and the phylanx version.

### DIFF
--- a/tests/regressions/python/sum_stat_perf_test.py
+++ b/tests/regressions/python/sum_stat_perf_test.py
@@ -15,12 +15,25 @@ def phy_run(y, k):
     for i in range(k):
         x[y[i]] += 1.0
 
+def py_run_mat(y, k):
+    x = np.zeros((k, k))
+    print(x.shape)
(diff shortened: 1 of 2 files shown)


hkaiser:
Thanks Chris, we'll investigate.

ct-clmsn:
@hkaiser updated with another edge case - would ya'll prefer if these tests were split out across multiple files?

hkaiser:
@ct-clmsn I think we're still fine, thanks!

ct-clmsn:
@hkaiser no problem; thanks!

hkaiser:
@ct-clmsn Actually, closer investigation shows, that the bad performance is caused by the (bad) PhySL code generated. Let's see how it will fare once the frontend etc. has been fixed.

ct-clmsn:
@hkaiser apologies on my end for not taking time to vet the PhySL layer. I'll add that to my code review process.

PR #507, opened 2018-06-29
simple lda initial impl

ct-clmsn:
initial import of the 'simple' latent dirichlet allocation


ct-clmsn:
data sets added. going to sort out the proper return value in a bit and the inferencing function. what's been committed right now is the training function.

ct-clmsn:
a couple of small implementation quirks are annotated in the source as a comment with the keyword 'NOTE:' at the beginning of the comment.

ct-clmsn:
@hkaiser yep! will get on it later today.

ct-clmsn:
@hkaiser alright, fixed a bunch of scripting errors; renamed some of those unused variables, and did a quick port to numpy of the algorithm just to make sure it works. it *should* be in a lot better shape now. apologies for the sloppy shape of this PR.

PR #511, opened 2018-07-03
Make sure, all eval-actions on primitives are run asynchronously if run on a non-HPX thread

hkaiser:
- flyby: enable forcing sync execution for debug purposes


hkaiser:
@parsa this should solve the issue you reported with running Phylanx remotely.

PR #525, opened 2018-07-11
Fixing list slicing to properly handle [::-1] 

### LINKED ISSUE #524: list's looping problem
I am not quite sure how to exactly describe this problem, but I think it could be sorted to list's indexing problem. 

Here is what I did. I first tried to reverse a list and then excluded the last item in the list. The following code works well in pure Python but does not work in Phylanx. The python code printed `[7, 8] [5, 6] [3, 4]`, which is what I expected, but if I put `@Phylanx(debug=True)` just one line above the `def foo(x)`, it does not print anything.
```py
def foo(x):
    local_x = x
    for ele in local_x[::-1][:-1]:
        print(ele)


a = [[1, 2], [3, 4], [5, 6], [7, 8]]
foo(a)
```
FYI, I attached PhySL code here.
```
define(
    foo,
    x,
    block(
        define(local_x, x),
        map(
            lambda(
                ele,
                block(
                    (
                        cout(ele)
                    )
                )
            ),
            slice(
                slice(
                    local_x,
                    make_list(
                        0,
                        nil,
                        -(1)
                    )
                ),
                make_list(
                    0,
                    -(1)
                )
            )
        )
    )
)
```

I found an alternative to solve this list's indexing or looping issue. The code
```py
@Phylanx(debug=True)
def foo(x):
    local_x = x
    # for ele in local_x[::-1][:-1]:
    #     print(ele)
    for i in range(len(local_x) - 1, 0, -1):
        print(local_x[i])


a = [[1, 2], [3, 4], [5, 6], [7, 8]]
foo(a)
```
returns
```
list(7, 8)
list(5, 6)
list(3, 4)
```

FYI, I attached PhySL code here.
```
define(
    foo,
    x,
    block(
        define(local_x, x),
        map(
            lambda(
                i,
                block(
                    (
                        cout(
                            slice(local_x,i)
                        )
                    )
                )
            ),
            range(
                (
                    len(local_x)-1
                ),
                0,
                -(1)
            )
        )
    )
)
```

hkaiser:
- this fixes #524

### DIFF
--- a/src/plugins/matrixops/slicing_operation.cpp
+++ b/src/plugins/matrixops/slicing_operation.cpp
@@ -86,7 +86,7 @@ namespace phylanx {namespace execution_tree {    namespace primitives {
         }
         else if (step < 0)
         {
-            for (std::int64_t i = start; i > stop; i += step)
+            for (std::int64_t i = stop - 1; i >= start; i += step)
             {
                 result.push_back(i);
             }
@@ -552,7 +552,7 @@ namespace phylanx {namespace execution_tree {    namespace primitives {
         auto idx_end = index_list.end();
 
         // extract elements from list at given indices
-        if (row_start <= row_stop)
+        if (row_start <= row_stop && step > 0)
         {
             auto list_it = list.begin();
             auto list_end = list.end();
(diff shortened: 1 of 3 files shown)


hkaiser:
@weilewei could you please verify whether this fixes your issue?

PR #534, opened 2018-07-17
Move the functionality of the set primitive to the store primitive

justwagle:
This pull request moves the functionality of the set primitive to the store primitive. 
Uses:
```
store (input, list(start, stop, step), value_to_set)  for 1d
store (input, list(row_start, row_stop, step), list(col_start, col_stop, step), value_to_set) for 2d
```

### DIFF
--- a/src/execution_tree/primitives/define_variable.cpp
+++ b/src/execution_tree/primitives/define_variable.cpp
@@ -97,6 +97,63 @@ namespace phylanx { namespace execution_tree { namespace primitives
         p->store(hpx::launch::sync, std::move(val));
     }
 
+    ///////////////////////////////////////////////////////////////////
+
+    void define_variable::store_set_1d(
+        phylanx::ir::node_data<double>&& data, std::vector<int64_t>&& list)
+    {
+        if (!valid(operands_[1]))
+        {
+            HPX_THROW_EXCEPTION(hpx::invalid_status,
+                "define_variable::store",
+                execution_tree::generate_error_message(
+                    "the variable associated with this define has not been "
+                    "initialized yet",
+                    name_, codename_));
+        }
+
+        primitive* p = util::get_if<primitive>(&operands_[1]);
+        if (p == nullptr)
+        {
+            HPX_THROW_EXCEPTION(hpx::invalid_status,
+                "define_variable::store",
+                execution_tree::generate_error_message(
+                    "the variable associated with this define has not been "
+                    "properly initialized",
+                    name_, codename_));
+        }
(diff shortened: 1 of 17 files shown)


ct-clmsn:
@justwagle @rtohid great work ya'll!

src/execution_tree/primitives/define_variable.cpp
```diff
@@ -97,6 +97,63 @@ namespace phylanx { namespace execution_tree { namespace primitives
         p->store(hpx::launch::sync, std::move(val));
     }
 
+    ///////////////////////////////////////////////////////////////////
+
+    void define_variable::store_set_1d(
```
hkaiser: Are these functions (on `define_variable`) ever called?

src/execution_tree/primitives/store_operation.cpp
```diff
@@ -45,31 +51,344 @@ namespace phylanx { namespace execution_tree { namespace primitives
...
+            result.reserve((actual_start - actual_stop - step) / (-step));
+            for (std::int64_t i = actual_start; i > actual_stop; i += step)
+            {
+                result.push_back(i);
+            }
+        }
+
+        if (result.empty())
+        {
+            HPX_THROW_EXCEPTION(hpx::bad_parameter,
```
hkaiser: Is this really an error? I think that slice parameters are allowed to result in empty slices, aren't they?

tests/unit/execution_tree/primitives/store_operation.cpp
```diff
@@ -9,38 +10,362 @@
...
+    std::string const code = R"(block(
+        define(a,57.7),
+        store(a,42.0),
+        a
+    ))";
+
+    auto result =
+        phylanx::execution_tree::extract_numeric_value(compile_and_run(code));
+
+    HPX_TEST_EQ(result[0], 42.0);
```
hkaiser: Please add a check for `result.size() == 1` before blindly accessing `result[0]`. Same needs to happens for the for tests below.

justwagle: done.

src/execution_tree/primitives/store_operation.cpp
justwagle: @hkaiser This was leftover from the previous implementation of set. I have removed the error and also simplified the function.

src/execution_tree/primitives/define_variable.cpp
justwagle: @hkaiser I just tried it without the overloads and It worked fine without it. Will remove it in the next commit.

PR #536, opened 2018-07-24
adding append function to the list

weilewei:
Updating `append` function to the list. @hkaiser Could you please verify this version? Thank you so much.

### DIFF
--- a/phylanx/plugins/listops/append_operation.hpp
+++ b/phylanx/plugins/listops/append_operation.hpp
@@ -1,0 +1,57 @@
+// Copyright (c) 2018 Weile Wei
+// Copyright (c) 2018 R. Tohid
+//
+// Distributed under the Boost Software License, Version 1.0. (See accompanying
+// file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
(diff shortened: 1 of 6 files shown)


ct-clmsn:
@weilewei have you tried an expression like B = [ A, A, A ] where A is a single type of dense matrix?

weilewei:
@ct-clmsn Thanks for your attention. Do you mean that appending a non-zero array into a list where contains multiple same type arrays?

I tried a simple code as follows and used `np.array` as an example.
```py
from phylanx.ast import *
import numpy as np


@Phylanx(debug=True)
def foo(x, y):
    z = append(x, y)
    print(z)


a = [np.array([[1.0, 2.0], [3.0, 4.0]]), np.array([[5.0, 6.0], [7.0, 8.0]])]
b = np.array([[42.0, 42.0], [42.0, 42.0]])
foo(a, b)
```
which yields,
```
define(
    foo,  x, y,
    block(
        define(z, append(x, y)),
        (cout(z))
    )
)
list([[1, 2], [3, 4]], [[5, 6], [7, 8]], [[42, 42], [42, 42]])
```
Does it look good to you? If so, I can add this to testing case later. Thanks.

ct-clmsn:
@weilewei perfect!

PR #544, opened 2018-08-01
csc hdf5 blaze

ct-clmsn:
attempt to provide csc-hdf5 storage support for blaze::CompressedMatrix<double, blaze::columnMajor> structures. used the [h5sparse](https://github.com/appier/h5sparse) python library as a guidepost.


ct-clmsn:
need to provide a unit test.

ct-clmsn:
would it be possible to pass a the variable name as a string into the hdf5 file as the hdf5 group name for the input matrix? this might be something the IR analyzer could do?

PR #554, opened 2018-08-04
Making sure compiler can be invoked more than once

### LINKED ISSUE #543: Calling compile more than once in a row references wrong code
Calling `compile` more than once in a row using the same snippets object makes all code objects refer to the PhySL code compiled last,

hkaiser:
Making sure compiler can be invoked more than once in a row without causing wrong code to be referenced.

This fixes #543


hkaiser:
Investigating the failing test has shown that more work is needed on this.

stevenrbrandt:
@hkaiser do you have a sample test code that illustrates what's wrong?

hkaiser:
@stevenrbrandt the regression test `replace_names_374` fails coincidentally thus revealing a flaw in the current PR.

stevenrbrandt:
@hkaiser do we have an example which shows this failure?

PR #555, opened 2018-08-06
Adding dictionaries

weilewei:
Adding dictionary primitive for Phylanx. Could you please review it? Thanks. @parsa @hkaiser


src/execution_tree/primitives/dictionary.cpp
```diff
@@ -0,0 +1,140 @@
...
+                    "phylanx::ir::dictionary",
+                    "unsupported none 0-dimension node data type");
+            }
+        }
+
+        case 3:    // std::string
+        {
+            std::string val_temp = phylanx::util::get<3>(val);
+            boost::hash<std::string> hash_string;
+            return hash_string(val_temp);
```
hkaiser: You might want to try avoiding copying the string here.

src/execution_tree/primitives/dictionary.cpp
```diff
@@ -0,0 +1,140 @@
...
+        case 0:    // ast::nil
+            HPX_THROW_EXCEPTION(hpx::bad_parameter,
+                "phylanx::ir::dictionary",
+                "unsupported nil data type");
+
+        case 1:    // phylanx::ir::node_data<std::uint8_t>
+        {
+            phylanx::ir::node_data<std::uint8_t> val_temp =
+                phylanx::util::get<1>(val);
+            switch (val_temp.num_dimensions())
```
hkaiser: This creates a copy of the `node_data` object. It might not be worth avoiding, but you can try ;-) (same is true below for `node_data<std::int64_t>` and `node_data<double>`).

src/execution_tree/primitives/dictionary.cpp
```diff
@@ -0,0 +1,140 @@
...
+                    "unsupported none 0-dimension node data type");
+            }
+        }
+
+        case 5:    // primitive
+        {
+            hash hash_primitive;
+            return hash_primitive(val);
+        }
+
```
hkaiser: This case should throw as well.

src/execution_tree/primitives/dictionary.cpp
```diff
@@ -0,0 +1,140 @@
...
+                "unsupported dictionary data type");
+        }
+        default:
+            break;
+        }
+
+        HPX_THROW_EXCEPTION(hpx::bad_parameter,
+            "phylanx::ir::dictionary",
+            "unsupported data type");
+    }
```
hkaiser: You might want to try refactoring this such that you throw an exception in one spot only while still generating a reasonable error message.

phylanx/execution_tree/primitives/primitive_argument_type.hpp
```diff
@@ -322,6 +325,14 @@ namespace phylanx { namespace execution_tree
...
         {}
 
+        primitive_argument_type(ir::dictionary const& val)
+          : argument_value_type{val}
+        {}
+        primitive_argument_type(ir::dictionary&& val)
+          : argument_value_type{std::move(val)}
+        {}
+
+
```
hkaiser: Please limit white space to one line.

phylanx/ir/dictionary.hpp
```diff
@@ -0,0 +1,43 @@
...
+// Distributed under the Boost Software License, Version 1.0. (See accompanying
+// file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+#if !defined(PHYLANX_DICTIONARY)
+#define PHYLANX_DICTIONARY
+
+#include <phylanx/config.hpp>
+#include <phylanx/util/variant.hpp>
+
+#include <hpx/include/serialization.hpp>
```
hkaiser: Why is this header #include needed?

src/execution_tree/primitives/dictionary.cpp
```diff
@@ -0,0 +1,140 @@
...
+#include <cstddef>
+#include <cstdint>
+#include <memory>
+#include <string>
+#include <utility>
+#include <vector>
+
+#include <phylanx/execution_tree/primitives/primitive_argument_type.hpp>
+#include <phylanx/util/variant.hpp>
+#include <boost/functional/hash.hpp>
```
hkaiser: Please keep the Phylanx headers grouped at the top of the file.

phylanx/ir/dictionary.hpp
weilewei: @hkaiser I noticed that you included three `#include` headers in your comment, do you specifically point to `#include <hpx/include/serialization.hpp>`? I thought I just copied them from somewhere but I like to investigate it further.

src/execution_tree/primitives/dictionary.cpp
```diff
@@ -0,0 +1,140 @@
...
+
+///////////////////////////////////////////////////////////////////////////////
+namespace phylanx { namespace execution_tree { namespace primitives {
+
+}}}
+
```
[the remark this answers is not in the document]
hkaiser: This is an empty namespace declaration, I don't see how this could introduce the `primitive_argument_type` name...

src/execution_tree/primitives/dictionary.cpp
```diff
@@ -0,0 +1,140 @@
...
+        case 2:    // phylanx::ir::node_data<std::int64_t>
+        {
+            phylanx::ir::node_data<std::int64_t> val_temp =
+                phylanx::util::get<2>(val);
+            switch (val_temp.num_dimensions())
+            {
+            case 0:
+                boost::hash<__int64_t> hash_int64_t;
+                return hash_int64_t(val_temp[0]);
```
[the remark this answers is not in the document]
hkaiser: That would be appreciated, yes. Thanks!

phylanx/ir/dictionary.hpp
hkaiser: I think `#include <hpx/include/serialization.hpp> ` is not needed here as your header file does not refer to anything related to serialization. We should try limiting the used `#include`s to what is needed by a particular file.

src/execution_tree/primitives/dictionary.cpp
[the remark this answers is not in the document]
weilewei: The reason for putting this empty namespace declaration here is to break the dependency which is related to `primitive_argument_type.hpp`. @parsa Do you have any suggestions here?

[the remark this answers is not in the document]
weilewei: This empty namespace is deleted in the upcoming new PR. I looked at the `.hpp` file. By mistake.

src/execution_tree/primitives/dictionary.cpp
```diff
@@ -0,0 +1,138 @@
...
+                    "cannot hash none 0-dimension node data type");
+            }
+        }
+
+        case 3:    // std::string
+        {
+            std::string val_temp = phylanx::util::get<3>(val);
+            boost::hash<std::string> hash_string;
+            return hash_string(std::move(val_temp));
+        }
```
hkaiser: This will still create a copy. Also, the `std::move` will have no effect as the hash function probably takes its argument by `const&`. Why not write it as:
```
boost::hash<std::string> hash_string; 
return hash_string(phylanx::util::get<3>(val)); 
```
?

src/execution_tree/primitives/dictionary.cpp
```diff
@@ -0,0 +1,138 @@
...
+
+///////////////////////////////////////////////////////////////////////////////
+namespace std {
+    std::size_t hash<phylanx::util::recursive_wrapper<
+        phylanx::execution_tree::primitive_argument_type>>::
+    operator()(argument_type const& s) const noexcept
+    {
+        phylanx::execution_tree::primitive_argument_type const& val = s.get();
+        std::size_t result;
+        switch (val.index())
```
hkaiser: This local variable is not needed, is it?

src/execution_tree/primitives/dictionary.cpp
```diff
@@ -0,0 +1,138 @@
...
+                phylanx::util::get<1>(val);
+            switch (val_temp.num_dimensions())
+            {
+            case 0:
+                boost::hash<std::uint8_t> hash_uint8_t;
+                return hash_uint8_t(std::move(val_temp[0]));
+            case 1:
+                HPX_FALLTHROUGH;
+            case 2:
+                HPX_THROW_EXCEPTION(hpx::bad_parameter,
```
hkaiser: You might want to add a `default` case (same below)

src/execution_tree/primitives/dictionary.cpp
```diff
@@ -0,0 +1,138 @@
...
+
+        default:
+            break;
+        }
+
+        HPX_THROW_EXCEPTION(hpx::bad_parameter,
+            "phylanx::ir::dictionary",
+            "dictionary's key object holds unhashable data type");
+    }
+}
```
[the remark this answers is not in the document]
parsa: @weilewei You can use `detail::get_primitive_argument_type_name` to extract the name of the type in the variant.

PR #560, opened 2018-08-07
Modify the patterns.cpp function so that it can print sqrt.

stevenrbrandt:
Currently, the "./build/bin/physl --docs" command fails to find sqrt(), log(), etc. This PR fixes that.


hkaiser:
@stevenrbrandt please fix the inspect/formatting errors reported. Otherwise we'll not be able to merge this.

PR #566, opened 2018-08-14
Refactor Frontend

### LINKED ISSUE #442: Creating new empty arrays

### LINKED ISSUE #522: Multi-dimension np.array cannot be created in Phylanx
I expect the following code works
```py
@Phylanx
def foo():
    x = np.array([[1, 1]])
    return x


a = foo()
```
However, it complains
```py
RuntimeError: <unknown>(6, 8): hstack:: primitive_argument_type does not hold a numeric value type (type held: 'phylanx::ir::range'): HPX(bad_parameter)
```
However, one dimensional np.array works fine inside the Phylanx. For example, if you change `x` to `np.array([1, 1])`

One potential testing code could be as follows (I am unsure if it helps)
```py
@Phylanx
def foo():
    x = np.array([[1, 1]])
    return x


a = foo()
v = np.array([[1, 1]]) == foo()
assert (True == v.any())
```

### LINKED ISSUE #530: Set primitive is not enabled for augmented assignment
`Set primitive` is not enabled for augmented assignment, as it is translated to `store` instead.

I expect `[57  1]` to be returned by the following code, however, it prints `[57. 56.]`.
```py
@Phylanx(debug=True)
def foo():
    local_a = np.array((2, 1))
    local_a[0] += 55
    return local_a


c = foo()
print(c)
```
FYI, I attached the PhySL code
```
define(
    foo,
    block(
        define(
            local_a,
            hstack(2, 1)
        ),
        store(local_a, local_a+55),
        local_a
    )
)
```

Therefore, the following testing code fails.
```py
@Phylanx(debug=True)
def foo():
    local_a = np.array((2, 1))
    local_a[0] += 55
    return local_a


c = foo()
v = c == np.array((57, 1))
assert (True == v.all())
```

### LINKED ISSUE #549: Missing np.add
I expect the result `[[2 4] [6 8]]` to be returned by the following code
```py
import numpy as np
from phylanx.ast import *


@Phylanx
def foo():
    A = np.array([[1, 2], [3, 4]])
    B = np.array([[1, 2], [3, 4]])
    C = np.add(A, B)
    print(C)


foo()
```
However, it complains
```
RuntimeError: Error! Incomplete parse: line 1:
define$6$0(foo$6$0, block$6$0(define$7$4(A$7$4, hstack(make_list$7$18(1,2),make_list$7$26(3,4))), define$8$4(B$8$4, hstack(make_list$8$18(1,2),make_list$8$26(3,4))), define$9$4(C$9$4, A$9$15, B$9$18)), (cout$10$4(C$10$10))))
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
: HPX(bad_parameter)
```
Therefore, `np.add` is not existing.

rtohid:
In addition to refactoring the frontend, this pull request resolves the following issues:
- fixes #442 
- fixes #522 
- fixes #530
- fixes #549


hkaiser:
Marvelous! Many thanks!

PR #569, opened 2018-08-20
adding dictionary case for base_primitive.cpp

weilewei:
adding dictionary case supports in base_primitive.cpp.
Could you please review these changes? @parsa @hkaiser

PR #573, opened 2018-08-24
Create dictionary object

weilewei:
Enabling Phylanx to create `dict` object. Please review and comment @parsa @hkaiser Thanks!

Note, this pull request is sitting on top of previous pull request #569

### DIFF
--- a/phylanx/execution_tree/primitives/base_primitive.hpp
+++ b/phylanx/execution_tree/primitives/base_primitive.hpp
@@ -144,6 +144,9 @@ namespace phylanx { namespace execution_tree
         primitive_argument_type&& val,
         std::string const& name = "",
         std::string const& codename = "<unknown>");
+    PHYLANX_EXPORT bool is_dictionary_operand(
+        primitive_argument_type const& val);
+
     template <typename T>
     primitive_argument_type extract_ref_value(ir::node_data<T> const& val)
     {
--- a/phylanx/execution_tree/primitives/base_primitive.hpp
+++ b/phylanx/execution_tree/primitives/base_primitive.hpp
@@ -136,6 +136,14 @@ namespace phylanx { namespace execution_tree
         std::string const& name = "",
         std::string const& codename = "<unknown>");
 
+        PHYLANX_EXPORT ir::dictionary extract_dict_value(
+                primitive_argument_type const& val,
+                std::string const& name = "",
+                std::string const& codename = "<unknown>");
+        PHYLANX_EXPORT ir::dictionary extract_dict_value(
+                primitive_argument_type && val,
+                std::string const& name = "",
+                std::string const& codename = "<unknown>");
     template <typename T>
     primitive_argument_type extract_ref_value(ir::node_data<T> const& val)
     {
--- a/phylanx/execution_tree/primitives/base_primitive.hpp
+++ b/phylanx/execution_tree/primitives/base_primitive.hpp
@@ -136,14 +136,14 @@ namespace phylanx { namespace execution_tree
         std::string const& name = "",
         std::string const& codename = "<unknown>");
 
-        PHYLANX_EXPORT ir::dictionary extract_dict_value(
-                primitive_argument_type const& val,
-                std::string const& name = "",
-                std::string const& codename = "<unknown>");
-        PHYLANX_EXPORT ir::dictionary extract_dict_value(
-                primitive_argument_type && val,
-                std::string const& name = "",
-                std::string const& codename = "<unknown>");
+    PHYLANX_EXPORT ir::dictionary extract_dictionary_value(
+        primitive_argument_type const& val,
+        std::string const& name = "",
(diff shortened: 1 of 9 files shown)


src/plugins/listops/dictionary_operation.cpp
```diff
@@ -0,0 +1,120 @@
...
+            {
+                HPX_THROW_EXCEPTION(hpx::bad_parameter,
+                    "dictionary_operation::generate_dict",
+                    util::generate_error_message("dict_operation needs exactly "
+                                                 "two parameters for key-value",
+                        name_, codename_));
+            }
+            dict[phylanx::execution_tree::extract_list_value(*it).args()[0]] =
+                phylanx::execution_tree::extract_list_value(*it).args()[1];
+        }
```
hkaiser: I think you should call `extract_list_value()` only once for each element.

src/plugins/listops/dictionary_operation.cpp
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    ///////////////////////////////////////////////////////////////////////////
+    primitive_argument_type dict_operation::generate_dict(
+        std::vector<phylanx::ir::range>&& args) const
+    {
+        phylanx::ir::dictionary dict;
+
+        ir::range args_list =
+            extract_list_value_strict(std::move(args[0]), name_, codename_);
```
hkaiser: If `dict()` is called without arguments, then `args.size() == 0`. This would lead to an index out of bounds error when accessing `args[0]`. Please add a test for this case.

tests/unit/plugins/listops/dictionary_operation.cpp
```diff
@@ -0,0 +1,139 @@
...
+    std::string const& expected_str)
+{
+    HPX_TEST_EQ(compile_and_run(code), compile_and_run(expected_str));
+}
+
+//////////////////////////////////////////////////////////////////////////
+
+void test_dict_operation()
+{
+    char const* const code = "'('(1, 2), '(3, 4))";
```
hkaiser: Please use `list()` instead of `'()`. We will remove the quoting operator in the future.

src/plugins/listops/dictionary_operation.cpp
```diff
@@ -0,0 +1,124 @@
...
+                    .size() != 2)
+            {
+                HPX_THROW_EXCEPTION(hpx::bad_parameter,
+                    "dictionary_operation::generate_dict",
+                    util::generate_error_message("dict_operation needs exactly "
+                                                 "two parameters for key-value",
+                        name_, codename_));
+            }
+            auto const& key_value = phylanx::execution_tree::extract_list_value(*it);
+            dict[key_value.args()[0]] = key_value.args()[1];
```
hkaiser: This still calls `extract_list_value` more than once...

tests/unit/plugins/listops/dictionary_operation.cpp
```diff
@@ -0,0 +1,186 @@
...
+
+    HPX_TEST(exception_thrown);
+}
+
+void test_dict_empty_arg_operation()
+{
+    char const* const code = "";
+
+    phylanx::ir::range key_value =
+        phylanx::execution_tree::extract_list_value(compile_and_run(code));
```
hkaiser: What is the purpose of compiling an empty string?

phylanx/execution_tree/primitives/base_primitive.hpp
```diff
@@ -136,6 +136,14 @@ namespace phylanx { namespace execution_tree
...
         std::string const& codename = "<unknown>");
 
+    PHYLANX_EXPORT ir::dictionary extract_dictionary_value(
+        primitive_argument_type const& val,
+        std::string const& name = "",
+        std::string const& codename = "<unknown>");
+    PHYLANX_EXPORT ir::dictionary extract_dictionary_value(
+        primitive_argument_type&& val,
+        std::string const& name = "",
+        std::string const& codename = "<unknown>");
```
hkaiser: You also might want to declare the `is_dictionary_operand` function you defined.

tests/unit/plugins/listops/dictionary_operation.cpp
weilewei: To test the empty argument. I will revise this test in the new update.

src/plugins/listops/dictionary_operation.cpp
```diff
@@ -43,17 +43,14 @@ namespace phylanx { namespace execution_tree { namespace primitives
     {
         phylanx::ir::dictionary dict;
 
+        if (args.size() == 0)
+            return primitive_argument_type(dict);
```
hkaiser: You might want to move the `dict` object here.

hkaiser: Same is true for all other returns in this function.

Issue #441, opened 2018-05-18
Doc strings for primitives


stevenrbrandt:
It's getting hard to keep track of what primitives exist. My suggestion is that the base class for primitives should require a doc string (that can be printed at runtime) describing the primitive and how to use it. Further, there should be an option to physl to print all primitives and their doc strings or just a specific one.

hkaiser:
The physl interpreter has been extended to print some docs in #481.

hkaiser:
@parsa has added some documentation for the primitives here: https://github.com/STEllAR-GROUP/phylanx/wiki/Primitives-Documentation

stevenrbrandt:
@parsa What is lacking here is a way to know if the wiki is up to date. The script below can be used to check, but if you run it too frequently the wiki may fail to serve up a 200. Missing primitives (according to my script) currently include:
```
['len', '__le', '__lt', '_linear_solver', 'linearmatrix', 'list', 'lra', 'mean', '__mul', '__ne', '__or', 'set_row', 'slice_row', 'set_seed_action', 'square_root', '__sub', '__minus', '__not']

```
One thing the script does not currently have the ability to check is whether something is documented in the wiki, but not present in physl.
```
import re, os, requests
special = {
    "add_dim":"add_dimension" # This one name is different in the wiki
    }
missing = []
with os.popen("./build/bin/physl --docs","r") as fd:
    for line in fd.readlines():
        g = re.match('^pattern:\s+(\w+)',line)
        if g:
            key = g.group(1)
            if key in special:
                key = special[key]
            r = requests.get("https://github.com/STEllAR-GROUP/phylanx/wiki/"+key)
            if r.status_code != 200:
                key += "-Primitive"
                r = requests.get("https://github.com/STEllAR-GROUP/phylanx/wiki/"+key)
            print(g.group(1),key,r.status_code)
            if r.status_code != 200:
                missing += [g.group(1)]

print("=" * 50)
print(missing)
```

stevenrbrandt:
The following script does a better job of checking the wiki. It makes a single http request, and can find things that are present in the wiki but not physl:
```
import requests, re, os

# Get Primitives that Physl knows about
physl = {}
with os.popen("./build/bin/physl --docs","r") as fd:
    for line in fd.readlines():
        g = re.match('^pattern:\s+(\w+)',line)
        if g:
            key = g.group(1)
            if key == "add_dim":
                key = "add_dimension"
            physl[key] = 1

[...]

PR #577, opened 2018-08-29
Adding help

stevenrbrandt:
This pull request adds docstrings to Phylanx. Thus, the code below
```
from phylanx.util import phyhelp

phyhelp("lra")
```
Shows a help string for the "lra" primtive.
Note that the tests will check docstrings for formatting, hopefully ensuring good docs in the future.


src/execution_tree/primitives/assert_condition.cpp
```diff
@@ -37,7 +37,13 @@ namespace phylanx { namespace execution_tree { namespace primitives
...
         hpx::util::make_tuple("assert",
             std::vector<std::string>{"assert(_1)"},
-            &create_assert_condition, &create_primitive<assert_condition>)
+            &create_assert_condition, &create_primitive<assert_condition>,
+            "cond\n"
+            "Args:\n"
+            "\n"
+            "    cond (boolean expression) : if not true, raise an AssertionError\n"
+            "Returns:\n"
+            )
```
hkaiser: Have you considered using raw strings for the documentation string? This would remove the need to explicitly place `\n` and similar in the strings. For instance the above snippet would read (similar to pythons `'''` string notation):
```
        hpx::util::make_tuple("assert", 
            std::vector<std::string>{"assert(_1)"}, 
            &create_assert_condition, &create_primitive<assert_condition>) 
            R"(
cond 
Args:
    cond (boolean expression) : if not true, raise an AssertionError
Returns: 
   <nothing>
)"
```

tools/CMakeLists.txt
```diff
@@ -2,6 +2,7 @@
 #
 # Distributed under the Boost Software License, Version 1.0. (See accompanying
 # file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+cmake_minimum_required(VERSION 3.7)
```
hkaiser: Shouldn't this go into the main CMakeFile.txt?

tests/unit/python/execution_tree/check_docs.py
```diff
@@ -0,0 +1,94 @@
...
+    v = argcount[a]
+    if "any" in v:
+        continue
+    if re.search(r"@Deprecated@", v["help"]):
+        continue
+    if v["max"] not in v["patterns"]:
+        print("=" * 50)
+        print("Error: argument / pattern mismatch")
+        print(a, v["max"], v["patterns"])
+        print(v["help"])
```
hkaiser: Should we add this file to the tools directory instead?

src/execution_tree/primitives/assert_condition.cpp
stevenrbrandt: I didn't know about them.

tools/CMakeLists.txt
stevenrbrandt: Possibly. I was getting an error message from that file saying it was required.

tests/unit/python/execution_tree/check_docs.py
stevenrbrandt: As long as it gets run before a pull request is accepted, that would be reasonable.

hkaiser: Ok. In this case it also needs to be integrated into the circleci build, similar to fake8 and inspect. See here how that is done: https://github.com/STEllAR-GROUP/phylanx/blob/master/.circleci/config.yml#L101-L135.

tools/CMakeLists.txt
hkaiser: You probably got an error while using this CMakeLists.txt as the main cmake file. Please always use the one in the root directory as your target,

hkaiser:
This fixes #441

stevenrbrandt:
@hkaiser can you take a look at this? All that's blocking it is appveyor.

hkaiser:
@stevenrbrandt we can ignore appveyor at this point.

PR #579, opened 2018-09-01
Find MPI package before generating phylanxrun.py

khuck:
If find_package(MPI...) isn't called before phylanxrun.py is
generated, MPI CMake macros aren't defined and the generated
script is incapable of running MPI tests.


CMakeLists.txt
```diff
@@ -201,6 +201,7 @@ if(PHYLANX_WITH_TESTS_BENCHMARKS OR PHYLANX_WITH_TESTS_REGRESSIONS OR PHYLANX_WI
 endif()
 
 # Configure phylanxrun.py
+find_package(MPI QUIET)
```
hkaiser: This should be executed only if HPX was configured with MPI itself (`HPX_WITH_PARCELPORT_MPI` is equal to `On`).

PR #580, opened 2018-09-02
Propagate type of node_data for comparison operations

hkaiser:
- flyby: avoid signed/unsigned mismatch warnings

The comparison operations before either returned a `node_data<bool>` or, if requested, a `node_data<double>`. This patch let's it return a `node_data<T>` if requested instead.


hkaiser:
@scheherzade thanks for the fix!

PR #591, opened 2018-09-11
updating existing ALS algorithm codes

scheherzade:


scheherzade:
@khuck Yes, my mistake I changed the PhySL code but forgot to change how I should be printing it. I'm working on it.

PR #594, opened 2018-09-17
Fixing the problem with printing the output of the ALS algorithm

scheherzade:


examples/algorithms/als/als.cpp
```diff
@@ -3,6 +3,7 @@
 //   Distributed under the Boost Software License, Version 1.0.0. (See accompanying
 //   file LICENSE_1_0.0.txt or copy at http://www.boost.org/LICENSE_1_0.0.txt)
 
+#include <phylanx/execution_tree/primitives/slice_node_data.hpp>
```
hkaiser: This header should be added to `phylanx/execution_tree/primitives.hpp` instead (together with the other headers related to slicing...)

PR #597, opened 2018-09-18
Removing the set primitive

justwagle:
This pull request removes the set primitive as its functionalities have now been incorporated into the store primitive.


examples/algorithms/als/als.cpp
```diff
@@ -77,7 +77,7 @@ char const* const als_code = R"(
                             store(p_u, __ne(conf_u, 0.0, true)),
                             store(A, dot(dot(transpose(Y), c_u), Y)+ YtY),
                             store(b, dot(dot(transpose(Y),(c_u + I_i)), transpose(p_u))),
-                            set_row(X, u, u + 1, 1, dot(inverse(A), b)),
+                            store(slice(X, list(u, u + 1, 1),nil), dot(inverse(A), b)),
```
hkaiser: This could be written as:
```
store(slice(X, u), dot(inverse(A), b)), 
```
which should be slightly more efficient. OTOH, `slice(X, u)` returns a scalar (as does `slice(X, list(u))`), while `slice(X, list(u, u+1, 1))` returns a vector (or matrix - depending on the type of `X`).

justwagle: I check with  ```store(slice(X, u), dot(inverse(A), b)),``` and the results were different from the expected result. However, both ```store(slice(X, list(u)), dot(inverse(A), b)),``` and ```store(slice(X, list(u, u+1, 1), nil), dot(inverse(A), b)),``` produced the expected results.

hkaiser:
@justwagle please rebase this PR for us to be able to merge.

PR #598, opened 2018-09-18
Make the help string Hartmut used in his last tutorial work.

stevenrbrandt:
When Hartmut gave his talk about creating plugins, the string he used for the help string would not have passed the checker. IMHO, it should have. This pull request fixes the issue.


hkaiser:
@stevenrbrandt is there a way to automatically transform all help strings to C++ raw strings? I think this would a) simplify writing those, and b) increase readability.

PR #600, opened 2018-09-24
Fmap primitive

taless474:
This PR renames `map` primitive to `fmap`. This is done because the `map` function in Python returns an iterable instead of a list, as Phylanx `map` does.


hkaiser:
@taless474 do you plan to implement a 'real' `map` primitive now (i.e. one that does return an iterable range)?

taless474:
Yes, I will definitely work on it.

Issue #595, opened 2018-09-18
Invalid code (PhySL) generation for multi-value return statements.


rtohid:
If a function returns a tuple or, equivalently, multiple values, only the last value is returned in PhySL.
```py
from phylanx import Phylanx
import numpy as np

@Phylanx(debug=True)
def foo(x, y):
    return (x, y)

x = 1
y = 2
print(foo(x, y))
```
corresponding PhySL which returns `2`:
```
define(foo, x, y, x, y)
```

hkaiser:
Should'nt this be:

hkaiser:
@rtohid any news on this?

hkaiser:
This was fixed by merging #595

PR #613, opened 2018-10-09
Fixing 595

stevenrbrandt:
This small change fixes bug #595, allowing tuples to be converted to lists on returning from a function.

### DIFF
--- a/python/phylanx/ast/physl.py
+++ b/python/phylanx/ast/physl.py
@@ -706,7 +706,7 @@ class PhySL:
         """class Tuple(elts, ctx)"""
 
         expr = tuple(map(self.apply_rule, node.elts))
-        return expr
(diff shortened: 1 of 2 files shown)


python/phylanx/ast/physl.py
```diff
@@ -632,7 +632,10 @@ def _Return(self, node):
             end of the function!
         """
 
-        return self.apply_rule(node.value)
+        if type(node.value) == ast.Tuple:
+            return ["make_list", self.apply_rule(node.value)]
```
hkaiser: I think we're trying to move away from `make_list()` in favor of just `list()`. Would you mind changing this?

stevenrbrandt:
@hkaiser I made the change you requested

Issue #614, opened 2018-10-09
Support dictionaries in Python


stevenrbrandt:
One should be able to create and slice the dictionary type in Python, e.g.
```
@Phylanx
def dc2():
    a = {1:2, 3:4}
    a[1]=5
    return a
```

hkaiser:
This was fixed by merging #615

PR #615, opened 2018-10-09
Adds support for using dictionaries in Python

stevenrbrandt:
Fixes ticket #614 and allows users to create and slice dictionaries in Python.

### DIFF
--- a/src/execution_tree/primitives/slice.cpp
+++ b/src/execution_tree/primitives/slice.cpp
@@ -141,7 +141,7 @@ namespace phylanx { namespace execution_tree
         } 
         if (is_dictionary_operand(data))
         {
-            auto f = phylanx::execution_tree::extract_dictionary_value(data);
+            auto&& f = phylanx::execution_tree::extract_dictionary_value(data);
             f[indices] = value;
             return primitive_argument_type{f};
         }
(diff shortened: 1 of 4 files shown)


rtohid:
Looks good to me, thanks.

src/execution_tree/primitives/slice.cpp
```diff
@@ -133,12 +138,18 @@ namespace phylanx { namespace execution_tree
                         std::move(value), name, codename),
                     name, codename)};
             }
+        } 
+        if (is_dictionary_operand(data))
+        {
+            auto f = phylanx::execution_tree::extract_dictionary_value(data);
+            f[indices] = value;
+            return primitive_argument_type{f};
```
hkaiser: I would have expected for similar code to be necessary in the variable primitive. How is
```
@Phylanx
def dc2():
    a = {1:2, 3:4}
    a[1] = 5
    return a
```
translated to PhySL? For best possible performance I'd like for `a[1] = 5` to become `store(slice(a, 1, nil), 5)`. The PhySL compiler will transform this into a (extended) call to `variable::store` (see: https://github.com/STEllAR-GROUP/phylanx/blob/master/src/execution_tree/primitives/store_operation.cpp#L176).

stevenrbrandt: It gets translated as "store(slice(a,1), 5)". There really isn't anything that can be done about that, since we can't have the translation depend on the variable type.

hkaiser: That's fine.

PR #616, opened 2018-10-09
Advanced indexing

### LINKED ISSUE #624: Branch 'advanced_indexing': Unexpected behavior of slicing operation.
Slicing with an implicit upper bound, applied to the second dimension, does not operate correctly on 2D arrays. It fails if the size of the first dimension is greater than the second's:
```py
import numpy as np
from phylanx import Phylanx

@Phylanx
def foo():
    x = np.zeros((4, 3))
    return x[0, :]

assert (np.array([0, 0, 0]) == foo()).all
```
with the following error:
```
 variable$x:: cannot extract the requested matrix elements: HPX(bad_parameter)
```
and produces incorrect results if the second dimension is larger:
```py
import numpy as np
from phylanx import Phylanx

@Phylanx
def foo():
    x = np.zeros((3, 4))
    return x[0, :]

print(foo())
assert (np.array([0, 0, 0, 0]) == foo()).all
```
output:
```
[0. 0. 0.]
  File "C:\Users\rod\src\vs\phyton\test.py", line 172, in <module>
    assert (np.array([0,0,0,0]) == foo()).all
AttributeError: 'bool' object has no attribute 'all'
```
(It performs as expected in the case of square matrices.)

hkaiser:
This change set adds advanced integer indexing and advanced Boolean indexing for 1d arrays.

This also fixes #624.


scheherzade:
Everything works fine for me, thanks.

PR #621, opened 2018-10-11
Force CircleCI to build Phylanx with C++14

parsa:
This PR forces CircleCI builders to use C++14 for building Phylanx.

Phylanx is supposed to support at least C++14. Previous builds were all tested with C++17 mode enabled.


hkaiser:
I think we can close this unmerged.

PR #623, opened 2018-10-12
Adding the infrastructure to support dtype specification

### LINKED ISSUE #635: 'arange' with one or two arguments.
If the number of arguments to 'arange' is less than 3, e.g.:
```py
from phylanx import Phylanx
import numpy as np

@Phylanx
def test_arange_one_args():
    a = np.arange(3)
    return a

assert(test_arange_one_args() == np.arange(3)).all()


@Phylanx
def test_arange_two_args():
    a = np.arange(1, 3)
    return a

assert(test_arange_two_args() == np.arange(1, 3)).all()
```

Phylanx fails with:
```
RuntimeError: .../test_dtype.py(11, 8): call-function$arange:: the arange primitive requires two or three arguments.: HPX(bad_parameter)
```

### LINKED ISSUE #636: Boolean NumPy arrays.
NumPy arrays with boolean elements work as expected in Phylanx kernels. However, if returned from a function, their boolean values are replaced with 0s, and 1s, e.g.:

```py
from phylanx import Phylanx
import numpy as np

@Phylanx
def bool_array():
    a = np.array([True, False])
    print(a) # prints '[true, false]' as expected.
    return a

print(bool_array()) # prints '[1 0]'.
```
```
[true, false]
[1 0]
```

hkaiser:
- adding dtype support to `arange`, `cumsum`, `hstack`, `constant`, `identity`, and `vstack`

This automatically exposes primitives with `__bool`, `__int`, and `__float` appended (e.g. `arange__int`, `arange__float`, `arange__bool`).

This also:
- fixes #635
- fixes #636


hkaiser:
@rtohid could you please add support for mapping any `dtype=...` named argument to the new primitive names?

python/phylanx/ast/physl.py
```diff
@@ -309,28 +310,66 @@ def _Call(self, node):
...
+                return [symbol, ('1', [op, args])]
+            else:
+                return [symbol, ('1', args)]
+        elif 'identity' in symbol:
+            symbol = symbol.replace('identity', 'identity' + dtype)
+        elif 'arange' in symbol:
+            symbol = symbol.replace('arange', 'arange' + dtype)
+        elif 'cumsum' in symbol:
+            symbol = symbol.replace('cumsum', 'cumsum' + dtype)
         # +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
```
hkaiser: Could we add `numpy.full` as well. Also, I believe we can add `ones_like`, `zero_like`, and `full_like`.

PR #626, opened 2018-10-17
Updating simple als test

scheherzade:
I changed the test so that it would check the physl and cpp versions with each other using random matrices, but checks the results of physl code using fixed input matrices with the expected results from python with the same input matrices.


hkaiser:
The regression test failure is unrelated and will be fixed separately

PR #627, opened 2018-10-17
Expose performance information (CSV, DOT) to Python

hkaiser:
This adds a new option to the @PhySL decorator (`performance=…`, the value of this decorator argument is ignored, currently). If this is set, then a decorated function object `func` will expose `func.__perfdata__`, which is a tuple containing the CSV data and the DOT tree structure (both strings).

### DIFF
--- a/python/phylanx/ast/physl.py
+++ b/python/phylanx/ast/physl.py
@@ -103,6 +103,9 @@ class PhySL:
                 self.numpy_aliases.add(key)
         self.file_name = self.fglobals['__file__']
 
+        self.performance = not (kwargs.get('performance') is None)
+        self.__perfdata__ = (None, None)
+
         # Add arguments of the function to the list of discovered variables.
         if inspect.isfunction(tree.body[0]):
             for arg in tree.body[0].args.args:
@@ -123,7 +126,8 @@ class PhySL:
         elif PhySL.compiler_state is None:
             PhySL.compiler_state = compiler_state()
 
-        phylanx.execution_tree.compile(self.file_name, self.__src__, PhySL.compiler_state)
+        phylanx.execution_tree.compile(
+            self.file_name, self.__src__, PhySL.compiler_state)
 
     def generate_physl(self, ir):
         if len(ir) == 2 and isinstance(ir[0], str) and isinstance(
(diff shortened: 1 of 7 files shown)


python/phylanx/ast/physl.py
```diff
@@ -103,6 +103,9 @@ def __init__(self, func, tree, kwargs):
                 self.numpy_aliases.add(key)
         self.file_name = self.fglobals['__file__']
 
+        self.performance = not (kwargs.get('performance') is None)
```
rtohid: This should also work since None and False behave the same as a flag:
```suggestion
        self.performance = kwargs.get('performance')
```

hkaiser: I did that before, but then this didn't work as I expected: https://github.com/STEllAR-GROUP/phylanx/pull/627/files#diff-2d3509625b4e7c1f8fb2ffdf8f3e53d2R171. Not sure why.

hkaiser:
> Please include the Newick tree structure 

Will do. No problem!

alex-r-bigelow:
I'm poking at this a little bit to see whether there might be a good way to integrate accessing `func.__perfdata__` directly with a Flask server for hosting visualizations. While it runs fine without any arguments in the decorator, I'm hitting an odd error when I try to ask for performance data (I'm blindly [replicating](https://github.com/alex-r-bigelow/phylanx-test) a lot of the stuff from the Jupyter notebook docker build):

```
In [5]: @Phylanx(performance='x') 
   ...: def fib(n): 
   ...:     if n < 2: 
   ...:         return n 
   ...:     else: 
   ...:         return fib(n-1)+fib(n-2) 
   ...:                                                                                                                                

In [6]: fib(10)                                                                                                                        
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
<ipython-input-6-4f8ac2e26ef8> in <module>
----> 1 fib(10)

~/.local/lib/python3.6/site-packages/phylanx-0.0.1-py3.6-linux-x86_64.egg/phylanx/ast/transducer.py in __call__(self, *args)
     80                     "OpenSCoP kernels are not yet callable.")
     81 
---> 82             result = self.backend.call(args)
     83 
     84             self.__perfdata__ = self.backend.__perfdata__

~/.local/lib/python3.6/site-packages/phylanx-0.0.1-py3.6-linux-x86_64.egg/phylanx/ast/physl.py in call(self, args)
    367 
    368         result = phylanx.execution_tree.eval(
--> 369             self.file_name, func_name, PhySL.compiler_state, *args)
    370 
    371         if self.performance:

RuntimeError: invalid reference id: HPX(bad_parameter)

```

Any clues what I might be doing wrong?

hkaiser:
Looks like a bug to me - will investigate.

PR #628, opened 2018-10-19
Preconditioner

diehlpk:
Add iterative solvers from BlazeIterative
- Conjugate Gradient
- Biconjugate gradient stabilized method

For the Biconjugate gradient stabilized method following preconditioners are available

- Cholesky
- LU
- RQ
- QR

### DIFF
--- a/cmake/FindBlazeIterative.cmake
+++ b/cmake/FindBlazeIterative.cmake
@@ -1,0 +1,17 @@
+# Copyright (c) 2018 Patrick Diehl
+#
+# Distributed under the Boost Software License, Version 1.0. (See accompanying
+# file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+find_package(PkgConfig)
+
+find_path(BLAZEITERATIVE_INCLUDE BlazeIterative/BlazeIterative.hpp HINTS /usr/include /usr/local/include  "${BLAZEITERATIVE_DIR}/include/" )
+
+mark_as_advanced(BLAZEITERATIVE_DIR)
+mark_as_advanced(BLAZEITERATIVE_INCLUDE)
+
+if(NOT BLAZEITERATIVE_INCLUDE)
+message(FATAL_ERROR "Blazeiterative library not found: Specify the BLAZEITERATIVE_DIR") 
+else()
+include_directories(${BLAZEITERATIVE_INCLUDE}/BlazeIterative)
+add_definitions(-DPHYLANX_HAS_BLAZEITERATIVE)
+endif()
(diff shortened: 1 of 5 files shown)


diehlpk:
@parsa Could you please rebuild the docker file using the latest master version of BlazeIterative?

This should solve the compilation errors.

cmake/FindBlazeIterative.cmake
```diff
@@ -0,0 +1,17 @@
+# Copyright (c) 2018 Patrick Diehl
+#
+# Distributed under the Boost Software License, Version 1.0. (See accompanying
+# file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+find_package(PkgConfig)
```
hkaiser: This will not work on non-Linux platforms that don't have pkgconfig available.

src/plugins/solvers/linear_solver.cpp
```diff
@@ -24,6 +25,11 @@
 
 #include <blaze/Math.h>
 
+#ifdef PHYLANX_HAS_BLAZEITERATIVE
+#include <BlazeIterative.hpp>
```
hkaiser: Shouldn't this be `#include <BlazeIterative/BlazeIterative.hpp>` instead?

src/plugins/solvers/linear_solver.cpp
```diff
@@ -74,7 +80,84 @@ namespace phylanx { namespace execution_tree { namespace primitives
...
+         "\n"
+         "    a (matrix) : a matrix\n"
+         "    b (vector) : a vector\n"
+         "\n"
+         "Returns:\n"
+         "\n"
+         "A matrix `x` such that `a x = b`, solved using the "
+         "conjugate gradient"
+        ),
+        PHYLANX_LIN_MATCH_DATA("iterative_solver_BiCGSTAB",
```
hkaiser: Could we please stick to our naming convetions we use in HPX/Phylanx (see: https://github.com/STEllAR-GROUP/hpx/wiki/HPX-Source-Code-Structure-and-Coding-Standards)?

cmake/FindBlazeIterative.cmake
diehlpk: Good comment, I will add support for Windows as well.

diehlpk: https://github.com/STEllAR-GROUP/hpx/blob/master/cmake/FindOrangeFS.cmake#L7
https://github.com/STEllAR-GROUP/hpx/blob/master/cmake/FindJemalloc.cmake#L10

Ok, for HPX we do it similar and I assume that the QUIET option should help to get it run on Windows or?

PR #645, opened 2018-10-30
Fix #639.

### LINKED ISSUE #639: Add full, zeros_like, ones_like, and full_like
These functions should be added to the Python frontend mapping things to various invocations of the `constant` primitive.

rtohid:
This pull request adds frontend support for the following NumPy methods:
- full
- full_like
- ones_like
- zeros_like

Fixes #639

### DIFF
--- a/python/phylanx/ast/physl.py
+++ b/python/phylanx/ast/physl.py
@@ -375,6 +375,18 @@ class PhySL:
             else:
                 symbol = symbol.replace('hstack', 'hstack' + dtype)
             args = args[0][1]
+        elif 'zeros_like' in symbol:
+            symbol = symbol.replace('zeros_like', 'constant' + dtype)
+            op = get_symbol_info(node.func, 'shape')
+            return [symbol, ('0', [op, args])]
+        elif 'ones_like' in symbol:
+            symbol = symbol.replace('ones_like', 'constant' + dtype)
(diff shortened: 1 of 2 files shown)


python/phylanx/ast/physl.py
```diff
@@ -389,6 +401,13 @@ def _Call(self, node):
                 return [symbol, ('1', [op, args])]
             else:
                 return [symbol, ('1', args)]
+        elif 'full' in symbol:
+            symbol = symbol.replace('full', 'constant' + dtype)
+            op = get_symbol_info(node.func, 'list')
+            if isinstance(args[0], tuple):
+                return [symbol, (args[1], [op, args[0]])]
+            else:
+                return [symbol, (args[1], args[0])]
```
hkaiser: This might not belong into this PR, however `vstack()` should obey any given `dtype` as well.

rtohid: It is already supported, please see [here](https://github.com/STEllAR-GROUP/phylanx/blob/f4e0fb1f46c8c93c511bd69c61c6bcf72cf378d8/python/phylanx/ast/physl.py#L367).

hkaiser: dtype support for vstack will be added by a separate PR.

PR #646, opened 2018-10-30
Change constant() to support ...like and empty operations

### LINKED ISSUE #638: Add ability to create uninitialized array.
This would correspond to `numpy.empy` and `numpy.empty_like`. This could be implemented as an extension to the existing `constant()` primitive.

hkaiser:
This fixes the constant primitive to support `empty` and all `...like` operations.

This fixes #638


hkaiser:
@rtohid this change has implications on #645:

- `empty` (and friends) is now supported
- all `...like` functions need to be mapped onto `constant_like` now.

PR #649, opened 2018-10-31
Add parallel_block capability to Phylanx

stevenrbrandt:
This PR allows one to write code containing parallel blocks. To use it, first import "parallel" from phylanx, then use "with parallel" or "with phylanx.parallel" as appropriate to generate a parallel_block() in physl. See test "parallel.py".

### DIFF
--- a/python/phylanx/__init__.py
+++ b/python/phylanx/__init__.py
@@ -13,5 +13,3 @@ class parallel:
     @staticmethod
     def is_parallel_block():
         return True
-
-
--- a/python/phylanx/__init__.py
+++ b/python/phylanx/__init__.py
@@ -6,10 +6,4 @@
 from .core import *
 from .core.config import *
 from .ast.transducer import Phylanx
-
-
-class parallel:
-
-    @staticmethod
-    def is_parallel_block():
-        return True
+from .execution_tree import parallel
--- a/python/phylanx/__init__.py
+++ b/python/phylanx/__init__.py
@@ -6,3 +6,12 @@
 from .core import *
 from .core.config import *
 from .ast.transducer import Phylanx
+
+
+class parallel:
+
+    @staticmethod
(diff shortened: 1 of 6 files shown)


hkaiser:
@stevenrbrandt please fix the flake errors reported.

python/phylanx/__init__.py
```diff
@@ -6,3 +6,10 @@
 from .core import *
 from .core.config import *
 from .ast.transducer import Phylanx
+
+
+class parallel:
```
rtohid: Please move this to an appropriate module- perhaps under execution_tree subpackage.

python/phylanx/ast/physl.py
```diff
@@ -10,6 +10,7 @@
 import inspect
 import phylanx.execution_tree
 from phylanx import compiler_state
+from phylanx.ast.utils import dump_info
```
rtohid: This is not used, please remove it.

python/phylanx/ast/physl.py
```diff
@@ -772,6 +773,20 @@ def _NestedSubscript(node):
         #         return [op, (value, slice_)]
         return [op, (value, slice_)]
 
+    def _With(self, node):
+        if 0 < len(node.items) and type(node.items[0]) == ast.withitem:
```
rtohid: This should be handled in the `_Withitem` method.

stevenrbrandt: Not sure how that can be done. The information needed for the block is not completely contained under ast.withitem.

python/phylanx/__init__.py
stevenrbrandt: Do you have a suggestion?

stevenrbrandt: It needs to be in package phylanx.

PR #650, opened 2018-10-31
Tree traversal

folshost:
Adding code for allowing traversal of AST with the end goal of extracting matrix operation tiling constraints. This code is going to be shelved after this point until Phylanx supports some distributed operations, where it may be resumed.


hkaiser:
@folshost could you please squash the commits for this PR? Also, please fix the reported formatting issues.

PR #654, opened 2018-11-04
initial import of randomforest

ct-clmsn:
randomforest implementation using numpy and python dictionaries.


ct-clmsn:
flake issues resolved; getting a weird error out on image deployment.

PR #657, opened 2018-11-07
Introduce eval_context (to be used later for local variables etc.)

hkaiser:
This is a first step towards fixing function local variables in PhySL.


stevenrbrandt:
Wow, this is a big change.

PR #662, opened 2018-11-09
Adding benchmarks to check the performance of Blaze

scheherzade:
This PR implements daxpy and dvecdvecadd benchmarks. This way we can check the effect of changing the parameters used by Hpx backend of Blaze in Phylanx.


hkaiser:
@scheherzade could you rebase this PR on master please? This should make the reported errors to go away.

PR #663, opened 2018-11-16
Tree traversal

folshost:
Branch for tree traversal, which includes variable list and type extraction.


folshost:
I have no clue what those code formatting failures are about. There seems to be no useful information in the failures that would tell me how to fix them.

hkaiser:
This should be integrated with the existing frontend code. Could you discuss this with @rtohid, please?

PR #665, opened 2018-11-18
Adding 3D tensor support to node_data (requires BlazeTensor library)

hkaiser:
- flyby: renamed configuration constant for BlazeIterative


hkaiser:
@parsa this is still failing on CircleCI… Would you be able to have a look?

PR #671, opened 2018-11-27
Provide an error on illegal returns

stevenrbrandt:
Detect illegal use of return statements within a Python code and raise an exception. Illegal uses include (1) use of return inside while/for, or at any place prior to the end of a block.

### DIFF
--- a/python/phylanx/ast/physl.py
+++ b/python/phylanx/ast/physl.py
@@ -98,6 +98,70 @@ def remove_line(a):
     return re.sub(r'\$.*', '', a)
 
 
+def is_fun(f, a):
+    return type(a) == list and type(a[0]) == str and re.match(f + r'\b', a[0])
+
+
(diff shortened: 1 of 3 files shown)


rtohid:
@stevenrbrandt Could you please use more descriptive variable names and also add docstrings to functions?

python/phylanx/ast/physl.py
```diff
@@ -98,6 +98,70 @@ def remove_line(a):
...
+
+def check_noreturn(a):
+    if type(a) not in [list, tuple]:
+        return
+    if len(a) == 0:
+        return
+    elif len(a) == 1:
+        check_noreturn(a[0])
+    elif is_fun('return', a):
+        raise NotImplementedError("Illegal return")
```
hkaiser: Could you add line information to the exception message, please?

PR #675, opened 2018-11-28
Dynamic configuration of HPX.

rtohid:
Enable users to dynamically set the number of HPX threads in Python.


tests/regressions/python/466_compiler_problems.py
```diff
@@ -6,7 +6,9 @@
 # #466: ast compiler hangs and gives the wrong answer
 
 import numpy as np
-from phylanx import Phylanx
+from phylanx import Phylanx, PhylanxSession
+
+PhylanxSession(1)
```
hkaiser: Could we make it so that this does not have to be called, in which case some defaults are being used.

hkaiser:
Would it be possible to deferred initialize HPX to allow for setting those HPX parameters at runtime?

stevenrbrandt:
Maybe I'm missing something. Is num_threads the only hpx startup parameter that can be set? Why have that limitation?

stevenrbrandt:
Can we provide a test for this code?

hkaiser:
> @hkaiser Do you mean to defer the compilation until the function is called (evaluation stage)? I can make the compilation during the transformation stage optional. So, if HPX is already initialized, compile at the end of transformation stage, otherwise, right before evaluation. What do you think?

That might be a solution, yes. Good thinking!

stevenrbrandt:
I'd like to see the following test added.
```
from phylanx.exceptions import RuntimeAlreadyInitializedError

try:
    PhylanxSession.init(1)
    PhylanxSession.init(1)
    raise Exception()
except RuntimeAlreadyInitializedError:
    pass
```

hkaiser:
@rtohid could you fix the formatting issues, please?

hkaiser:
@rtohid could you please fix the failing tests as well? Also, please rebase onto master to resolve the conflicts.

PR #677, opened 2018-11-30
randomforest algorithm numpy and cpp primitive

ct-clmsn:
work in progress - pushed into phylanx upstream/master for review and commentary.

### DIFF
--- a/phylanx/plugins/algorithms/impl/randomforest.hpp
+++ b/phylanx/plugins/algorithms/impl/randomforest.hpp
@@ -296,55 +296,61 @@ struct randomforest_impl {
                 , [&joint, &jsz](auto rval) { joint[jsz] = rval; ++jsz; });
 
             auto term = to_terminal(train_labels, joint, classes);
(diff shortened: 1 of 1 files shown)


ct-clmsn:
will need to add support for the following capabilities

- [x] implement training/prediction of randomforest
- [ ] implement model serialization/de-serialization to file string path

hkaiser:
> where should cpp unit test code go?

@ct-clmsn what about adding it here: https://github.com/STEllAR-GROUP/phylanx/tree/master/tests/unit/algorithm?

ct-clmsn:
will try to modify the implementation to use primitive types.

examples/algorithms/randomforest/randomforest.py
```diff
@@ -0,0 +1,130 @@
+from numpy.random import rand
```
hkaiser: Would  you mind adding the Boost license information (and your copyright) to this file?

ct-clmsn: sure thing!

PR #679, opened 2018-11-30
Vsplit

folshost:
Addition of vsplit primitive. Imitates functionality of numpy method of the same name (https://docs.scipy.org/doc/numpy/reference/generated/numpy.vsplit.html).


hkaiser:
@folshost Please don't close PRs if you still intend to work on it. You can just keep pushing to the underlying branch to add/modify the initial proposed changes.

hkaiser:
@folshost could you resolve the conflict, please? Otherwise we'll not be able to merge your work.

src/plugins/matrixops/vsplit_operation.cpp
```diff
@@ -0,0 +1,291 @@
...
+        std::vector<std::pair<double, double>> ranges;
+
+        if (block_dims < 1)    // Scalar tiling
+        {
+            phylanx::ir::node_data<double> val =
+                extract_numeric_value(args[1], name_, codename_);
+
+            std::int64_t num_blocks = val.scalar();
+            //    extract_numeric_value(args[1], name_, codename_).scalar();
+            if (num_blocks <= 0)
```
hkaiser: FWIW, you can use `extract_scalar_numeric_value(…)` to ensure the target is actually a scalar (will throw otherwise).

PR #681, opened 2018-12-04
Eye primitive

taless474:
Eye return an N x M matrix with ones on the k-th diagonal and zeros elsewhere as numpy.eye does.

### DIFF
--- a/src/plugins/matrixops/eye_operation.cpp
+++ b/src/plugins/matrixops/eye_operation.cpp
@@ -1,9 +1,9 @@
-//   Copyright (c) 2018 Bita Hasheminezhad
-//   Copyright (c) 2018 Shahrzad Shirzad
-//   Copyright (c) 2018 Hartmut kaiser
+// Copyright (c) 2018 Bita Hasheminezhad
+// Copyright (c) 2018 Shahrzad Shirzad
+// Copyright (c) 2018 Hartmut kaiser
 //
-//   Distributed under the Boost Software License, Version 1.0. (See accompanying
-//   file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+// Distributed under the Boost Software License, Version 1.0. (See accompanying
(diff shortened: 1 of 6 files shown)


src/plugins/matrixops/eye_operation.cpp
```diff
@@ -0,0 +1,243 @@
...
+            vecsize = N + k;
+            if (M < vecsize)
+                vecsize = M;
+        }
+
+        if (vecsize > 0)
+        {
+            blaze::DynamicVector<T> ones(vecsize, T(1));
+            blaze::Band<blaze::DynamicMatrix<T>> eye = blaze::band(result, k);
+            eye = ones;
```
hkaiser: This would be better written as `blaze::band(result, k) = ones;`, as it would save the additional allocation for the `eye` matrix. Possibly you could even write it as: `blaze::band(result, k) = T(1);`, which would avoid allocating the vector as well.

src/plugins/matrixops/eye_operation.cpp
```diff
@@ -0,0 +1,232 @@
...
+
+        auto vecsize = static_cast<T>(1);    // a positive number (k=0)
+
+        if (k > 0)
+            vecsize = M - k;
+        else if (k < 0)
+            vecsize = N + k;
+
+        if (vecsize > 0)
+            blaze::band(result, k) = T(1);
```
hkaiser: If `k == 0` this should return the identity matrix. AFAICS, currently the code returns a matrix with all zeros.

taless474: I think when `k == 0` we have `vecsize == 1` and since `vecsize` is positive we still have `blaze::band(result, k) = T(1)`
Please tell me if I am missing something.

hkaiser: You're right, I misunderstood the code.

hkaiser:
Could you please rebate this PR on master to resolve the conflicts?

PR #683, opened 2018-12-05
Fixing minor issues with extract_shape and generic operations primitives

scheherzade:
Shape should return a list of the vector size and nil for vectors:
```np.shape([1.0, 2.0]) = (2,)``` while ```shape([1.0, 2.0]) returns (2)``` in phylanx.


hkaiser:
Hmmm, I'm not sure this is right. In Python `len(np.array([1,2]).shape) == 1`, while now in Phylanx `len(shape([1, 2]) == 2`. Phylanx was already returning a list of length one.

PR #685, opened 2018-12-07
Adding flatten primitive

scheherzade:
The flatten primitive returns a copy of the array collapsed in one dimension.
https://docs.scipy.org/doc/numpy-1.14.0/reference/generated/numpy.chararray.flatten.html

### DIFF
--- a/src/plugins/matrixops/flatten.cpp
+++ b/src/plugins/matrixops/flatten.cpp
@@ -1,0 +1,304 @@
+// Copyright (c) 2018 Shahrzad Shirzad
+// Copyright (c) 2018 Hartmut Kaiser
+//
+// Distributed under the Boost Software License, Version 1.0. (See accompanying
+// file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+#include <phylanx/config.hpp>
+#include <phylanx/execution_tree/primitives/node_data_helpers.hpp>
+#include <phylanx/ir/node_data.hpp>
+#include <phylanx/plugins/matrixops/flatten.hpp>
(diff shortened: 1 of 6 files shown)


src/plugins/matrixops/flatten.cpp
```diff
@@ -0,0 +1,313 @@
...
+                        switch (a_dims)
+                        {
+                        case 0:
+                            return this_->flatten0d(std::move(arg));
+
+                        case 1:
+                            return this_->flatten1d(std::move(arg));
+
+                        case 2:
+                            return this_->flatten2d(std::move(arg), order);
```
hkaiser: Minor optimization: you might want to move the `order` argument. No need to create another copy. Alternatively, just pass the single character `'C'` or `'F'` to `flatted2d`.

src/plugins/matrixops/flatten.cpp
```diff
@@ -0,0 +1,313 @@
...
+
+        auto this_ = this->shared_from_this();
+        if (operands.size() == 1)
+        {
+            return hpx::dataflow(hpx::launch::sync,
+                hpx::util::unwrapping(
+                    [this_ = std::move(this_)](primitive_argument_type&& arg)
+                        -> primitive_argument_type {
+                        std::string order = "C";
+                        if (order != "C" && order != "F")
```
hkaiser: `order` will always be equal to `'C'`, is this an oversight, will there be more added later?

scheherzade: When order is not specified it should be 'C', I copied from the part when operands.size() !=1 but forgot to change this part sorry.

PR #688, opened 2018-12-10
Adding Softmax operation

taless474:
Softmax returns the normalized exponential function of the given array. Since blaze has this operation, I think we can have softmax primitive, too.
It will be updated when issue #218 in blaze (https://bitbucket.org/blaze-lib/blaze/issues/218/provide-a-softmax-overload-with-support) is resolved.

### DIFF
--- a/src/plugins/matrixops/softmax_operation.cpp
+++ b/src/plugins/matrixops/softmax_operation.cpp
@@ -74,7 +74,7 @@ namespace phylanx { namespace execution_tree { namespace primitives
         blaze::DynamicMatrix<double> result(m.rows(), m.columns());
         for (std::size_t i = 0; i < m.columns(); ++i)
         {
-            column(result, i) = softmax(column(m, i));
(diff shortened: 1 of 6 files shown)


src/plugins/matrixops/softmax_operation.cpp
```diff
@@ -0,0 +1,188 @@
...
+    }
+
+    primitive_argument_type softmax_operation::softmax2d_axis0(
+        arg_type&& arg) const
+    {
+        auto m = arg.matrix();
+        blaze::DynamicMatrix<double> result(m.rows(), m.columns());
+        for (std::size_t i = 0; i < m.columns(); ++i)
+        {
+            column(result, i) = softmax(column(m, i));
```
hkaiser: While `column` and `softmax` are being found via ADL, I'd suggest to explicitly decorate those as `blaze::column` and `blaze::softmax`.

Issue #689, opened 2018-12-12
check return does not work with a nested function


stevenrbrandt:
The following should not cause an error:
```
@Phylanx
def recur12():
    def add(a,b):
        return a+b 
    return add(3,2)
```

hkaiser:
This should have been fixed by merging #690

PR #690, opened 2018-12-12
Fixes the problem with returns from inner functions.

stevenrbrandt:
Allows this function to pass checks.
```
@Phylanx
def recur12():
    def add(a,b):
        return a+b 
    return add(3,2)
```


hkaiser:
This fixes #689

stevenrbrandt:
Ping @hkaiser @rtohid

PR #691, opened 2018-12-12
Adding Reshape primitive

taless474:
Reshape primitive has the functionality of `numpy.reshape` except for the order. It returns a new shape to an array without changing its data.


src/plugins/matrixops/reshape_operation.cpp
```diff
@@ -0,0 +1,406 @@
...
+                     compatible with the original shape (number of elements in
+                     both arrays are the same). If an integer, then the result
+                     will be a 1-D array of that length. The last parameter of
+                     the newshape can be -1. In this case, the value is inferred
+                     from the length of the array and remaining dimensions.
+
+            Returns:
+
+            Returns a new shape to an array without changing its data."
+            )") };
```
hkaiser: In order for the `...__float` names etc. to be recognized by the compiler, the `match_data` needs to be passed a `true` as the second argument during construction.

taless474: Thank you for your comment, I will correct this in the `eye` primitive. NumPy [reshape](https://docs.scipy.org/doc/numpy-1.15.1/reference/generated/numpy.reshape.html) does not have a `dtype`, so here, I only tried to extract the given type and produce the same thing in the new shape.

hkaiser: I'm referring to https://github.com/STEllAR-GROUP/phylanx/pull/691/files#diff-be4e16b025144db3a166da2121df2dcaR58 where the current code tries to extract the `dtype` for the operation from the primitive's name.

PR #692, opened 2018-12-12
Adding prod primitive

scheherzade:
The prod primitive returns the product of array elements over a given axis.
https://docs.scipy.org/doc/numpy/reference/generated/numpy.prod.html


hkaiser:
Should we add this to the arithmetics plugin instead?

hkaiser:
@scheherzade yah, let's move sum over there too ;-)

PR #696, opened 2018-12-17
Eye Primitive: modified `match_data` to accept dtype

taless474:
In the previous version, dtype support was not enabled.


src/plugins/matrixops/eye_operation.cpp
```diff
@@ -44,7 +44,10 @@ namespace phylanx { namespace execution_tree { namespace primitives {
             "Returns:\n"
             "\n"
             "Return an N x M matrix with ones on the k-th diagonal and zeros\n"
-            "elsewhere")};
+            "elsewhere)",
+            true
```
hkaiser: If you use a raw string (`R"(…)"`) you don't have to use quotes and '\n' inside the actual string.

PR #697, opened 2018-12-19
Adding Max primitive

taless474:
Although there is an `amax` in generic_operations, having `axis` and `keepdims` options come handy in many cases. Therefore, max_primitive has a part of the functionality of [NumPy amax](https://docs.scipy.org/doc/numpy-1.15.0/reference/generated/numpy.amax.html) 
`max(a,axis,keepdims)`

### DIFF
--- a/phylanx/plugins/matrixops/max_operation.hpp
+++ b/phylanx/plugins/matrixops/max_operation.hpp
@@ -22,15 +22,13 @@
 #include <vector>
 
 namespace phylanx { namespace execution_tree { namespace primitives {
-    /// \brief
+    /// \brief Calculates the maximum of an array or maximum along an axis.
     /// \param a         The scalar, vector, or matrix to perform max over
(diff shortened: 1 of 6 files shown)


phylanx/plugins/matrixops/max_operation.hpp
```diff
@@ -0,0 +1,88 @@
...
+            ir::node_data<T>&& arg, bool keep_dims) const;
+        template <typename T>
+        primitive_argument_type max2d_axis1(
+            ir::node_data<T>&& arg, bool keep_dims) const;
+        template <typename T>
+        primitive_argument_type maxnd(ir::node_data<T>&& arg,
+            hpx::util::optional<std::int64_t> axis, bool keep_dims) const;
+
+    private:
+        node_data_type dtype_;
```
hkaiser: This shouldn't be needed.

hkaiser:
I took the liberty to move this primitive into the new statistics plugin (see #706).

PR #701, opened 2018-12-19
Adding Flip primitive

taless474:
Flip primitive imitates the functionality of [NumPy flip](https://docs.scipy.org/doc/numpy-1.14.0/reference/generated/numpy.flip.html).

### DIFF
--- a/phylanx/util/detail/bad_swap.hpp
+++ b/phylanx/util/detail/bad_swap.hpp
@@ -1,0 +1,83 @@
+// Copyright (c) 2018 Parsa Amini
+// Copyright (c) 2018 Bita Hasheminezhad
+// Copyright (c) 2018 Hartmut Kaiser
+//
+// Distributed under the Boost Software License, Version 1.0. (See accompanying
+// file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+#include <phylanx/config.hpp>
+
+#include <algorithm>
+
+#include <blaze/Math.h>
+
+//////////////////////////////////////////////////////////////////////////
+namespace blaze
+{
+    namespace _row_swap
+    {
+        using std::swap;
+
+        template <typename T>
+        void check_swap() noexcept(
+            noexcept(swap(std::declval<typename T::BaseType&>(),
+                std::declval<typename T::BaseType&>())))
+        {}
+    }
+
+    // BADBAD: This overload of swap is necessary to work around the problems
+    //         caused by matrix_row_iterator not being a real random access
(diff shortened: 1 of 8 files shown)


src/plugins/matrixops/flip_operation.cpp
```diff
@@ -0,0 +1,219 @@
...
+                "flip_operation::flip1d",
+                generate_error_message(
+                    "the flip_operation primitive requires operand axis to be "
+                    "either 0 or -1 for vectors."));
+        }
+
+        auto v = arg.vector();
+        blaze::DynamicVector<T> result(v.size());
+        std::reverse_copy(std::begin(v), std::end(v), std::begin(result));
+
```
hkaiser: If `arg` is not a reference you should be able to avoid allocating the result. In this case the operation could be performed in-place. I.e. something like:
```
if (!arg.is_ref())
{
    auto v = arg.vector(); 
    std::reverse(std::begin(v), std::end(v));
    return primitive_argument_type{std::move(arg)};
}
else
{
    auto v = arg.vector(); 
    blaze::DynamicVector<T> result(v.size()); 
    std::reverse_copy(std::begin(v), std::end(v), std::begin(result));
    return primitive_argument_type{std::move(result)};
}
```
should do the trick.

src/plugins/matrixops/flip_operation.cpp
```diff
@@ -0,0 +1,219 @@
...
+        }
+    }
+
+    template <typename T>
+    primitive_argument_type flip_operation::flip2d_axis0(
+        ir::node_data<T>&& arg) const
+    {
+        using phylanx::util::matrix_row_iterator;
+
+        auto m = arg.matrix();
```
hkaiser: Same here, if `arg` is not a reference the allocation of the result can be avoided.

hkaiser:
Please resolve the conflicts for this PR to be merged.

phylanx/util/detail/bad_swap.hpp
```diff
@@ -0,0 +1,87 @@
...
+        noexcept(*std::declval<typename T::Iterator>()))
+    {
+        for (auto a = x.begin(), b = y.begin(); a != x.end() && b != y.end();
+            ++a, ++b)
+        {
+            iter_swap(a, b);
+        }
+    }
+}
+#endif
```
hkaiser: Thanks for extracting this code into a header file. Could you now use this header file in the squeeze primitive as well, please?

hkaiser: I meant the `shuffle` primitive, sorry.

PR #714, opened 2018-12-29
Attempting to improve CircleCI workflow

sithhell:
This steps fixes the issue that the testing stage is rebuilding parts of
the library.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -68,7 +68,7 @@ anchors:
             when: always
             command: |
                 mkdir -p ${CIRCLE_JOB}
-                xsltproc conv.xsl Testing/$(head -n 1 < Testing/TAG)/Test.xml >${CIRCLE_JOB}/Test.xml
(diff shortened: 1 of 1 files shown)


hkaiser:
@sithhell any idea why the deploy step started to fail again after the merge from master?

sithhell:
@hkaiser no idea. looks like the latest merge has been successful... I improved it a bit further to only have the install step run on one core only.

parsa:
@sithhell CircleCI still exhibits the same issue as before. CMake still occasionally re-triggers building parts of the library it is not supposed to. [This build](https://circleci.com/workflow-run/b43fce69-7878-491a-ab5e-6262c6fbb86c) failed, but [re-triggering it](https://circleci.com/workflow-run/c442262e-b44f-4ab0-ab02-775875d94f5b) passed. This is the problem we had before as well.

sithhell:
Ok, there's something in cmake causing this. I have no idea what. One thing I noticed was that I had to persist the source directory as well as something in the cmake script writes to the source directory

hkaiser:
@sithhell does it make sense to merge this now? What does this PR change comparing to master?

PR #718, opened 2019-01-03
Python decorator wrapper modifications

Yeahhhh:
1. Better kwargs handeling: Default kwargs, key validations, value validations
2. Dropping "util.py:dump_info" in favor of astpretty library, which can print to string besides printing to stdout
3. Optionally dumping intermediate codes to files.

Thanks!
Ye Fang <yfang11 AT lsu.edu>


python/phylanx/ast/oscop.py
```diff
@@ -40,6 +39,11 @@ def __init__(self, root, kwargs):
         self.__src__ = self.template.substitute(
             self.oscop_global) + '\n' + self.__src__
 
+        if kwargs["dump_openscop"] == True:
```
hkaiser: It might be useful to allow for the user to specify a file name where to dump the data to. the named argument could be used for that.

python/phylanx/ast/physl.py
```diff
@@ -227,6 +228,11 @@ def __init__(self, func, tree, kwargs):
         check_return(self.ir)
         self.__src__ = self.generate_physl(self.ir)
 
+        if kwargs["dump_physl"] == True:
+            ofn = "./dump_pyhsl_" + kwargs["python_src_tag"] + ".txt"
```
hkaiser: Same here: let's use the named argument to determine where to dump the information to.

python/phylanx/ast/physl.py
```diff
@@ -227,6 +228,11 @@ def __init__(self, func, tree, kwargs):
...
+            dump_to_file(self.__src__, ofn, kwargs["verbos"])
+
```
hkaiser: This might be a typo. Shouldn't that be `"verbose"` instead?

python/phylanx/ast/transducer.py
```diff
@@ -16,113 +16,80 @@
 from phylanx import execution_tree
 from phylanx.ast import generate_ast as generate_phylanx_ast
 from phylanx.exceptions import InvalidDecoratorArgumentError
-
-from .utils import dump_to_file
-
+from phylanx.ast.utils import dump_to_file
+# from phylanx.ast.utils import python_ast_format
```
hkaiser: Please remove unneeded code.

python/phylanx/ast/utils.py
```diff
@@ -94,27 +96,23 @@ def _format(node, level=0):
...
+    if verbose is True:
         print(ofn, "saved")
 
 
 def dump_to_file(data, key, kwargs):
-    if kwargs[key] != False:
-        if kwargs[key] == True:
+    if kwargs[key] is not False:
+        if kwargs[key] is True:
             # generating file name automatically
```
hkaiser: Isn't `kwargs[key] is not False` implying `kwargs[key] is True`?

Yeahhhh: kwargs["dump_python_src"] may have 3 kinds of values:
True:  Dumping to output file using auto-generated file name
False: Do not dump
String: Dumping to output file using this user provided "string" as the file name

I will document this behavior besides this procedure.

PR #723, opened 2019-01-06
Adding the environment variable `PHYLANX_PLUGINS_PATH` 

### LINKED ISSUE #722: physl plugin support features
Please add the following to the physl interpreter:

1) Add flag: ``--ls-plugins`` which shows:
* the directories(y) checked for plugins
* plugins discovered at runtime/startup; in the absence of code to run, the flag causes a print out of the plugins discovered

2) Add flag: ``--help-primitive=`` which accepts a PhySL function/primitive string name (example: 'eye') as an argument. the command returns:
* the full path and name of the plugin shared object the primitive belongs to (the so or dll file)
* the 'help' string of the primitive
* if the primitive is not found, print '%s not found' ('%s' is the physl function/primitive queried) and the path(s) searched for plugins

3) Add environment variable for finding plugins installed to a non-standard directory:
* ``$PHYSL_PLUGINS_DIRS``
* The envvar is a ':' delimited list of directories to search for plugins at interpreter startup.

hkaiser:
This allows to specify additional search directories for Phylanx plugins.

This environment variable may contain one or more directories that during startup will be searched for Phylanx plugins. The directories should be delimited by `';'` (windows) or `':'` (otherwise).

- note: this depends on https://github.com/STEllAR-GROUP/hpx/pull/3607
- flyby: adding full plugin path to internal configuration registry

This PR is providing the first changes needed to fix #722

### DIFF
--- a/phylanx/execution_tree/primitives/primitive_component_base.hpp
+++ b/phylanx/execution_tree/primitives/primitive_component_base.hpp
@@ -202,8 +202,8 @@ namespace phylanx { namespace execution_tree
         std::vector<hpx::util::tuple<std::string, match_pattern_type>>;
 
     ///////////////////////////////////////////////////////////////////////////
-    PHYLANX_EXPORT void register_pattern(
-        std::string const&, match_pattern_type const& pattern);
+    PHYLANX_EXPORT void register_pattern(std::string const& name,
+        match_pattern_type const& pattern, std::string const& fullpath);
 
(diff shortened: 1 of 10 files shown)


hkaiser:
@ct-clmsn https://github.com/STEllAR-GROUP/hpx/pull/3607 has been merged to HPX master (requires recompile of HPX). Could you check whether the `$PHYLANX_PLUGINS_PATH` environment variable helps with your problems, please?

PR #724, opened 2019-01-07
Making function-local variables non-static

### LINKED ISSUE #364: Multipule Function Calls Produces Incorrect Results
The following code should print 6 (i.e. `[(1 * 2) + (2 * 2)])`. However, it prints 8
```python
import phylanx
from phylanx.ast import *

@Phylanx("PhySL")
def func (a):
    b = 0
    b = a
    b = b * 2
    return b

@Phylanx("PhySL")
def foo (a, b):
    aa = 0
    bb = 0
    aa = a
    bb = b
    return func(aa) + func(bb)

print(foo(1, 2))
```

hkaiser:
This fixes #364.

@stevenrbrandt, @aserio please check whether this solves your problems.


stevenrbrandt:
This works, but functions using local variables are insanely slow (more than 10,000 x).
The code:
```
from phylanx import Phylanx
from time import time


def fib(n):
    if n < 2:
        return n
    else:
        return fib(n-1) + fib(n-2)


@Phylanx
def fib2(n):
    r = 0
    if n < 2:
        r = n
    else:
        f1 = fib2(n - 1)
        f2 = fib2(n - 2)
        r = f1 + f2
    return r


def t(f,j,s):
    t1 = time()
    v = f(j)
    t2 = time()
    print("Time to compute {}({})={} is {}".format(s,j,v,t2-t1))
    return v

for j in range(14,17):
    v1 = t(fib,j,"fib")
    v2 = t(fib2,j,"fib2")
    assert v1 == v2
```
produces this
```
Time to compute fib(14)=377 is 0.0001876354217529297
Time to compute fib2(14)=377 is 5.381385326385498
Time to compute fib(15)=610 is 0.0002987384796142578
Time to compute fib2(15)=610 is 7.295608997344971
Time to compute fib(16)=987 is 0.0004911422729492188
Time to compute fib2(16)=987 is 11.635223150253296
```

hkaiser:
Thanks @stevenrbrandt! Working on the performance of our evaluation engine is still on the list of things todo.

hkaiser:
@stevenrbrandt The main reason why the second version is much slower than the first is that the first version evaluates the two sub-trees of the Fibonacci calculation concurrently. The second version enforces sequential execution of the two sub-trees. I'm sure that running this code on two cores will make the difference even larger.

stevenrbrandt:
@hkaiser I'm running the example single-threaded. This test, however, proves your concurrent thread theory incorrect:
```
@Phylanx
def fib(n):
    if n < 2:
        return n
    else:
        return fib(n-1) + fib(n-2)


@Phylanx
def fib2(n):
    r = 0
    if n < 2:
        r = n
    else:
        f1 = fib2(n - 1) + fib2(n - 2)
        f2 = 0
        r = f1 + f2
    return r
```
Results in
```
Time to compute fib(14)=377 is 3.497098207473755
Time to compute fib2(14)=377 is 6.550185203552246
Time to compute fib(15)=610 is 3.3779537677764893
Time to compute fib2(15)=610 is 7.517587661743164
Time to compute fib(16)=987 is 5.391682147979736
Time to compute fib2(16)=987 is 12.156063795089722
```

stevenrbrandt:
So for the above python codes, running them through physl instead of python makes them go 5x faster. 🤔🤔🤔

hkaiser:
So I have done some investigation and here is the reason for the second version being slower.
Results for first version (`fib(20)`):
```
Elapsed time: 0.455103 [s]
6765
/runtime/count/component@phylanx_primitive_component_type,1,0.545905,[s],23
/runtime/count/action-invocation@phylanx_primitive_store_single_action,1,0.545916,[s],3
/runtime/count/action-invocation@phylanx_primitive_eval_action,1,0.545920,[s],218912
/runtime/count/action-invocation@phylanx_primitive_bind_action,1,0.545924,[s],2
```

Results for second version (`fib2(20)`):
```
Elapsed time: 1.30182 [s]
6765
/runtime/count/component@phylanx_primitive_component_type,1,1.409006,[s],43820
/runtime/count/action-invocation@phylanx_primitive_store_single_action,1,1.409019,[s],43788
/runtime/count/action-invocation@phylanx_primitive_eval_action,1,1.409021,[s],415927
/runtime/count/action-invocation@phylanx_primitive_bind_action,1,1.409024,[s],43783
```

Those performance counters show that the numbers of created primitives are `23` vs, `43820` and the numbers of invocations of an `eval` function for those are `218912` vs. `415927` (amongst other things). I think that explains the difference.

sithhell:
ASAN shows the following: https://gist.github.com/sithhell/29efaddc4bd0f7e2b4cfea21aa0405a2

hkaiser:
@stevenrbrandt your proposed fix for the problem is either too narrow or too general -- or both ;-)

It might be too general in the sense as now all variable definitions of all function call levels are kept alive throughout the whole execution. Variables are supposed to go out of scope once a the function they are defined in exits. I believe that the proposed change prevents this from happening.

The change is too narrow in the sense that if we apply it [here](https://github.com/STEllAR-GROUP/phylanx/blob/fixing_364/src/execution_tree/primitives/primitive_component_base.cpp#L112), it should also be applied at least [here](https://github.com/STEllAR-GROUP/phylanx/blob/fixing_364/src/execution_tree/primitives/primitive_component_base.cpp#L144) (possibly also [here](https://github.com/STEllAR-GROUP/phylanx/blob/fixing_364/src/execution_tree/primitives/primitive_component_base.cpp#L166) and [here](https://github.com/STEllAR-GROUP/phylanx/blob/fixing_364/src/execution_tree/primitives/primitive_component_base.cpp#L168)).

hkaiser:
All tests pass now, @stevenrbrandt I'd assume you won't object to merging this.

stevenrbrandt:
@hkaiser  Of course not!

Issue #727, opened 2019-01-09
Expand_dims arguments are different from NumPy expand_dims


taless474:
Expand_dims primitive is slightly different from [numpy.expand_dims](https://docs.scipy.org/doc/numpy-1.15.1/reference/generated/numpy.expand_dims.html). In NumPy version axis a necessary argument. 
The current version of expand_dims always assumes that the given axis is 0.

PR #728, opened 2019-01-10
Modifying expand dims and tile primitives

taless474:
The first commit addresses issue #727.
The second commit adds the `blaze` decorator where it is forgotten in the current version of the tile primitive.


taless474:
I had written 1 way of testing each axis, so I added the other way of calling it for each case. 
The axis is mandatory, so I cannot add the tests that do not call the axis.

PR #729, opened 2019-01-10
Generic operations by default always return double arrays (for now)

hkaiser:
- generic operations now support the specification of a dtype
- flyby: move algorithm tests into plugin folder

@taless474 this should solve the issues you reported, please verify


taless474:
Thank you, as I understand, generic operations here accept any argument type and always return a double.
This is probably my mistake; I still cannot call them with a dtype in PhySL:
``` console
cout(floor__int(10))
```
```log
physl: exception caught:
physl.txt(74, 6): floor__int:: unknown function requested: floor__int: HPX(bad_parameter)
```
And using the dtype requires me to imput the same dtype:
``` console
cout(floor__int(10.2))
```
``` log
physl: exception caught:
physl.txt(74, 6): floor__int:: primitive_argument_type does not hold an integer value type (type held: 'phylanx::ir::node_data<double>'): HPX(bad_parameter)
```

hkaiser:
@taless474 should be fine now. Thanks for the report.

taless474:
It is, thanks.

PR #731, opened 2019-01-10
Adding Clip primitive

scheherzade:
Given an interval, values outside the interval are clipped to the interval edges. https://docs.scipy.org/doc/numpy-1.14.0/reference/generated/numpy.clip.html

### DIFF
--- a/phylanx/plugins/matrixops/clip.hpp
+++ b/phylanx/plugins/matrixops/clip.hpp
@@ -1,0 +1,82 @@
+// Copyright (c) 2018 Shahrzad Shirzad
+// Copyright (c) 2018 Hartmut Kaiser
+//
+// Distributed under the Boost Software License, Version 1.0. (See accompanying
+// file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+#if !defined(PHYLANX_PRIMITIVES_CLIP)
+#define PHYLANX_PRIMITIVES_CLIP
+
+#include <phylanx/config.hpp>
+#include <phylanx/execution_tree/primitives/base_primitive.hpp>
+#include <phylanx/execution_tree/primitives/primitive_component_base.hpp>
+
+#include <hpx/lcos/future.hpp>
+
+#include <cstddef>
+#include <cstdint>
+#include <memory>
+#include <string>
+#include <utility>
+#include <vector>
+
+namespace phylanx { namespace execution_tree { namespace primitives
+{
+    /// \brief Implementation of clip as a Phylanx primitive.
+    /// Given an interval, values outside the interval are clipped to the interval edges.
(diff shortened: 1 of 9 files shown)


phylanx/util/detail/numeric_limits_min.hpp
```diff
@@ -0,0 +1,27 @@
...
+    {
+        return -(std::numeric_limits<T>::max)();
+    }
+
+    template <>
+    inline std::uint8_t numeric_limits_min<std::uint8_t>()
+    {
+        return 0;
+    }
+}}}
```
hkaiser: I think we can make both functions `constexpr` nowadays as `std::numeric_limits<>::max` is `constexpr` since C++11 (see https://en.cppreference.com/w/cpp/types/numeric_limits/max).

src/plugins/matrixops/clip.cpp
```diff
@@ -0,0 +1,182 @@
...
+    primitive_argument_type clip::clip2d(ir::node_data<T>&& arg,
+        ir::node_data<T>&& min, ir::node_data<T>&& max) const
+    {
+        auto result = (blaze::max)(
+            min.matrix(), (blaze::min)(max.matrix(), arg.matrix()));
+
+        return primitive_argument_type{ir::node_data<T>{result}};
+    }
+
+    template <typename T>
```
hkaiser: It should be almost trivial to add 3D support here, something like:
```
#if defined(PHYLANX_HAVE_BLAZE_TENSOR)
    template <typename T> 
    primitive_argument_type clip::clip3d(ir::node_data<T>&& arg, 
        ir::node_data<T>&& min, ir::node_data<T>&& max) const 
    { 
        auto result = (blaze::max)( 
            min.tensor(), (blaze::min)(max.tensor(), arg.tensor())); 

        return primitive_argument_type{ir::node_data<T>{result}}; 
    } 
#endif
```
should do the trick.

phylanx/plugins/matrixops/clip.hpp
```diff
@@ -65,6 +65,12 @@ namespace phylanx { namespace execution_tree { namespace primitives
         primitive_argument_type clip2d(ir::node_data<T>&& arg,
             ir::node_data<T>&& min, ir::node_data<T>&& max) const;
 
+#if defined(PHYLANX_HAVE_BLAZE_TENSOR)
+        template <typename T>
+        primitive_argument_type clip::clip3d(ir::node_data<T>&& arg,
+            ir::node_data<T>&& min, ir::node_data<T>&& max) const;
```
hkaiser: Please remove the `clip::` qualification, this will not compile otherwise.

src/plugins/matrixops/clip.cpp
hkaiser: You probably want to call this function as well...

PR #734, opened 2019-01-11
Adding logsumexp primitive

taless474:
logsumexp primitive imitates the functionality of [SciPy logsumexp](https://docs.scipy.org/doc/scipy-0.19.1/reference/generated/scipy.misc.logsumexp.html)
There is also an explanation for the 4th argument of statistics `min`, `max`, `prod` and `sum` in their help.

### DIFF
--- a/tests/unit/plugins/statistics/CMakeLists.txt
+++ b/tests/unit/plugins/statistics/CMakeLists.txt
@@ -4,6 +4,7 @@
 # file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
 
 set(tests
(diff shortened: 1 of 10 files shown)


tests/unit/plugins/statistics/CMakeLists.txt
```diff
@@ -4,6 +4,7 @@
 # file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
 
 set(tests
+	logsumexp_operation
```
hkaiser: Please convert the TAB to spaces...

PR #763, opened 2019-01-28
Adding concatenate primitive

scheherzade:


src/plugins/matrixops/concatenate.cpp
```diff
@@ -0,0 +1,835 @@
...
+
+            if (num_dims != 1)
+                HPX_THROW_EXCEPTION(hpx::bad_parameter,
+                    "phylanx::execution_tree::primitives::concatenate::get_"
+                    "tensor_size",
+                    generate_error_message("the concatenate primitive requires "
+                                           "for all input arrays to "
+                                           "have the same dimension"));
+
+            if (num_dims == 0)
```
hkaiser: I think `num_dims == 0` could never be hit (see the condition above).

src/plugins/matrixops/concatenate.cpp
```diff
@@ -0,0 +1,835 @@
...
+            auto&& val = extract_node_data<T>(std::move(arg));
+            std::size_t num_d = val.num_dimensions();
+            if (num_d == 0)
+            {
+                *iter++ = val.scalar();
+            }
+            else
+            {
+                std::copy(val.vector().begin(), val.vector().end(), iter);
+                iter += val.size();
```
hkaiser: `std::copy` returns an iterator that points to the first element after the one assigned last (see https://en.cppreference.com/w/cpp/algorithm/copy). So the above could be simplified as:
```
iter = std::copy(val.vector().begin(), val.vector().end(), iter);
```

hkaiser: Also, you might want to use `std::move` instead of `std::copy` here (https://en.cppreference.com/w/cpp/algorithm/move).

src/plugins/matrixops/concatenate.cpp
```diff
@@ -0,0 +1,835 @@
...
+                            "value as its first operand"));
+                ir::range lhs = extract_list_value_strict(
+                    std::move(args[0]), this_->name_, this_->codename_);
+
+                if (lhs.is_ref())
+                {
+                    ops = lhs.copy();
+                }
+
+                ops = lhs.args();
```
hkaiser: This unconditionally overwrites `ops` which was just assigned above.

src/plugins/matrixops/concatenate.cpp
```diff
@@ -0,0 +1,817 @@
...
+                            "value as its first operand"));
+                ir::range lhs = extract_list_value_strict(
+                    std::move(args[0]), this_->name_, this_->codename_);
+
+                if (lhs.is_ref())
+                {
+                    ops = lhs.copy();
+                }
+                else
+                    ops = lhs.args();
```
hkaiser: Both conditional branches have the same effect, namely that the range is copied.

khuck:
@hkaiser @scheherzade This broke all the buildbot builds:
```
/var/lib/buildbot/slaves/phylanx/x86_64-gcc7-release/build/src/plugins/matrixops/concatenate.cpp: In member function ‘phylanx::execution_tree::primitive_argument_type phylanx::execution_tree::primitives::concatenate::concatenate_flatten_helper(phylanx::execution_tree::primitive_arguments_type&&) const’:
/var/lib/buildbot/slaves/phylanx/x86_64-gcc7-release/build/src/plugins/matrixops/concatenate.cpp:386:20: error: ‘concatenate_flatten3d’ was not declared in this scope
             return concatenate_flatten3d<T>(std::move(args));
                    ^~~~~~~~~~~~~~~~~~~~~
/var/lib/buildbot/slaves/phylanx/x86_64-gcc7-release/build/src/plugins/matrixops/concatenate.cpp:386:20: note: suggested alternative: ‘concatenate_flatten1d’
             return concatenate_flatten3d<T>(std::move(args));
                    ^~~~~~~~~~~~~~~~~~~~~
                    concatenate_flatten1d
/var/lib/buildbot/slaves/phylanx/x86_64-gcc7-release/build/src/plugins/matrixops/concatenate.cpp:386:43: error: expected primary-expression before ‘>’ token
             return concatenate_flatten3d<T>(std::move(args));
                                           ^
```

see http://ktau.nic.uoregon.edu:8020/#/builders/8/builds/410 for example

scheherzade:
@khuck, @hkaiser I had forgotten to add ```#if defined(PHYLANX_HAVE_BLAZE_TENSOR)``` for that 3d case, I added a commit to the existing pull request for concatenate primitive to fix it.

PR #766, opened 2019-02-01
Reformat help strings to use C++ raw strings.

stevenrbrandt:
Reformat help strings to use C++ raw strings.


hkaiser:
@stevenrbrandt could you please have a look what happened to the check_format build step?

stevenrbrandt:
@hkaiser at a workshop in New Orleans, so it's difficult. I think the tests may simply need to be re-run. I'm trying that now.

stevenrbrandt:
@hkaiser there was a format check failure on code I didn't change. I've updated with master and re-pushed. Hopefully, it will all pass now.

PR #769, opened 2019-02-04
Fix is_list_operand in the base_primitive

taless474:
This PR changes the `is_list_operand` which should return `true` where case 7 is selected (an ir::range)


src/execution_tree/primitives/base_primitive.cpp
```diff
@@ -2480,8 +2480,8 @@ namespace phylanx { namespace execution_tree
         case 4: HPX_FALLTHROUGH;    // phylanx::ir::node_data<double>
         case 5: HPX_FALLTHROUGH;    // primitive
         case 6: HPX_FALLTHROUGH;    // std::vector<ast::expression>
-        case 7: HPX_FALLTHROUGH;    // phylanx::ir::range
-        case 8:                     // phylanx::ir::dictionary
+        case 8: HPX_FALLTHROUGH;    // phylanx::ir::dictionary
+        case 7:                     // phylanx::ir::range
             return true;
```
hkaiser: This change does not modify anything. Could you explain what you had in mind?

taless474:
I believe I misread the code.

Issue #770, opened 2019-02-05
Buildbot builds all failing


khuck:
Commits merged on Feb 4 have broken the build:
http://ktau.nic.uoregon.edu:8020/#/builders/8/builds/385

```
[ 25%] Linking CXX shared library ../lib/libhpx_phylanx.so
CMakeFiles/phylanx_component.dir/execution_tree/primitives/slice_node_data.cpp.o: In function `phylanx::ir::node_data<unsigned char> phylanx::execution_tree::slice_assign<unsigned char>(phylanx::ir::node_data<unsigned char>&&, phylanx::execution_tree::primitive_argument_type const&, phylanx::ir::node_data<unsigned char>&&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)':
slice_node_data.cpp:(.text._ZN7phylanx14execution_tree12slice_assignIhEENS_2ir9node_dataIT_EEOS5_RKNS0_23primitive_argument_typeES6_RKNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEESH_[_ZN7phylanx14execution_tree12slice_assignIhEENS_2ir9node_dataIT_EEOS5_RKNS0_23primitive_argument_typeES6_RKNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEESH_]+0xa0): undefined reference to `void phylanx::execution_tree::extract_value_scalar<unsigned char>(phylanx::ir::node_data<unsigned char>::storage0d_type&, phylanx::ir::node_data<unsigned char>&&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)'
CMakeFiles/phylanx_component.dir/execution_tree/primitives/slice_node_data.cpp.o: In function `phylanx::ir::node_data<double> phylanx::execution_tree::slice_assign<double>(phylanx::ir::node_data<double>&&, phylanx::execution_tree::primitive_argument_type const&, phylanx::ir::node_data<double>&&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)':
[...]

Issue #767, opened 2019-02-04
Buildbot crash on simple test


khuck:
Reproduced the transient argument processing error (see http://ktau.nic.uoregon.edu:8020/#/builders/8/builds/383/steps/12/logs/stdio) with the test program:

```/var/lib/buildbot/slaves/phylanx/x86_64-gcc7-release/build/tools/buildbot/build-delphi-x86_64-Linux-gcc/phylanx-Release/bin/generic_operation_test --hpx:threads=1 --hpx:localities=1 --hpx:node=0```

Will crash regardless of whether any arguments are passed to it (I tried eliminating them one at a time).
The output in gdb is (unfortunately a Release build, so not much useful info):

```
[buildbot@delphi phylanx-Release]$ gdb --args /var/lib/buildbot/slaves/phylanx/x86_64-gcc7-release/build/tools/buildbot/build-delphi-x86_64-Linux-gcc/phylanx-Release/bin/generic_operation_test --hpx:threads=1 --hpx:localities=1 --hpx:node=0
GNU gdb (GDB) Red Hat Enterprise Linux 7.6.1-114.el7
Copyright (C) 2013 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.  Type "show copying"
and "show warranty" for details.
This GDB was configured as "x86_64-redhat-linux-gnu".
For bug reporting instructions, please see:
<http://www.gnu.org/software/gdb/bugs/>...
Reading symbols from /var/lib/buildbot/slaves/phylanx/x86_64-gcc7-release/build/tools/buildbot/build-delphi-x86_64-Linux-gcc/phylanx-Release/bin/generic_operation_test...(no debugging symbols found)...done.
(gdb) run
Starting program: /var/lib/buildbot/slaves/phylanx/x86_64-gcc7-release/build/tools/buildbot/build-delphi-x86_64-Linux-gcc/phylanx-Release/bin/generic_operation_test --hpx:threads=1 --hpx:localities=1 --hpx:node=0
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib64/libthread_db.so.1".
[...]

PR #772, opened 2019-02-07
Moving definition from source to header.

khuck:
This fixes issue #770.


khuck:
@hkaiser Ha! CircleCI tests failed because of this:  https://github.com/STEllAR-GROUP/phylanx/issues/767

And I have some new clues on that one...I'll add them in a minute

PR #773, opened 2019-02-07
Updating softmax based on blaze::softmax updates

taless474:
Blaze has provided an [overload](https://bitbucket.org/blaze-lib/blaze/issues/218/provide-a-softmax-overload-with-support) for softmax. This PR modifies the softmax operation to benefit from the overload and also account for when the `arg` is a reference.


khuck:
@hkaiser are we back to using blaze HEAD now?

hkaiser:
@khuck uggh - I promised to keep you in the loop... Will write an email to the list.

khuck:
@hkaiser no prob - I just rebased to master to run tests, and saw an error, and looked at the recent pull requests.  Thanks!

PR #778, opened 2019-02-11
Making the ctest timeout 200 seconds for Debug

### LINKED ISSUE #777: tests.unit.plugins.booleans.where_operation times out with Debug build
one buildbot failure with the Debug configuration - tests.unit.plugins.booleans.where_operation will time out after 100 seconds.  This is with hpx, boost and phylanx all built with Debug.  Every other test finishes in time.  tests.unit.execution_tree.generate_tree also takes at least 96 seconds.  Most other tests are less than 10 seconds, some are between 10 and 40 seconds.  Should we increase the timeout for Debug build tests?  In isolation, it takes ~135 seconds to run.

khuck:
Some of the tests (tests.unit.plugins.booleans.where_operation for
example) take close to (or more than) 100 seconds to execute in the
debug build.  We should look into why these tests are so slow, but
in the meantime, the builds shouldn't fail.  They will continue to
fail on KNL, where even the Release build of
tests.unit.plugins.booleans.where_operation takes 40+ seconds.

This fixes #777

### DIFF
--- a/cmake/Phylanx_AddTest.cmake
+++ b/cmake/Phylanx_AddTest.cmake
@@ -51,6 +51,10 @@ macro(add_phylanx_test category name)
     add_test(
       NAME "${category}.${name}"
       COMMAND ${cmd} ${args})
+      # The debug tests can take longer than 100 seconds - so double the timeout.
+      if("${CMAKE_BUILD_TYPE}" STREQUAL "Debug")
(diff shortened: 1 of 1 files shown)


khuck:
...by the way, nearly 90% of the time is spent in the /hstack$0/0$1$1::eval() call(s).

PR #782, opened 2019-02-15
Adding insert primitive

scheherzade:
This PR implements insert primitive for cases where ```axis = None```.

### DIFF
--- a/src/plugins/matrixops/insert.cpp
+++ b/src/plugins/matrixops/insert.cpp
@@ -1,0 +1,368 @@
+//   Copyright (c) 2017 Shahrzad Shirzad
+//
+//   Distributed under the Boost Software License, Version 1.0. (See accompanying
+//   file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+#include <phylanx/config.hpp>
+#include <phylanx/ir/node_data.hpp>
+#include <phylanx/plugins/matrixops/insert.hpp>
+#include <phylanx/util/matrix_iterators.hpp>
+
+#include <hpx/include/lcos.hpp>
+#include <hpx/include/naming.hpp>
+#include <hpx/include/util.hpp>
+#include <hpx/throw_exception.hpp>
+#include <hpx/util/optional.hpp>
+
+#include <cmath>
+#include <cstddef>
+#include <cstdint>
+#include <memory>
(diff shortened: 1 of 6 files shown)


src/plugins/matrixops/insert.cpp
```diff
@@ -0,0 +1,362 @@
...
+#include <cmath>
+#include <cstddef>
+#include <cstdint>
+#include <memory>
+#include <string>
+#include <utility>
+#include <vector>
+
+#include <blaze/Math.h>
+
```
hkaiser: You should (conditionally) #include BlazeTensor as well

src/plugins/matrixops/insert.cpp
```diff
@@ -0,0 +1,362 @@
...
+        std::string const& codename)
+      : primitive_component_base(std::move(operands), name, codename)
+      , dtype_(extract_dtype(name_))
+    {
+    }
+
+    ///////////////////////////////////////////////////////////////////////////
+
+    template <typename T>
+    std::size_t insert::get_vec(ir::node_data<T> const& arg) const
```
hkaiser: Would it be better to chose a more fitting name, like `get_vector_size` or similar? Additionally, this function could be `static` (but that's a very minor detail).

src/plugins/matrixops/insert.cpp
```diff
@@ -0,0 +1,362 @@
...
+    {
+        switch (arg.num_dimensions())
+        {
+        case 0:
+            return 1;
+        case 1:
+            return arg.size();
+        case 2:
+            return arg.dimension(1);
+        default:
```
hkaiser: Shoudl this support 3D data as well?

src/plugins/matrixops/insert.cpp
```diff
@@ -0,0 +1,362 @@
...
+                return a;
+            });
+
+        blaze::DynamicVector<T> sorted_indices(indices_vec.size());
+
+        std::iota(sorted_indices.begin(), sorted_indices.end(), 0);
+        std::stable_sort(sorted_indices.begin(), sorted_indices.end(),
+            [&](double c, double b) {
+                return indices_vec[c] < indices_vec[b];
+            });
```
hkaiser: Why is this using `double`'s as indices? Is that intended?

Issue #684, opened 2018-12-07
Primitives needed for Keras backend


taless474:
We can start with:
- [x] A max constant similar to `numpy.inf` (see #699)
- [x] NumPy equivalents of:
    - [x] `tanh`
    - [x] `var` (see #706)
    - [x] `std` (see #706)
    - [x] `prod` (see #692)
    - [x] `cumprod` (#758)
    - [x] `clip` (see #731)
    - [x] `reshape` (see #691)
    - [x] `flatten` (see #685)
    - [x] `repeat` (see #726)
    - [x] `tile` (see #694)
    - [x] `dtype.name` (#806, #856)
    - [x] `flip` (see #701)
    - [x] `ndim` (see #730)
    - [x] `zeros`
- [x] Softmax function (Blaze has an implementation, see #688)
- [x] `logsumexp` to compute the log of the sum of exponentials of input elements as in `scipy.misc.logsumexp` (see #734)

ct-clmsn:
@taless474 great idea!

ct-clmsn:
@taless474 not sure if Keras does auto-differentiation but, if it does, would integrating [tangent](https://github.com/google/tangent) be useful?

taless474:
@ct-clmsn Thank you for introducing the tangent. I will study it. For implementing gradients I was looking into [JAX](https://github.com/google/jax).

taless474:
We can also add NumPy equivalents of:
- [ ] `concatenate`
- [ ] `maximum`
- [ ] `minimum`

taless474:
- We need to update:
     - [ ] `all` where it gets `axis` and `keepdims` as arguments.
     - [ ] `transpose` to accept `axes`

- To support CNNs we can add NumPy equivalents of:
    - [ ] `tensordot`
    - [ ] `pad` in its `constant` mode
    - [ ] `flipud` and `fliplr`

- It's good to have Numpy equivalents of:
    - [ ] `insert`
    - [ ] `sign`
    - [ ] `asarray` (I mean a function to convert input array to node_data)

ct-clmsn:
@taless474 would baking in support for the [ONNX](https://onnx.ai/) model format be a reasonable follow up or additional feature? If keras is supported ya'll might be able to load ONNX models built with other toolkits into Phylanx and run them...

hkaiser:
@rtohid would you see a way to implement functionality equivalent to `dtype.name` in the frontend?

rtohid:
[...]

PR #783, opened 2019-02-15
Adding Tensordot primitive

taless474:
This PR adds the equivalents of [NumPy tensordot](https://docs.scipy.org/doc/numpy/reference/generated/numpy.tensordot.html) and [NumPy outer](https://docs.scipy.org/doc/numpy/reference/generated/numpy.outer.html) to the current `dot_operation` except for the case of `tensordot` that has a range of two ranges `axes`.
Working toward #684


khuck:
@hkaiser @taless474 This PR broke the buildbot build (see http://ktau.nic.uoregon.edu:8020/#/builders/8/builds/415) - and yes, buildbot always builds against the head of Blaze:
```
In file included from /var/lib/buildbot/slaves/phylanx/x86_64-gcc7-release/build/src/plugins/matrixops/dot_operation_double.cpp:11:0:
/var/lib/buildbot/slaves/phylanx/x86_64-gcc7-release/build/phylanx/plugins/matrixops/dot_operation_impl.hpp: In member function ‘blaze::DynamicVector<T> phylanx::execution_tree::primitives::dot_operation::convert_to_1d(phylanx::ir::node_data<T>&&) const’:
/var/lib/buildbot/slaves/phylanx/x86_64-gcc7-release/build/phylanx/plugins/matrixops/dot_operation_impl.hpp:46:40: error: ‘ravel’ is not a member of ‘blaze’
             return blaze::trans(blaze::ravel(arr.matrix()));
                                        ^~~~~
```

hkaiser:
@khuck Right. Now it the moment at which you need to install [BlazeTensor](https://github.com/STEllAR-GROUP/blaze_tensor) and we will have to stop pretending its optional. 

@taless474  would it make sense to have the ravel stuff used optionally only? Or would that actually break things?

The `blaze::ravel` functionality is fully implemented by BlazeTensor (even for 2D types). There might be a time when this is integrated into blaze itself, but that might take a while.

taless474:
I moved the 2D use of `blaze::ravel` under the `PHYLANX_HAVE_BLAZE_TENSOR` and I hope it is fine for now.

PR #784, opened 2019-02-16
Add symbol table infrastructure.

rtohid:


hkaiser:
@rtohid could you please explain what this PR adds? Is there any new functionality enabled by this? Is it pure infrastructure?

rtohid:
@hkaiser This only provides the infrastructure. I will add the code to generate the IR and PhySL, like so:
https://gist.github.com/rtohid/ebe8596f4ce1181234fe09d8e1a53ea1

I haven't include the above code as I would like to separate the generation of the IR from that of PhySL.

PR #786, opened 2019-02-19
PHYSL_IR environment variable

ct-clmsn:
Added in support for loading a base64 encoded PhySL dump from an environment variable `PHYSL_IR`. uses std::getenv to access environment variables and a function extracted from Boost.Beast to decode base64 encoded strings (function was extracted to remove a boot.beast dependency).

Boost.Beast's base64 decoder has been tested locally against Python's base64 encoder and it works. Not sure if that test code needs to be added into the PR.

Also added command-line flag `no-ast-env` to tell the interpreter to skip checking for the `PHYSL_IR` environment variable.

> This PR had a bunch of superfluous files added into it that have since been removed, my Git skills continue to improve. 2d6af19 is where the actual work started.


ct-clmsn:
should the string in the PHYSL_IR also be zip compressed?

hkaiser:
> should the string in the PHYSL_IR also be zip compressed?

@ct-clmsn it's probably not worth it at this point. Let's reconsider once we end up in the MBytes range with our scripts ;-).

examples/interpreter/physl.cpp
```diff
@@ -45,6 +47,14 @@ std::string read_user_code(std::string const& path)
     return str_stream.str();
 }
 
+std::tuple<bool, std::string> get_env_physl_ir() {
+    auto e = boost::this_process::environment();
+    auto env_physl_ir = e.find("PHYSL_IR");
```
hkaiser: A simple `std::getenv("PHYSL_IR")` would probably do the trick here (https://en.cppreference.com/w/cpp/utility/program/getenv) avoiding to pull in Boost.Process just for this. This would also fix the compilation errors we're seeing on CircleCI caused by missing #includes in Boost.Process.

examples/interpreter/physl.cpp
```diff
@@ -45,6 +47,14 @@ std::string read_user_code(std::string const& path)
...
+    if(env_physl_ir == e.end()) { return std::make_tuple(false, std::string{}); }
+    auto physl_ir_str = boost::beast::detail::base64_decode(env_physl_ir.second);
+    return std::make_tuple(true, physl_ir_str);
```
hkaiser: May I suggest to actually have the Base64 decoding included in this file directly? We try to reduce dependencies on Boost and this would add another Boost library to the mix. Aditionally, Boost.Beast was added to Boost just recently, so this wouldn't compile for older versions.

Here is the code:
```
std::string base64_decode(std::string const& encoded_string) {
  int in_len = encoded_string.size();
  int i = 0;
  int j = 0;
  int in_ = 0;
  unsigned char char_array_4[4], char_array_3[3];
  std::string ret;

  while (in_len-- && ( encoded_string[in_] != '=') && is_base64(encoded_string[in_])) {
    char_array_4[i++] = encoded_string[in_]; in_++;
    if (i ==4) {
      for (i = 0; i <4; i++)
        char_array_4[i] = base64_chars.find(char_array_4[i]);

      char_array_3[0] = ( char_array_4[0] << 2       ) + ((char_array_4[1] & 0x30) >> 4);
      char_array_3[1] = ((char_array_4[1] & 0xf) << 4) + ((char_array_4[2] & 0x3c) >> 2);
      char_array_3[2] = ((char_array_4[2] & 0x3) << 6) +   char_array_4[3];

      for (i = 0; (i < 3); i++)
        ret += char_array_3[i];
      i = 0;
    }
  }

  if (i) {
    for (j = 0; j < i; j++)
      char_array_4[j] = base64_chars.find(char_array_4[j]);

    char_array_3[0] = (char_array_4[0] << 2) + ((char_array_4[1] & 0x30) >> 4);
    char_array_3[1] = ((char_array_4[1] & 0xf) << 4) + ((char_array_4[2] & 0x3c) >> 2);

    for (j = 0; (j < i - 1); j++) ret += char_array_3[j];
  }

  return ret;
}
```

examples/interpreter/physl.cpp
ct-clmsn: Right! I have a fix in the works.

examples/interpreter/physl.cpp
ct-clmsn: Will add that into the commit. Should that code have a home in physl.cpp or another file?

hkaiser: For now I think it's fine to have it right here. We'll go and move it if we come across another use somewhere else.

examples/interpreter/physl.cpp
```diff
@@ -160,6 +170,8 @@ int handle_command_line(int argc, char* argv[], po::variables_map& vm)
                 "of the created execution tree as a Newick tree to a file")
             ("transform,t", po::value<std::string>(),
                 "file to read transformation rules from")
+            ("no-ast-env,e", po::value<std::string>()->implicit_value("<none>"),
```
parsa: The default typically is to read from STDIN and read as is. Maybe I suggest providing a flag that does the opposite and reads base64 if the flag is present?

ct-clmsn: @parsa great question! the environment variable is meant to hold the IR and enable a cloud deployment of phylanx to a remote system for batch execution. a user can tell the python-side of phylanx to deploy the phylanx interpreter, as a container, using a cloud scheduler. to make that work without having to manage socket handshakes between the python-side and the remote system, the IR should be deployed with the container. packing IR into an environment variable is arguably tacky, but makes it possible to ship IR with the container job description for remote deployments.

ct-clmsn: @hkaiser @parsa i've been thinking if the environment variable is too tacky, a command line flag that accepts base64 encoded IR would also work.

ct-clmsn: @parsa I've added in the command line argument would "b64-physl" work?

parsa: @ct-clmsn Sure. Or `--base64,b` or something users can remember.

ct-clmsn: @parsa thanks! will add that into the fix.

ct-clmsn:
@hkaiser the code you suggested integrating had a missing function, i fitted it with code from this [link](https://github.com/ReneNyffenegger/cpp-base64/blob/master/base64.cpp).

hkaiser:
> the code you suggested integrating had a missing function, i fixed it with code from this link.

@ct-clmsn Thanks! That's where I copied the code from initially but had missed the `is_base64()`

PR #790, opened 2019-02-21
Add sign and square to support the Keras backend.

stevenrbrandt:
Add sign and square to support the Keras backend.


taless474:
Thank you.

PR #797, opened 2019-02-21
Making sure [hvd]stack() can be called with a range.

### LINKED ISSUE #792: Conversion of range to array by hstack requires a variable name
I've noticed that in PhySL code,

```
cout(hstack(range(1,4)))
```
fails. However,

```
define(v,range(1,4)),
cout(hstack(v))
```
Succeeds Not sure why a named variable helps.

hkaiser:
Fixes #792


hkaiser:
@stevenrbrandt could you please verify whether this fixes the issue you reported?

PR #798, opened 2019-02-22
hyperlda

ct-clmsn:
numpy implementation of hyper-parameterized lda as per [mallet](https://github.com/mimno/Mallet)


ct-clmsn:
@hkaiser if i need to modify the portions that use [numpy.ma](https://docs.scipy.org/doc/numpy/reference/maskedarray.html), just say the word!

ct-clmsn:
@hkaiser flake8 *really* hates my conditional expressions.

hkaiser:
> i'll be adding a phylanx annotated version of this code next. 

Thanks! Please let us know if you run into missing capabilities on the Phylanx end.

> do ya'll want me to hand-write this code in PhySL?

The benefit of this would be minimal, I think. We can always generate the PhySL version from your @Phylanx-annotated code, if needed.

ct-clmsn:
@hkaiser I gave the code a run through with the @Phylanx('PhySL') annotation; it crashed with a 'src/tcmalloc.cc:278] Attempt to free invalid pointer' error. What method do ya'll use for debugging these issues?

justwagle:
@ct-clmsn try ld preloading the tcmqlloc library. That should help. Looks like python is using the system allocator and you built phylanx with tcmalloc.

ct-clmsn:
@justwagle thanks! i tried `LD_LOAD=<path-to-libtcmalloc> python phylanx_lra_csv.py` and got the same issue. will continue to fiddle until it works. oh, i'm using ubuntu gcc 7.3.0.

@hkaiser fair enough! thanks for clearing the PR! will shift back to randomforest.

hkaiser:
@ct-clmsn Could you try `LD_PRELOAD=<path-to-libtcmalloc> python phylanx_lra_csv.py` instead, please?

ct-clmsn:
@justwagle @hkaiser we are in business! thanks again! i'm now getting transformation errors! progress!

PR #805, opened 2019-02-26
Add support to eval method.

rtohid:
Adds a new configuration option, 'lazy', to the Phylanx decorator. If set, the evaluation of the function is postponed untill the 'eval' method is invoked.

### DIFF
--- a/python/phylanx/ast/physl.py
+++ b/python/phylanx/ast/physl.py
@@ -350,7 +350,6 @@ class PhySL:
         self.ir = self.apply_rule(tree.body[0])
         check_return(self.ir)
         self.__src__ = self.generate_physl(self.ir)
-        self.lazy = kwargs.get('lazy')
 
         if self.kwargs.get("debug"):
             print_physl_src(self.__src__)
@@ -423,7 +422,7 @@ class PhySL:
 
         return result
 
-    def call(self, args):
+    def lazy(self, args):
         self.__perfdata__ = (None, None, None)
         self.performance_primitives = None
 
@@ -446,11 +445,11 @@ class PhySL:
             self.is_compiled = True
 
         self.args = args
-        if self.lazy:
-            return self
+        return self
 
-        result = self.eval()
-        return result
+    def call(self, args):
+        self.lazy(args)
+        return self.eval()
 
 # #############################################################################
 # Transducer rules
--- a/python/phylanx/ast/physl.py
+++ b/python/phylanx/ast/physl.py
@@ -350,6 +350,7 @@ class PhySL:
         self.ir = self.apply_rule(tree.body[0])
         check_return(self.ir)
         self.__src__ = self.generate_physl(self.ir)
+        self.lazy = kwargs.get('lazy')
 
         if self.kwargs.get("debug"):
             print_physl_src(self.__src__)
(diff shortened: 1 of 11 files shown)


hkaiser:
@rtohid is it really a good idea to hide the two completely different modes of operation behind the same interface? Wouldn't it be better to make this more explicit by exposing it to the user? Perhaps something that would require to write:
```
def lazy_eval_some_phylanx_functionality():
    @Phylanx
    def eval_some_phylanx_functionality():
        return some_phylanx_functionality()
    return eval_some_phylanx_functionality.code()
```
this would be equivalent to your proposed solution, but more explicit. If the user needs both, lazy and direct evaluation, this could be written as:
```
@Phylanx
def eval_some_phylanx_functionality():
    return some_phylanx_functionality()

def lazy_eval_some_phylanx_functionality():
    return eval_some_phylanx_functionality.code()
```

hkaiser:
@rtohid now, that I think of this some more, wouldn't even this solution be sufficient:
```
@Phylanx
def eval_some_phylanx_functionality():
    return some_phylanx_functionality()

def lazy_eval_some_phylanx_functionality():
    return eval_some_phylanx_functionality    # return callable
```
In this case we wouldn't even need to apply any changes to the existing scheme.

rtohid:
@hkaiser Thanks. You are right, the current (per function) approach could be confusing and difficult to track. My goal was to provide the same interface as what Keras uses. Do you think it make sense to keep the interface but set the mode of operation globally- through PhylanxSession?
Something like:
```py

PhylanxSession(num_threads=1, lazy=True)

@Phylanx
def foo(x):
    return x

bound_foo = foo(2)
result = bound_foo.eval()
```

tests/unit/python/execution_tree/lazy_eval.py
```diff
@@ -0,0 +1,43 @@
...
+assert (evaluated_foo == expected).all()
+
+
+@Phylanx
+def eye_eager(size):
+    return np.eye(size)
+
+
+def eye(size):
+    return eye_eager.lazy(3)
```
hkaiser: This loses the argument. Is that intended?

taless474:
`lazy` works well for functions; is it possible to apply it also to variables?
```python
from keras import backend as K
val = np.array([[1, 2], [3, 4]])
kvar = K.variable(value=val, dtype='float64', name='example_var')
K.eval(kvar)
```
>array([[ 1.,  2.],
       [ 3.,  4.]])

Using a `phylanx.execution_tree.var(np.array(value, dtype))` for variables, I get this error:
```log
RuntimeError: C:\Repos\keras\keras\backend\phylanx_backend.py(49, 11): shape:: primitive_argument_type does not hold a numeric value type (type held: 'phylanx::execution_tree::primitive'): HPX(bad_parameter)
```

hkaiser:
@taless474 What backend function is called for:
```
kvar = K.variable(value=val, dtype='float64', name='example_var')
```
?

taless474:
An example can be found here:
https://gist.github.com/taless474/ea52749781cbcf26cb5f144971bfa28f

hkaiser:
See here for a small reproducing example: https://gist.github.com/taless474/ea52749781cbcf26cb5f144971bfa28f

python/phylanx/ast/physl.py
```diff
@@ -422,27 +436,19 @@ def call(self, args):
...
                 PhySL.compiler_state = compiler_state()
 
             phylanx.execution_tree.compile(
                 self.file_name, self.__src__, PhySL.compiler_state)
             self.is_compiled = True
 
-        result = phylanx.execution_tree.eval(
-            self.file_name, func_name, PhySL.compiler_state, *args)
+        self.args = args
+        return self
```
hkaiser: This will fail if `lazy()` is called more than once to retrieve differently bound code objects before `eval()` is invoked on those. I think `lazy()` should return an object that exposes an `eval()` function and stores the `args` array in order to pass the correct arguments once invoked.

hkaiser:
@taless474 Your example code should now work as expected, please verify.
@rtohid Please verify the correctness of my changes to the PhySL transducer

python/phylanx/ast/physl.py
hkaiser: This was fixed by 801e117

taless474:
Thanks for the effort. Right now, tests that have only one variable pass but the ones with more than one variable fail. When using the second variable, the first one contains only zeros.
I will put a reproducing program shortly.

taless474:
[This](https://gist.github.com/taless474/6c92651f2e0954cdbf6d12b03b46f241) is a reproducing program.
The program works if arrays are directly passed to it (if we don't use the `parse_shape_or_val` it works).

hkaiser:
@taless474 all should be well now. Thanks for reporting the problems!

hkaiser:
The deployment step fails for unrelated reasons... I'm merging this now.

PR #806, opened 2019-02-26
Adding functionality to get the phylanx internal type id and name.

stevenrbrandt:
Adding functionality to get the phylanx internal type id and name. Requested by Bita.


hkaiser:
@stevenrbrandt could we name it just `type()` and `name()`? Alternatively, let's start using namespaces and call it `phylanx.type()` and `phylanx.name()`.

stevenrbrandt:
@hkaiser I'd prefer not to call it type() if it doesn't do what Python's type() does. As for the namespaces, are other primitives in a namespace? If so, which ones? (Looking for an example to use as a template)

hkaiser:
@stevenrbrandt we don't have any examples for using namespaces yet.

src/execution_tree/primitives/phyname.cpp
```diff
@@ -0,0 +1,108 @@
...
+                    name = "range";
+                    break;
+                case 8:
+                    name = "dict";
+                    break;
+                default:
+                    name  = "??";
+                    break;
+                }
+                return primitive_argument_type{name};
```
hkaiser: This functionality is already available from the main phylanx module (see https://github.com/STEllAR-GROUP/phylanx/blob/master/src/execution_tree/primitives/base_primitive.cpp#L65). It just needs to be exposed to the world (as its currently a static/internal function).

src/execution_tree/primitives/phytype.cpp
```diff
@@ -0,0 +1,81 @@
...
+                // If a single None is passed in python, we get a
+                // zero length arg list here.
+                if(args.size() == 0) {
+                    std::int64_t tid = 0;
+                    return primitive_argument_type{tid};
+                } else {
+                    std::int64_t tid = args[0].index();
+                    return primitive_argument_type{tid};
+                }
+            }),
```
hkaiser: Once #820 has been merged this 'fix' should not be needed anymore.

phylanx/execution_tree/primitives/base_primitive.hpp
```diff
@@ -927,6 +927,13 @@ namespace phylanx { namespace execution_tree
         std::string const& name = "",
         std::string const& codename = "<unknown>",
         eval_context ctx = eval_context{});
+
+    namespace detail
+    {
+        ///////////////////////////////////////////////////////////////////////
+        // Get the string representation for the type
+        char const *const get_primitive_argument_type_name(std::size_t index);
```
hkaiser: This will need a `PHYLANX_EXPORT` for the symbol to be always visible outside the main Phylanx module.

PR #818, opened 2019-02-28
Add test for range::size(), handle empty int ranges

parsa:
This PR fixes a bug in `phylanx::ir::range::size` that causes this example to print an unexpected value:
```cpp
phylanx::ir::range r(-1);
std::printf("r.size(): %d \n", r.size())
// Actual output:
// r.size(): -1
// Expected output:
// r.size(): 0
```

### DIFF
--- a/src/ir/ranges.cpp
+++ b/src/ir/ranges.cpp
@@ -352,6 +352,12 @@ namespace phylanx { namespace ir
                 std::int64_t const step = int_range.step();
 
                 std::int64_t const distance_to_start = stop - start;
+
+                // Adding step to start must get closer to stop
+                if ((distance_to_start > 0) != (step > 0))
+                {
+                    return 0;
+                }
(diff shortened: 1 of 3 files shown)


src/ir/ranges.cpp
```diff
@@ -352,6 +352,12 @@ namespace phylanx { namespace ir
                 std::int64_t const step = int_range.step();
 
                 std::int64_t const distance_to_start = stop - start;
+
+                // Adding step to start must get closer to stop
+                if ((distance_to_start > 0) != (step > 0))
```
hkaiser: Shouldn't this include comparing against `0`? e.g. something like:
```
if ((distance_to_start >= 0) != (step >= 0))
```

parsa: It doesn't need to compare against `0`. I added a test case that explicitly tests that case. `step == 0` was not handled anywhere else. It is handled now.

Issue #810, opened 2019-02-27
Run tests in parallel


stevenrbrandt:
It takes a long time to run the tests. The machine I typically run on has 40 cores (80 hyperthreads). It would be nice if I could use this parallelism to get through the testsuite faster.

khuck:
I would definitely endorse this

hkaiser:
@stevenrbrandt Wouldn't using the `-j` flag in the `make tests` do the trick?

khuck:
@hkaiser it builds the tests in parallel, but then cmake runs them serial.

khuck:
@hkaiser @stevenrbrandt you can set the CTEST_PARALLEL_LEVEL environment variable before running the test: 
https://cmake.org/cmake/help/v3.12/manual/ctest.1.html

stevenrbrandt:
@khuck interesting

khuck:
@stevenrbrandt @hkaiser I tried it out on the Phylanx build - the tests were run concurrently, but it made all the tests run 10-60x slower.  Why?

hkaiser:
@khuck HPX binds threads to cores. As the concurrently running tests don't know anything about each other all of them bind their threads to the same cores. That puts all the burden of multiplexing things on those cores on the OS....

khuck:
@hkaiser ah, of course.

stevenrbrandt:
@khuck @hkaiser so what we need is some way to figure this out. Maybe if 
CTEST_PARALLEL_LEVEL is set we could disable the core bindings or do some other intelligent thing?

khuck:
Is there an HPX argument that tells it not to pin threads?  If so, we could use that.  The other problem is that we can detect that CTEST_PARALLEL_LEVEL has been set, but we can't detect if someone calls "ctest -j ..."

hkaiser:
@khuck: good point. `--hpx:bind=none` should do what you're asking for (no thread binding).

stevenrbrandt:
@hkaiser @khuck so that's an argument that has to be passed to the tests when they are invoked. That's going to be tricky to add to make files, integrate with the python libraries, etc. Can that be set with an environment variable?

hkaiser:
@stevenrbrandt I have created an HPX ticket asking for this functionality as it is not available at this point (see https://github.com/STEllAR-GROUP/hpx/issues/3724).

hkaiser:
[...]

PR #824, opened 2019-03-01
Check CircleCI test coverage

parsa:
This PR adds a check to CircleCI that ensures all CircleCI builds run all available tests. It adds a small Python script that checks if there are any Phylanx tests that are not run by CircleCI.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -144,6 +144,15 @@ jobs:
                 paths:
                     - phylanx/build
                     - phylanx/source
+    check.test_coverage:
+        <<: *docker_config
+        working_directory: /phylanx/source
+        steps:
+            - <<: *attach_phylanx_tree
+            # Test coverage
+            - run:
+                name: Ensure all CTest tests are run
+                command: tools/circleci/check_test_coverage.py .circleci/config.yml <(cd /phylanx/build && ctest -N | grep -o 'tests[^ ]\+')
     build:
         <<: *docker_config
         working_directory: /phylanx/build
@@ -546,6 +555,9 @@ workflows:
             - configure:
                 requires:
                     - checkout_code
+            - check.test_coverage:
+                requires:
+                    - configure
             - build:
                 requires:
                     - configure
--- a/tools/circleci/check_test_coverage.py
+++ b/tools/circleci/check_test_coverage.py
@@ -1,0 +1,85 @@
+#!/usr/bin/env python
+# Copyright (c) 2019 Parsa Amini
+#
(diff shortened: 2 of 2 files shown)


khuck:
Why not just run all tests?  (i.e. ```cmake tests```)

parsa:
@khuck Because we run the tests in several separate jobs that can run in parallel.

khuck:
@parsa That'll just make it run slower.  See https://github.com/STEllAR-GROUP/phylanx/issues/810

parsa:
@khuck I don't see the connection. These jobs are run on different containers, most likely on different VMs on different physical machines.

khuck:
@parsa ok.  So you're saying each test suite does its own build?  Or that there is one build, and multiple test suites use the same build output?  Regardless, making all the tests (i.e. ```cmake tests```) run in parallel (for the general case) means that things aren't missed.  I keep getting failures on the buildbot servers that pass CI, and they shouldn't - unless CI isn't building and running _all_ the tests.

parsa:
@khuck Take a look: <https://circleci.com/workflow-run/d74e58b3-0b0a-4951-b95a-b2abe1be9ede>
`make all` is done in a separate job. The output from that is fed into test group jobs. Each test group job builds the tests it is supposed to and runs them. We used to do `make tests` some time ago, but breaking the building and running of the tests into several jobs improved the CI turnaround time. If you have better ideas feel free to implement them.

khuck:
I'm not authorized to see the CI output, apparently.

I don't have better ideas, I'm just saying the CI is missing obvious build/test failures.  See http://ktau.nic.uoregon.edu:8020/#/builders/8/builds/440/steps/14/logs/stdio (for example).  If you look at all the builders (http://ktau.nic.uoregon.edu:8020/#/builders) you'll see that none of the buildbot tests have actually worked in over a week, and even then, that was only one build configuration.

parsa:
@khuck I don't know what's preventing you from seeing that page. Anyone who is a member of the repo should be able to login with their Github account and see it. It looks like this:

untitled

I'm pretty sure [BuildBot's build 440](http://ktau.nic.uoregon.edu:8020/#/builders/8/builds/440/steps/14/logs/stdio) is not failing because of us not building a target on CircleCI. `tests.unit.plugins.controls` tests build and run without an issue: (e.g. <https://circleci.com/gh/STEllAR-GROUP/phylanx/18912>).

So far we've only had two cases of tests not actually being built and executed on CircleCI and this PR exists to make sure we won't have more.

parsa:
@khuck Trying logging out and logging back in with your Github account on CircleCI. That usually solves the issue.

PR #834, opened 2019-03-07
Fixing 0dim case for random

### LINKED ISSUE #827: Random primitive shape for 0-D arrays is always 1
Having
```py
@Phylanx
def random_uniform(shape, minval, maxval, dtype=None, seed=None):
	return random(shape, list("uniform", minval, maxval))
print(random_uniform(10, 0.0, 1.0))
```
one would expect 10 random generated numbers in [0, 1.) while only 1 number is generated.

hkaiser:
Fixes #827


taless474:
Thank you.

Issue #832, opened 2019-03-06
Phylanx Performance Data Broken


stevenrbrandt:
The following code no longer works:
```
from phylanx import Phylanx

@Phylanx(performance="x")
def fib(n):
    if n < 2:
        return n
    else:
        return fib(n-1)+fib(n-2)

fib(10)
```
Error message is:
```
Traceback (most recent call last):
  File "x.py", line 11, in <module>
    fib(10)
  File "/usr/local/lib/python3.6/dist-packages/phylanx-0.0.1-py3.6-linux-x86_64.egg/phylanx/ast/transducer.py", line 94, in __call__
    result = self.backend.call(args)
  File "/usr/local/lib/python3.6/dist-packages/phylanx-0.0.1-py3.6-linux-x86_64.egg/phylanx/ast/physl.py", line 465, in call
    return self.lazy(args).eval()
  File "/usr/local/lib/python3.6/dist-packages/phylanx-0.0.1-py3.6-linux-x86_64.egg/phylanx/ast/physl.py", line 432, in eval
    *self.args)
RuntimeError: invalid reference id: HPX(bad_parameter)
```

stevenrbrandt:
```
#0  0x00007feabcbc5611 in __cxa_throw () from /lib64/libstdc++.so.6
#1  0x00007feac1f5a81b in pybind11::detail::type_caster_base<phylanx::bindings::compiler_state>::operator phylanx::bindings::compiler_state& (
    this=0x7ffe3c5819c8) at /usr/local/include/pybind11/cast.h:884
#2  0x00007feac1f56977 in pybind11::detail::cast_op<phylanx::bindings::compiler_state&> (caster=...) at /usr/local/include/pybind11/cast.h:919
#3  0x00007feac1f50d01 in pybind11::detail::argument_loader<phylanx::bindings::compiler_state&, bool>::call_impl<std::vector<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::allocator<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > >, std::vector<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::allocator<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > > (*&)(phylanx::bindings::compiler_state&, bool), 0ul, 1ul, pybind11::detail::void_type> (this=0x7ffe3c5819c0, f=
[...]

PR #840, opened 2019-03-07
Fixing extraction of performance data in the frontend

hkaiser:
Fixes: #832


stevenrbrandt:
Seems to work. Thanks.

PR #850, opened 2019-03-08
Adding 3D support for the softmax operation

taless474:
This PR adds 3D support for the softmax operation.

### DIFF
--- a/src/plugins/matrixops/softmax_operation.cpp
+++ b/src/plugins/matrixops/softmax_operation.cpp
@@ -125,18 +125,17 @@ namespace phylanx { namespace execution_tree { namespace primitives
     primitive_argument_type softmax_operation::softmax3d_axis0(
         arg_type&& arg) const
     {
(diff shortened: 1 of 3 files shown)


src/plugins/matrixops/softmax_operation.cpp
```diff
@@ -109,13 +111,117 @@ namespace phylanx { namespace execution_tree { namespace primitives
...
+    }
+
+    ///////////////////////////////////////////////////////////////////////////
+#if defined(PHYLANX_HAVE_BLAZE_TENSOR)
+    primitive_argument_type softmax_operation::softmax3d_axis0(
+        arg_type&& arg) const
+    {
+        if (!arg.is_ref())
+        {
+            for (std::size_t i = 0; i != arg.tensor().rows(); ++i)
```
hkaiser: It might be slightly more efficient to call `arg.tensor()` once only...

Issue #852, opened 2019-03-09
Stack and concatenate need to accept lists


hkaiser:
This code currently fails even if it should succeed:
```
import numpy as np
from phylanx import Phylanx, PhylanxSession, execution_tree

PhylanxSession.init(1)
	
def variable(value, dtype=None, name=None, constraint=None):
    if constraint is not None:
        raise TypeError("Constraint is the projection function to be "
                        "applied to the variable after an optimizer update")
    return execution_tree.var(np.array(value, dtype))

def eval(func):
    return func.eval()
	
def parse_shape_or_val(shape_or_val):
	return shape_or_val, np.random.random(shape_or_val).astype(np.float32) - 0.5
	
##############################################################################	
@Phylanx
def concatenate_eager(tensors, axis):
	return np.concatenate(tensors, axis)

def concatenate(tensors, axis=-1):
	return concatenate_eager.lazy(tensors, axis)

@Phylanx
def stack_eager(x, axis):
	return np.stack(x, axis)

def stack(x, axis=0):
	return stack_eager.lazy(x, axis)
	
##############################################################################	
def test_concatenate_operation(x_shape_or_val, y_shape_or_val):
	x_shape, x_val = parse_shape_or_val(x_shape_or_val)
	y_shape, y_val = parse_shape_or_val(y_shape_or_val)
	t = concatenate([variable(x_val), variable(y_val)])
	z = eval(t)	
	return z
				

def test_stack_operation(x_shape_or_val, y_shape_or_val):
	x_shape, x_val = parse_shape_or_val(x_shape_or_val)
	y_shape, y_val = parse_shape_or_val(y_shape_or_val)
	t = stack([variable(x_val), variable(y_val)])
	z = eval(t)	
	return z				

print(test_concatenate_operation((4, 3), (4, 2)))
print(test_stack_operation((5,4,6), (5,4,6)))
```

hkaiser:
Adressing this problem requires to fix

- [ ] the `stack` primitive
- [ ] the `concatenate` primitive

hkaiser:
This should be fine now, please reopen if necessary.

PR #853, opened 2019-03-10
Making sure concatenate accepts lists of variables

hkaiser:
Working towards #852


taless474:
Having the example in #852, I get
```pytb
  File "test9.py", line 49, in <module>
    print(test_concatenate_operation((4, 3), (4, 2)))
  File "test9.py", line 38, in test_concatenate_operation
    z = eval(t)
  File "test9.py", line 13, in eval
    return func.eval()
  File "C:\Repos\phylanx\cmake-build-debug\python\build\lib.win-amd64-3.6\phylanx\ast\physl.py", line 431, in eval
    *self.args)
RuntimeError: test9.py(21, 8): concatenate:: all the input array dimensions except for the concatenation axis must match exactly : HPX(bad_parameter)
```

hkaiser:
@taless474 yes, you're right. If you explicitly specify the `axis=1` it works, however :/

hkaiser:
@taless474 this should be fine now, thanks for noticing.

PR #856, opened 2019-03-10
Adding support to execution_tree.var() to expose the numpy dtype of the value it holds

hkaiser:
- flyby: streamline (internal) __type() and __name() primitives

Working towards #684


hkaiser:
This PR enables writing:
```
def phylanx_type(x):
    return phylanx.execution_tree.var(x).dtype

assert phylanx_type(numpy.array([1.0, 2.0]) == numpy.dtype('float64')
```
and similar.

PR #858, opened 2019-03-11
Making sure initial value for statistics primitives works as advertized

### LINKED ISSUE #839: Cannot call statistics primitives with `initial`
The 4th argument in the statistics is the `initial` value. In physl calling
```console
amax([1,2,3,4],nil,false,10)
```
would result in
```log
physl: exception caught:
physl.txt(172, 1): amax:: primitive_argument_type does not hold an integer value type (type held: 'phylanx::ir::node_data<std::int64_t>'): HPX(bad_parameter)
```
and calling
```console
amax([1.,2.,3.,4.],0,true,10.)
```
would raise
```log
physl: exception caught:
physl.txt(173, 1): amax:: primitive_argument_type does not hold a floating point value type (type held: 'phylanx::ir::node_data<double>'): HPX(bad_parameter)
```

hkaiser:
Fixes #839

### DIFF
--- a/phylanx/plugins/statistics/statistics_base_impl.hpp
+++ b/phylanx/plugins/statistics/statistics_base_impl.hpp
@@ -474,7 +474,7 @@ namespace phylanx { namespace execution_tree { namespace primitives
         if (valid(initial))
         {
             initial_value =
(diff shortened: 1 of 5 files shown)


hkaiser:
@scheherzade thanks a lot!

PR #860, opened 2019-03-11
Sign primitive returns correct type

### LINKED ISSUE #849: The `sign` primitive returns float instead of int.
```py
from phylanx import Phylanx
import numpy as np

@Phylanx
def do_sign(n):
    return np.sign(n)

print('--')
print(np.sign(2))
print(do_sign(2))
print('--')
print(np.sign(0))
print(do_sign(0))
print('--')
print(np.sign(-2))
print(do_sign(-2))
print('--')
```

output:
```
--
1
1.0
--
0
0.0
--
-1
-1.0
--

```

hkaiser:
Fixes #849

### DIFF
--- a/phylanx/plugins/arithmetics/generic_operation.hpp
+++ b/phylanx/plugins/arithmetics/generic_operation.hpp
@@ -38,7 +38,7 @@ namespace phylanx { namespace execution_tree { namespace primitives
         using arg_type = ir::node_data<T>;
 
     public:
-        static std::vector<match_pattern_type> const match_data;
+        static std::vector<std::pair<match_pattern_type, bool>> const match_data;
 
         generic_operation() = default;
 
@@ -107,6 +107,7 @@ namespace phylanx { namespace execution_tree { namespace primitives
     private:
         std::string func_name_;
         node_data_type dtype_;
+        bool retain_argument_type_;
     };
 
     inline primitive create_generic_operation(hpx::id_type const& locality,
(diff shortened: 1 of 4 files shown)


stevenrbrandt:
I'm surprised that you chose to make this its own primitive rather than allowing the generic mechanism to return integer.

hkaiser:
> I'm surprised that you chose to make this its own primitive rather than allowing the generic mechanism to return integer.

The generic mechanism relies on returning the same type as the argument's type. Extending this would have caused a rewrite of the whole generics primitive. Extracting `sign` was a lot less effort...

stevenrbrandt:
@hkaiser I don't think this should be its own primitive. I think it should be another kind of generic primitive, one that doesn't automatically upcast all results to be doubles. If you do that, then np.square() should also move here--and possibly other primitives.

hkaiser:
@stevenrbrandt better now?

stevenrbrandt:
@hkaiser it is better, but there is no test of the new functionality.

stevenrbrandt:
@hkaiser want me to add a test?

hkaiser:
> want me to add a test?

I already added some (here: https://github.com/STEllAR-GROUP/phylanx/pull/860/files#diff-027e33a92665bf61eab6668eb58bcd47R755), but please feel free to add more.

stevenrbrandt:
@hkaiser OK, I see them now. Those seem good.

PR #865, opened 2019-03-12
Adding truncated_normal random number distribution

### LINKED ISSUE #864: Request for truncated normal distribution
A [truncated normal](https://keras.io/backend/#truncated_normal) has values generated follow a normal distribution with specified mean and standard deviation, except that values whose magnitude is more than two standard deviations from the mean are dropped and re-picked.
It should be able to accept `shape`, `mean`, `stddev`, `dtype` and `seed`

hkaiser:
Fixes #864

This enables to write:
```
@Phylanx
def truncated_normal(shape, mean=0.0, stddev=1.0, dtype=None, seed=None):
    if not seed is None:
        set_seed(seed)
    return random(shape, ["truncated_normal", mean, stddev])
```
The `dtype` is currently not supported by the `random` primitive. We should add that once we have found a way to support named arguments.

### DIFF
--- a/phylanx/include/util.hpp
+++ b/phylanx/include/util.hpp
@@ -10,10 +10,12 @@
 #include <phylanx/util/hashed_string.hpp>
 #include <phylanx/util/none_manip.hpp>
 #include <phylanx/util/performance_data.hpp>
+#include <phylanx/util/random.hpp>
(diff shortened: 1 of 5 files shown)


taless474:
It works, thanks.

Issue #857, opened 2019-03-11
Request for implementing Keras activation functions


taless474:
Keras has a few activation functions in its backend that we might want to implement in Phylanx. Phylanx supports `softmax` and `tanh` and we can add:
- [ ] relu
- [ ] elu
- [ ] softplus
- [ ] hard_sigmoid

hkaiser:
@taless474 would you mind adding links describing the actual math to implement for the primitives you listed?

taless474:
@hkaiser, I added the link to their description in the Keras documentation which contains their numpy implementation. Do you need the actual math behind them from scholar articles, too?

hkaiser:
@taless474 those links are fine, thank you!

taless474:
To support Keras, we can also add:
- [ ] [one_hot](https://keras.io/backend/#one_hot)
- [ ] [categorical_crossentropy](https://keras.io/backend/#categorical_crossentropy)
- [ ] [binary_crossentropy](https://keras.io/backend/#binary_crossentropy)
- [ ] [resize_images](https://keras.io/backend/#resize_images) where the interpolation is `bilinear`

taless474:
Keras slice has the same functionality as [TensorFlow slice](https://www.tensorflow.org/api_docs/python/tf/slice). Having an nD input array, `start` and `size` are vectors of size n. To slice a matrix, first we find the element with indices in the start (in the following example the `[0,2]` (the element in 0th row and 2nd column) element is `3`) as the most top left element of the out put submatrix, then we slice a submatrix with the given size (in the following example the size is `[2,1]`, so the output is `2x1`).
```py
m = tf.constant([[1,2,3],[4,5,6]])
s = tf.slice(m,[0,2],[2,1])
k.get_value(s) # [[3], [6]]
```
Slicing a `2x1` start at element in the 0th row and 2nd column, we get a `2x1` matrix as `[[3], [6]]`.
The following is another example of a `2x1x3` tensor. The element in position `(1,0,0)` is `4` and we get a `1x1x3` tensor by slicing it.
```py
t = tf.constant([[[1,2,3]],[[4,5,6]]])
s = tf.slice(t,[1,0,0],[1,1,3])
k.get_value(s) #[[[4, 5, 6]]]
```

hkaiser:
@taless474 thank you for you clarifications. IIUC, this is equivalent to the 'normal' python/numpy slicing:
```
[...]

PR #881, opened 2019-03-19
Adding l2_normalize operation

taless474:
This PR adds the functionality of Keras [l2_normalize](https://keras.io/backend/#l2_normalize) except for the case that the input is 3d and no axis is given (l2_normalize3d_flatten). It will be updated when `l2Norm` is added to the blaze_tensor.
It also modifies the transpose operation for the case of `axes=None`. The current version produces the correct response even if the nil value of `axes` is wrapped in a variable.
```py
import numpy as np

PhylanxSession.init(1)
	
def variable(value, dtype=None, name=None, constraint=None):
    return execution_tree.var(np.array(value, dtype))
	
@Phylanx
def permute_dimensions_eager(x, pattern):
	return np.transpose(x, pattern)

def permute_dimensions(x, pattern):
	return permute_dimensions_eager.lazy(x, pattern)

def eval(func):
    return func.eval()
	

def test_transpose_operation(axis):
	x_val = [[1,2,3],[4,5,6]]
	t1 = permute_dimensions(variable(x_val),axis)
	z1 = eval(t1)
	return z1
	
test_transpose_operation(None)
```
Working toward #857

### DIFF
--- a/phylanx/plugins/keras_support/keras_support.hpp
+++ b/phylanx/plugins/keras_support/keras_support.hpp
@@ -9,6 +9,7 @@
 #include <phylanx/plugins/keras_support/batch_dot_operation.hpp>
 #include <phylanx/plugins/keras_support/elu_operation.hpp>
 #include <phylanx/plugins/keras_support/hard_sigmoid_operation.hpp>
+#include <phylanx/plugins/keras_support/l2_normalize_operation.hpp>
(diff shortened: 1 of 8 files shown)


hkaiser:
@taless474 please rebase this branch for it to be merged.

hkaiser:
@taless474 should we merge this even without 3D support?

PR #882, opened 2019-03-19
Adding elu_operation primitive

JPenuchot:
Working towards #857

### DIFF
--- a/phylanx/plugins/keras_support/elu_operation.hpp
+++ b/phylanx/plugins/keras_support/elu_operation.hpp
@@ -52,7 +52,12 @@ namespace phylanx {  namespace execution_tree {  namespace primitives
             std::string const& name, std::string const& codename);
 
     private:
-        primitive_argument_type elu(mat_type&& arg, double alpha = 1.) const;
+        primitive_argument_type elu0d(mat_type&& arg, double alpha) const;
+        primitive_argument_type elu1d(mat_type&& arg, double alpha) const;
+        primitive_argument_type elu2d(mat_type&& arg, double alpha) const;
+#if defined(PHYLANX_HAVE_BLAZE_TENSOR)
+        primitive_argument_type elu3d(mat_type&& arg, double alpha) const;
+#endif
     };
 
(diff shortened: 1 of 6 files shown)


phylanx/plugins/keras_support/elu_operation.hpp
```diff
@@ -0,0 +1,72 @@
+// Copyright (c) 2018-2019 Hartmut Kaiser
+//
```
hkaiser: You should add your copyright.

src/plugins/keras_support/elu_operation.cpp
```diff
@@ -0,0 +1,207 @@
...
+        auto elu_ = [alpha](auto&& x)
+        {
+            return (x >= 0.) * ( x )
+                 + (x <  0.) * ( alpha * (std::exp(x) - 1.) );
+        };
+
+        return primitive_argument_type{mat_type::storage1d_type{
+            arg.is_ref() ? blaze::map(arg.vector(), elu_)
+                : blaze::map(std::move(arg.vector()), elu_)
+        }};
```
hkaiser: Are you sure that `blaze::map` essentially turns into an in-place operation when used on rvalues? … interesting, I was not aware of this.

hkaiser:
@JPenuchot Please rebase your branch on top of master to resolve the conflicts.

hkaiser:
Also, please fix the inspect errors reported for your code (see here: https://22249-101409977-gh.circle-artifacts.com/0/artifacts/phylanx_inspect_report.html)

src/plugins/keras_support/elu_operation.cpp
JPenuchot: I looked it up, it only accepts `const&` parameters. Do you want me to remove that ?

hkaiser: We should do different things as if `arg.is_ref()` we should allocate new memory, otherwise the operation should be performed in-place. `blaze::map` always allocates new memory, IIRC.

JPenuchot:
I fixed the copy/move semantics in my last commit, it should be good now

Issue #866, opened 2019-03-12
The problem with the `dtype` as an argument


taless474:
```py
import numpy as np
from phylanx import Phylanx


@Phylanx
def fx1(x):
	return np.eye(x, dtype='float')

print(fx1(5))  # OK


@Phylanx
def fx2(x, d):
	return np.eye(x, dtype=d)

print(fx2(5, 'float'))  # ValueError: dtype must a be string literal.


@Phylanx
def fx3(x, dtype):
	return np.eye(x, dtype=dtype)

print(fx3(5, 'float'))  # ValueError: dtype must a be string literal.
```
results in
```pytb
[[1. 0. 0. 0. 0.]
 [0. 1. 0. 0. 0.]
 [0. 0. 1. 0. 0.]
 [0. 0. 0. 1. 0.]
 [0. 0. 0. 0. 1.]]
Traceback (most recent call last):
  File "test18.py", line 12, in <module>
    @Phylanx
  File "C:\Repos\phylanx\cmake-build-debug\python\build\lib.win-amd64-3.6\phylanx\ast\transducer.py", line 104, in Phylanx
    return __PhylanxDecorator(__phylanx_arg)
  File "C:\Repos\phylanx\cmake-build-debug\python\build\lib.win-amd64-3.6\phylanx\ast\transducer.py", line 48, in __init__
    self.backend = self.backends_map[self.backend](f, python_ast, kwargs)
  File "C:\Repos\phylanx\cmake-build-debug\python\build\lib.win-amd64-3.6\phylanx\ast\physl.py", line 350, in __init__
    self.ir = self.apply_rule(tree.body[0])
  File "C:\Repos\phylanx\cmake-build-debug\python\build\lib.win-amd64-3.6\phylanx\ast\physl.py", line 394, in apply_rule
    return eval('self._%s' % node_name)(node)
  File "C:\Repos\phylanx\cmake-build-debug\python\build\lib.win-amd64-3.6\phylanx\ast\physl.py", line 840, in _FunctionDef
    body = self.block(node.body)
  File "C:\Repos\phylanx\cmake-build-debug\python\build\lib.win-amd64-3.6\phylanx\ast\physl.py", line 400, in block
    block = tuple(map(self.apply_rule, node))
  File "C:\Repos\phylanx\cmake-build-debug\python\build\lib.win-amd64-3.6\phylanx\ast\physl.py", line 394, in apply_rule
    return eval('self._%s' % node_name)(node)
[...]

PR #884, opened 2019-03-20
Adding support for named arguments and default argument values for primitives

### LINKED ISSUE #873: Add support for named arguments to frontend
@rtohid As discussed, the frontend should be changed to read all patterns as defined by the backend to figure out the placement of named arguments. We will start with adding the `dtype` named argument to the primitives listed in #866. 

A simple way of encoding this information in the patterns would be to use the convention `_N_dtype` (instead of the currently used `_N` placeholder -- where `N` is a sequence number of the argument). For instance, the `identity` primitive uses `identity(_1)` as its pattern today. That would be changed to `identity(_1, _2_dtype)` to specify the position where any `(… dtype=...)` argument should be passed.

If the frontend had support for analyzing the arguments listed in the primitive patterns, we were able to support arbitrary named arguments for all of them.

hkaiser:
This PR adds support for named arguments and default argument values for primitives in the PhySL compiler.

Fixes #873 
Working towards #866

- [x] add support for named arguments for primitives
- [x] add support for default arguments for primitives


hkaiser:
@rtohid Thanks a lot1

PR #893, opened 2019-03-21
Categorical crossentropy

stevenrbrandt:
First implementation of the categorical crossentropy function. Please take a look!


src/plugins/keras_support/categorical_crossentropy_operation.cpp
```diff
@@ -0,0 +1,245 @@
...
+
+    ///////////////////////////////////////////////////////////////////////////
+    categorical_crossentropy_operation::categorical_crossentropy_operation(primitive_arguments_type&& operands,
+        std::string const& name, std::string const& codename)
+      : primitive_component_base(std::move(operands), name, codename)
+    {}
+
+    primitive_argument_type categorical_crossentropy_operation::categorical_crossentropy0d() const
+    {
+        return primitive_argument_type{static_cast<double>(1.)};
```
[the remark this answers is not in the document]
stevenrbrandt: OK, I wasn't sure 0d was valid.

src/plugins/keras_support/categorical_crossentropy_operation.cpp
```diff
@@ -0,0 +1,245 @@
...
+#endif
+
+    ///////////////////////////////////////////////////////////////////////////
+    hpx::future<primitive_argument_type> categorical_crossentropy_operation::eval(
+        primitive_arguments_type const& operands,
+        primitive_arguments_type const& args,
+        eval_context ctx) const
+    {
+        #if 0
+        if (operands.empty() || operands.size() > 2)
```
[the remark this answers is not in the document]
stevenrbrandt: Notice the "#if 0". I should fix that.

src/plugins/keras_support/categorical_crossentropy_operation.cpp
```diff
@@ -0,0 +1,248 @@
...
+
+    ///////////////////////////////////////////////////////////////////////////
+#if defined(PHYLANX_HAVE_BLAZE_TENSOR)
+
+    primitive_argument_type cat_cross_operation::cat_cross3d(
+        arg_type&& target, arg_type&& output, bool from_logits) const
+    {
+        if(from_logits)
+        {
+            output.tensor() = blaze::softmax(output.tensor());
```
taless474: I think this does not work. `blaze::softmax` calculates `softmax` on the whole tensor, while Keras softmax calculates softmax where `axis=-1` in its default case. In the given NumPy implementation, Keras softmax is used. We have a softmax implementation in Phylanx which imitates its corresponding functionality in Keras (calculates on `axis=-1` where no axis is given). That is why I suggested categorical cross-entropy to be added to the softmax primitive.

taless474: As an example
```py
import keras.backend as K
x = K.constant([[[1, 2, 3, 4],[5, 6, 7, 8],[9, 10, 11, 12]],
       [[-1, -2, -3, -4], [-5, -6, -7, -8], [-9, -10, -11, -12]]])
a = K.softmax(x)
K.get_value(a)
```
```log
array([[[0.0320586 , 0.08714432, 0.2368828 , 0.6439142 ],
        [0.0320586 , 0.08714432, 0.2368828 , 0.6439142 ],
        [0.0320586 , 0.08714432, 0.2368828 , 0.6439142 ]],

       [[0.6439142 , 0.2368828 , 0.08714432, 0.0320586 ],
        [0.6439142 , 0.2368828 , 0.08714432, 0.0320586 ],
        [0.6439142 , 0.2368828 , 0.08714432, 0.0320586 ]]], dtype=float32)
```
and
```c++
blaze::DynamicTensor<int> t{
        {{1, 2, 3, 4}, {5, 6, 7, 8}, {9, 10, 11, 12}},
        {{-1, -2, -3, -4}, {-5, -6, -7, -8}, {-9, -10, -11, -12}}};

std::cout << blaze::softmax(t) << std::endl;
```
```log
(( 1.05575e-05  2.86983e-05    7.801e-05  0.000212053 ) (  0.00057642   0.00156687    0.0042592    0.0115777 ) (   0.0314715    0.0855484     0.232545     0.632122 ) )
((  1.4288e-06  5.25627e-07  1.93367e-07  7.11359e-08 ) ( 2.61694e-08   9.6272e-09  3.54165e-09   1.3029e-09 ) (  4.7931e-10  1.76328e-10  6.48676e-11  2.38634e-11 ) )
```

src/plugins/keras_support/categorical_crossentropy_operation.cpp
```diff
@@ -71,11 +71,39 @@ namespace phylanx { namespace execution_tree { namespace primitives
...
+
+        output.scalar() = blaze::max(blaze::min(
+            output.scalar(), hi.scalar()), lo.scalar());
+        // output is still 1 after clipping
+
+        // log of 1 is zero, so res is zero
+        auto res = target.scalar() * (-blaze::log(output.scalar()));
+        return primitive_argument_type{ blaze::sum(res) };
+        */
+        return primitive_argument_type{static_cast<double>(0.)};
```
taless474: I think we better to consider this case invalid since other other backends don't consider it.
```py
x = K.constant(1)
y = K.constant(0)
a = K.categorical_crossentropy(x,y)
K.get_value(a)
```
```pytb
ValueError: Invalid reduction dimension -1 for input with 0 dimensions. for 'Sum_2' (op: 'Sum') with input shapes: [], [] and with computed input tensors: input[1] = <-1>.
```
I think mapping 0d loss to zero is not correct because having a 0d array, we can have two classes; the loss should be different when the prediction (output) is close to the target and when it is not.

taless474:
@stevenrbrandt, the Keras signature of [categorical_crossentropy](https://keras.io/backend/#categorical_crossentropy) contains `axis` argument. The numpy backend probably missed it (as it is not tested ;) ). Is it possible for you to add that?

taless474:
@stevenrbrandt, the Keras signature of [categorical_crossentropy](https://keras.io/backend/#categorical_crossentropy) contains an `axis` argument. The numpy backend probably missed it (as it is not tested ;) ). Is it possible for you to add that, too?

stevenrbrandt:
@taless474 Can you show me a Python prototype that uses it? Does it just get passed to the sum arguments?

stevenrbrandt:
@taless474 is this correct?
{{{
def softmax(x, axis=-1):
    y = np.exp(x - np.max(x, axis, keepdims=True))
    return y / np.sum(y, axis, keepdims=True)


def cat_cross(target, output, from_logits=False, axis=-1):
    if from_logits:
        output = softmax(output, axis=axis)
    else:
        v = output.sum(axis=axis, keepdims=True)
        output /= v
    output = np.clip(output, 1e-7, 1 - 1e-7)
    rep = -target * np.log(output)
    ans = np.sum(rep, axis=axis, keepdims=False)
    return ans
}}}

src/plugins/keras_support/categorical_crossentropy_operation.cpp
```diff
@@ -0,0 +1,291 @@
...
+///////////////////////////////////////////////////////////////////////////////
+namespace phylanx { namespace execution_tree { namespace primitives
+{
+    ///////////////////////////////////////////////////////////////////////////
+    match_pattern_type const cat_cross_operation::match_data =
+    {
+        hpx::util::make_tuple("categorical_crossentropy",
+        std::vector<std::string>{
+            "categorical_crossentropy(_1,_2)",
+            "categorical_crossentropy(_1,_2,_3)"
```
hkaiser: We can now support keyword and default arguments. In this case you could replace the two patterns you defined with
```
categorical_crossentropy(_1_target, _2_output, __arg(_3_from_logits, false))
```
This would a) assign names to the arguments, and b) provide a default argument value for the last argument. As long as this primitive is invoked through PhySL you can be sure it will receive 3 arguments.

src/plugins/keras_support/categorical_crossentropy_operation.cpp
```diff
@@ -0,0 +1,291 @@
...
+
+            target (array_like) : input array
+            output (array_like) : output array
+            from_logits (boolean): boolean value
+
+        Returns:
+            The value should be the same as would be returned by the following
+            Python function:
+
+            def cat_cross(target, output, from_logits=False):
```
hkaiser: Shouldn't this be `categorical_crossentropy(...)`?

src/plugins/keras_support/categorical_crossentropy_operation.cpp
```diff
@@ -0,0 +1,291 @@
...
+            "categorical_crossentropy(_1,_2)",
+            "categorical_crossentropy(_1,_2,_3)"
+        },
+        &create_cat_cross_operation, &create_primitive<cat_cross_operation>,
+        R"(target, output, from_logits
+        Args:
+
+            target (array_like) : input array
+            output (array_like) : output array
+            from_logits (boolean): boolean value
```
hkaiser: This argument is optional, it's default value if `false`.

src/plugins/keras_support/categorical_crossentropy_operation.cpp
```diff
@@ -0,0 +1,291 @@
...
+                if from_logits:
+                    output = softmax(output)
+                else:
+                    output /= output.sum(axis=-1, keepdims=True)
+                output = np.clip(output, 1e-7, 1 - 1e-7)
+                return np.sum(target * -np.log(output), axis=-1, keepdims=False)
+        )")
+    };
+    const double clip_low = 1e-7;
+    const double clip_high = 1 - clip_low;
```
hkaiser: I'd suggest making those variables `constexpr` instead.

src/plugins/keras_support/categorical_crossentropy_operation.cpp
```diff
@@ -0,0 +1,291 @@
...
+      : primitive_component_base(std::move(operands), name, codename)
+    {}
+
+    primitive_argument_type cat_cross_operation::cat_cross0d(
+        arg_type&& target,arg_type&& output,bool from_logits) const
+    {
+        double v = 1;
+
+        if(!from_logits)
+            v = output.scalar()/output.scalar();
```
hkaiser: Isn't the result of this computation always `1.0`?

src/plugins/keras_support/categorical_crossentropy_operation.cpp
```diff
@@ -0,0 +1,291 @@
...
+    ///////////////////////////////////////////////////////////////////////////
+    using matrix_type =
+         blaze::DynamicMatrix<double>;
+
+    using vector_type =
+         blaze::DynamicVector<double>;
+
+    using tensor_type =
+         blaze::DynamicTensor<double>;
+
```
hkaiser: Please wrap all uses of Blaze Tensor types in pp guards:
```
#if defined(PHYLANX_HAVE_BLAZE_TENSOR)
…
#endif
```

src/plugins/keras_support/categorical_crossentropy_operation.cpp
```diff
@@ -0,0 +1,291 @@
...
+            output_ = blaze::softmax(output_);
+        }
+        else
+        {
+            output_ /= blaze::sum(output_);
+        }
+
+        target_ = blaze::map(target_, output_,[](double t_, double o_){
+            return -t_*std::log((std::min)(clip_high,(std::max)(clip_low,o_)));
+        });
```
hkaiser: You should be able to write this as:
```
if (target.is_ref())
{
    // we should not modify the `target`, re-allocate a new vector
    target = -target.vector() * blaze::log(
        (blaze::min)(clip_high, (blaze::max)(clip_low, output_)));
}
else
{
    // reuse memory 
    target.vector() = -target.vector() * blaze::log(
        (blaze::min)(clip_high, (blaze::max)(clip_low, output_)));
}
return primitive_argument_type(std::move(target));

```

src/plugins/keras_support/categorical_crossentropy_operation.cpp
```diff
@@ -0,0 +1,291 @@
...
+
+    ///////////////////////////////////////////////////////////////////////////
+    primitive_argument_type cat_cross_operation::cat_cross1d(
+        arg_type&& target,arg_type&& output,bool from_logits) const
+    {
+        vector_type&& output_ = output.vector();
+        vector_type&& target_ = target.vector();
+        if(from_logits)
+        {
+            output_ = blaze::softmax(output_);
```
hkaiser: This always allocates new memory. In case if `output` is not a reference (i.e. holds the actual data) we could optimize things and reuse the memory owned by `output`. Please see below as this can be achieved.

src/plugins/keras_support/categorical_crossentropy_operation.cpp
```diff
@@ -0,0 +1,291 @@
...
+            }
+        }
+
+        auto this_ = this->shared_from_this();
+        return hpx::dataflow(hpx::launch::sync,
+            hpx::util::unwrapping([this_ = std::move(this_)](
+                                      primitive_arguments_type&& args)
+                                      -> primitive_argument_type {
+                // Extract logits
+                std::int64_t from_logits =
```
hkaiser: Why do you use `std::int64_t` as the type of `from_logits`, wouldn't a simple `bool` work equally well?

taless474:
@stevenrbrandt yes, I believe this is correct.

src/plugins/keras_support/categorical_crossentropy_operation.cpp
stevenrbrandt: Not when output.scalar() is zero.

src/plugins/keras_support/categorical_crossentropy_operation.cpp
stevenrbrandt: It was a typo, but it compiled and ran so I didn't notice. :)

src/plugins/keras_support/categorical_crossentropy_operation.cpp
hkaiser: Ok, in this case, wouldn't it be more explicit to write this as:
```
double v = 1;
if (!from_logits && output.scalar() == 0)
    v = std::numeric_limits<double>::quiet_NaN();
```
?

stevenrbrandt:
This item should be done now, @hkaiser @taless474

src/plugins/keras_support/categorical_crossentropy_operation.cpp
```diff
@@ -0,0 +1,380 @@
...
+namespace phylanx { namespace execution_tree { namespace primitives
+{
+    ///////////////////////////////////////////////////////////////////////////
+    match_pattern_type const cat_cross_operation::match_data =
+    {
+        hpx::util::make_tuple("categorical_crossentropy",
+        std::vector<std::string>{
+            "categorical_crossentropy(_1_target,_2_output,"
+                "__arg(_3_from_logits,false),__arg(_4_axis,-1))"
+        },
```
hkaiser: As discussed, Phylanx does not support arguments to be used for 'out' values. That means that you cannot use the second argument to pass something back to the caller. The primitive has to be changed to return a pair (list) of results if we needed the information that would be returned through the `output` argument otherwise.

stevenrbrandt: @hkaiser I'm not using "output" as an output. Rather, I keep this name for reference to the algorithm. At the end of the function, output is passed out as a return value.

stevenrbrandt: @hkaiser I'm updating the docstring to make that clearer.

PR #908, opened 2019-03-29
Adding resize_images primitive

scheherzade:
Implementing resize_images with 'nearest' and 'bilinear' interpolation based on #857.

### DIFF
--- a/src/plugins/keras_support/resize_operation.cpp
+++ b/src/plugins/keras_support/resize_operation.cpp
@@ -1,0 +1,221 @@
+// Copyright (c) 2019 Shahrzad Shirzad
+// Copyright (c) 2018-2019 Hartmut Kaiser
+//
+// Distributed under the Boost Software License, Version 1.0. (See accompanying
+// file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+#include <phylanx/config.hpp>
+#include <phylanx/ir/node_data.hpp>
+#include <phylanx/plugins/keras_support/resize_operation.hpp>
+
+#include <hpx/include/lcos.hpp>
+#include <hpx/include/naming.hpp>
+#include <hpx/include/util.hpp>
+#include <hpx/throw_exception.hpp>
+
+#include <algorithm>
+#include <cmath>
+#include <cstddef>
+#include <cstdint>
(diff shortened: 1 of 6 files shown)


src/plugins/keras_support/resize_operation.cpp
```diff
@@ -0,0 +1,221 @@
...
+                lr = ur;
+            }
+            if (fy / width_factor == m.columns() - 1)
+            {
+                ur = ul;
+                lr = ll;
+            }
+
+            result(fx, fy) =
+                (ul + (fy % width_factor) * (ur - ul) / width_factor) +
```
hkaiser: I think `fy % width_factor` and `fx % height_factor` could be calculated once before the expression

src/plugins/keras_support/resize_operation.cpp
```diff
@@ -0,0 +1,221 @@
...
+        for (std::size_t i = 0;
+             i < m.rows() * height_factor * m.columns() * width_factor;
+             ++i)
+        {
+            auto fx = i / result.columns();
+            auto fy = i - fx * result.columns();
+            auto ul = m(fx / height_factor, fy / width_factor);
+            auto ur = m(fx / height_factor, fy / width_factor + 1);
+            auto ll = m(fx / height_factor + 1, fy / width_factor);
+            auto lr = m(fx / height_factor + 1, fy / width_factor + 1);
```
hkaiser: Are you sure what types those variables have? Wouldn't it be more explicit to specify the types you expect for those to have?

src/plugins/keras_support/resize_operation.cpp
```diff
@@ -0,0 +1,221 @@
...
+    primitive_argument_type resize_operation::nearest(ir::node_data<T>&& arg,
+        std::int64_t height_factor, std::int64_t width_factor,
+        std::string interpolation) const
+    {
+        auto m = arg.matrix();
+        blaze::DynamicMatrix<T> result(
+            m.rows() * height_factor, m.columns() * width_factor);
+
+        for (std::size_t i = 0;
+             i < m.rows() * height_factor * m.columns() * width_factor;
```
hkaiser: The upper limit for this loop could be calculated once, or simply use `result.size()`

PR #922, opened 2019-04-02
Add support for list comprehension.

### LINKED ISSUE #911: Add support for Python list comprehensions
List comprehensions are really useful for Python users. I like to suggest `@Phylanx` to support it. An example can be:
```py
import numpy as np
from phylanx import Phylanx

@Phylanx
def fx(x):
	return [[i+1] for i in x]

fx([1,2,3,4])
```
### Expected result
```pytb
[[2], [3], [4], [5]]
```

rtohid:
* Fix #911.
* Now an iteration space can be a variable.


taless474:
Thanks Rod.

PR #928, opened 2019-04-04
Throwing exceptions given invalid argument to random primitive

### LINKED ISSUE #891: Random to raise ValueError for negative probabilities
The following is an implementation of [Keras dropout](https://keras.io/backend/#dropout) which NumPy implementation is [here](https://github.com/STEllAR-GROUP/keras/blob/master/keras/backend/numpy_backend.py#L607):
```py
import numpy as np
import pytest

PhylanxSession.init(1)
	
def variable(value, dtype=None, name=None, constraint=None):
    return execution_tree.var(np.array(value, dtype))
	
def learning_phase():
	return True
	
@Phylanx
def _dropout(x, level, noise_shape, seed):
	if seed:
		set_seed(seed)
	noise = random(noise_shape if noise_shape else shape(x), ["bernoulli", 1 - level])
	return x * noise / (1-level)
	
@Phylanx
def _no_dropout(x):
	return x

def dropout_eager(x, level, noise_shape, seed):
	if learning_phase():
		return _dropout.lazy(x, level, noise_shape, seed)

	# dropout is not considered outside the learning phase
	return _no_dropout.lazy(x)

def dropout(x, level, noise_shape=None, seed=None):
	return dropout_eager(x, level, noise_shape, seed)
	
def eval(func):
    return func.eval()
	
def test_dropout(val):		
	# Test invalid use cases
	with pytest.raises(ValueError):
		z = dropout(variable(val), level=-0.5)

val = np.random.random((2, 5))
print(test_dropout(val))
```
We need to raise a ValueError for a negative `level`. In the PhySL code of:
```console
random(make_list(2,5),list("bernoulli",-.5))
```
I get

taless474:
Fixes #891

### DIFF
--- a/src/plugins/matrixops/random.cpp
+++ b/src/plugins/matrixops/random.cpp
@@ -231,130 +231,8 @@ namespace phylanx { namespace execution_tree { namespace primitives
                     HPX_FALLTHROUGH;
 
                 case 1:
-                {
-                    std::string dist = extract_string_value(args[0]);
-                    std::get<0>(result) = dist;
-
-                    if (dist == "bernoulli" || dist == "geometric" ||
-                        dist == "chi_squared" || dist == "poisson" ||
-                        dist == "exponential" || dist == "student_t")
-                    {
-                        double arg_1 = std::get<2>(result);
-
-                        if (arg_1 < 0)
-                            HPX_THROW_EXCEPTION(hpx::bad_parameter,
-                                "phylanx::execution_tree::primitives::"
-                                "extract_distribution_parameters",
-                                util::generate_error_message(
-                                    "the second argument for this distribution "
-                                    "cannot be negative",
-                                    name, codename));
-
-                        if (dist != "bernoulli" && arg_1 == 0)
-                            HPX_THROW_EXCEPTION(hpx::bad_parameter,
-                                "phylanx::execution_tree::primitives::"
-                                "extract_distribution_parameters",
-                                util::generate_error_message(
-                                    "the second argument for this distribution "
-                                    "cannot be zero",
-                                    name, codename));
-
-                        if (dist == "geometric" && arg_1 >= 1)
-                            HPX_THROW_EXCEPTION(hpx::bad_parameter,
-                                "phylanx::execution_tree::primitives::"
-                                "extract_distribution_parameters",
-                                util::generate_error_message(
-                                    "the geometric distribution requires "
-                                    "the given probability argument to be "
-                                    "in (0,1)",
-                                    name, codename));
-
-                        if (dist == "bernoulli" && arg_1 > 1)
-                            HPX_THROW_EXCEPTION(hpx::bad_parameter,
-                                "phylanx::execution_tree::primitives::"
-                                "extract_distribution_parameters",
-                                util::generate_error_message(
-                                    "the bernoulli distribution requires "
-                                    "the given probability argument to be "
-                                    "in [0,1]",
-                                    name, codename));
-                    }
-                    else
-                    {
-                        double arg_1 = std::get<2>(result);
-                        double arg_2 = std::get<3>(result);
-
-                        if (dist == "uniform_int" || dist == "uniform")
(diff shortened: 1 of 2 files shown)


src/plugins/matrixops/random.cpp
```diff
@@ -609,53 +480,387 @@ namespace phylanx { namespace execution_tree { namespace primitives
...
+    ///////////////////////////////////////////////////////////////////////////
+    PHYLANX_RANDOM_DISTRIBUTION_2(uniform_int,
+        std::uniform_int_distribution<std::int64_t>, std::int64_t,
+        std::int64_t);
+
+    std::unique_ptr<distribution> create_uniform_int(
+        distribution_parameters_type const& params, std::string const& name,
+        std::string const& codename)
+    {
+        if (std::get<2>(params) >= std::get<3>(params))
```
taless474: I think it's fine for them to be equal

src/plugins/matrixops/random.cpp
```diff
@@ -609,53 +480,387 @@ namespace phylanx { namespace execution_tree { namespace primitives
...
+
+    ///////////////////////////////////////////////////////////////////////////
+    PHYLANX_RANDOM_DISTRIBUTION_2(
+        uniform, std::uniform_real_distribution<double>, double, double);
+
+    std::unique_ptr<distribution> create_uniform(
+        distribution_parameters_type const& params, std::string const& name,
+        std::string const& codename)
+    {
+        if (std::get<2>(params) >= std::get<3>(params))
```
taless474: I think it's fine for them to be equal

src/plugins/matrixops/random.cpp
```diff
@@ -609,53 +480,387 @@ namespace phylanx { namespace execution_tree { namespace primitives
...
+
+    ///////////////////////////////////////////////////////////////////////////
+    PHYLANX_RANDOM_DISTRIBUTION_1(
+        bernoulli, std::bernoulli_distribution, double, std::uint8_t);
+
+    std::unique_ptr<distribution> create_bernoulli(
+        distribution_parameters_type const& params, std::string const& name,
+        std::string const& codename)
+    {
+        if (std::get<2>(params) <= 0 || std::get<2>(params) > 1)
```
taless474: I think Bernoulli's range is [0,1]

src/plugins/matrixops/random.cpp
```diff
@@ -609,53 +480,387 @@ namespace phylanx { namespace execution_tree { namespace primitives
...
+
+    ///////////////////////////////////////////////////////////////////////////
+    PHYLANX_RANDOM_DISTRIBUTION_1(
+        geometric, std::geometric_distribution<int>, double, double);
+
+    std::unique_ptr<distribution> create_geometric(
+        distribution_parameters_type const& params, std::string const& name,
+        std::string const& codename)
+    {
+        if (std::get<2>(params) < 0 || std::get<2>(params) >= 1)
```
taless474: I think geometric dist range is (0,1)

src/plugins/matrixops/random.cpp
hkaiser: I'm working on fixing those, thanks!

hkaiser:
@taless474 I hope everything is ok now

taless474:
I am not sure I see the final version, but here are changes I suggest:
'negative_binomial` second argument should be positive (0 is not acceptable) and its third argument should be in (0,1) (0 and 1 are not accepted either)
`geometric` range is (0,1), i.e. zero and one are not accepted.
`poisson` parameter should be strictly positive, same as `exponential` and `student_t`
`gamma`, `weibull` and `fisher_f` need both parameters to be strictly positive (the message is correct, but the condition does not prevent passing zeros)
`extreme_value` and all 3 kinds of `normal` don't accept zero for their third argument either.

taless474:
I forgot to mention `chi-squared`; it cannot accept zero.
Everything else is fine.

PR #931, opened 2019-04-05
Enable calling Python snippets in PhySL.

### LINKED ISSUE #905: Enable calling Python snippet from inside the Phylanx execution tree
[map_fn](https://keras.io/backend/#map_fn), [foldl](https://keras.io/backend/#foldl) and [foldr](https://keras.io/backend/#foldr) have two common arguments, `fn` (callable) and `elems`.
Here is an example of how `foldl` is used:
```py
import numpy as np

PhylanxSession.init(1)

def variable(value, dtype=None, name=None):
    return execution_tree.var(np.array(value, dtype=dtype))
	
def eval(func):
    return func.eval()

@Phylanx
def foldl_eager(fn, elems, initializer):
	return fold_left(fn, initializer, elems)

def foldl(fn, elems, initializer=None):
	return foldl_eager.lazy(fn, elems, initializer)
	
	
	
def test_foldl():
	x = np.random.rand(10, 3).astype(np.float32)
	kx = eval(foldl(lambda a, b: a + b, variable(x)))

	assert (3,) == kx.shape
	assert_allclose(x.sum(axis=0), kx, atol=1e-05)
	
test_foldl()
```

_Originally posted by @taless474 in https://github.com/STEllAR-GROUP/phylanx/issues/857#issuecomment-477378572_

rtohid:
Fix #905

### DIFF
--- a/tests/regressions/python/905_call_python_snippet.py
+++ b/tests/regressions/python/905_call_python_snippet.py
@@ -1,0 +1,32 @@
+#  Copyright (c) 2019 R. Tohid
+#
+#
+#  Distributed under the Boost Software License, Version 1.0. (See accompanying
+#  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+# #905: Enable calling Python snippet from inside the Phylanx execution tree
+
+import numpy as np
+from phylanx import Phylanx
+
+
+@Phylanx
+def foo(func, a, b):
+    return func(a, b)
+
+
+@Phylanx
+def phy_bar(x, y):
+    return x + y
+
+
+assert 5 == foo(phy_bar, 2, 3)
+
+
+def bar(x, y):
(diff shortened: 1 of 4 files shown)


tests/regressions/python/905_call_python_snippet.py
```diff
@@ -0,0 +1,32 @@
+#  Copyright (c) 2019 R. Tohid
+#
+#
+#  Distributed under the Boost Software License, Version 1.0. (See accompanying
+#  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+# #905: Enable calling Python snippet from inside the Phylanx execution tree
+
+import numpy as np
+from phylanx import Phylanx, PhylanxSession, execution_tree
```
hkaiser: Nitpick: PhylanxSession and execution_tree are not needed here

PR #935, opened 2019-04-08
Adding Switch operation

taless474:
This PR adds the functionality of Keras [switch](https://keras.io/backend/#switch) operations.
Working toward #857
It also adds three missed cases of broadcasting a tensor to another tensor (a tensor with only one row, one column and one column and row broadcast to another tensor).


src/plugins/keras_support/switch_operation.cpp
```diff
@@ -0,0 +1,397 @@
...
+#include <blaze_tensor/Math.h>
+#endif
+///////////////////////////////////////////////////////////////////////////////
+namespace phylanx { namespace execution_tree { namespace primitives
+{
+    ///////////////////////////////////////////////////////////////////////////
+    match_pattern_type const switch_operation::match_data =
+    {
+        hpx::util::make_tuple("switch",
+            std::vector<std::string>{"switch(_1, _2, _3)"},
```
hkaiser: You might want to write this as: `switch(_1, _2, __arg(_3_elsef, nil))` to make the third argument truly optional and force it to use a default value of `nil`.

src/plugins/keras_support/switch_operation.cpp
```diff
@@ -0,0 +1,397 @@
...
+
+                default:
+                    break;
+                }
+                HPX_THROW_EXCEPTION(hpx::bad_parameter,
+                    "switch_operation::validate_shapes",
+                    util::generate_error_message(
+                        "the condition shape is not compatible with the "
+                        "`then_expression` shape",
+                        name_, codename_));
```
hkaiser: This exception can be removed as the code falls through directly into an equivalent one.

src/plugins/keras_support/switch_operation.cpp
```diff
@@ -0,0 +1,397 @@
...
+        }
+        else if (!else_expr.is_ref())
+        {
+            e = c % t + (ones - c) % e;
+            return primitive_argument_type{std::move(else_expr)};
+        }
+
+        blaze::DynamicMatrix<double> result(rows, columns);
+
+        result = c % t + (ones - c) % e;
```
hkaiser: You might want to write this as:
```
blaze::DynamicMatrix<double> result = c % t + (ones - c) % e;
```
to avoid allocating the memory for the result twice (needs verification).

src/plugins/keras_support/switch_operation.cpp
```diff
@@ -0,0 +1,397 @@
...
+                            "the condition shape is incompatible with the "
+                            "`then_expression` and `else_expression` "
+                            "shapes",
+                            this_->name_, this_->codename_));
+
+                ir::node_data<std::uint8_t> cond = extract_boolean_value(
+                    std::move(args[0]), this_->name_, this_->codename_);
+                ir::node_data<double> then_expr = extract_numeric_value(
+                    std::move(args[1]), this_->name_, this_->codename_);
+                ir::node_data<double> else_expr = extract_numeric_value(
```
hkaiser: Is it ok for switch to always return doubles?

taless474: yes, I think it always returns doubles

PR #939, opened 2019-04-10
Enable iterating over `zip`.

### LINKED ISSUE #912: Add support for python `zip`
`zip` produces an iterable from two iterables, so we can loop over two things at the same time.
```py
import numpy as np
from phylanx import Phylanx

@Phylanx
def fx(start, size):
	indices = []
	for i, j in zip(start, size):
		indices.append([i, i+j])
	return indices
	
fx([0,1,1],[1,1,3])
```
### Expected
```pytb
[[0, 1], [1, 2], [1, 4]]
```

rtohid:
Fix #912.
Only supports `zip` as the iteration space of a `for` loop.

### DIFF
--- a/python/phylanx/ast/physl.py
+++ b/python/phylanx/ast/physl.py
@@ -13,6 +13,19 @@ import phylanx.execution_tree
 from phylanx import PhylanxSession
 
 
+def physl_zip(loop):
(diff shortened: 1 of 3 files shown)


taless474:
Thanks Rod.

PR #945, opened 2019-04-11
Create binary crossentropy primitive

stevenrbrandt:
Create binary crossentropy primitive


taless474:
Thanks a lot. I tested it in Keras and everything works. 
The only issue is that in both categorical and binary cross-entropies we also return the output value (in addition to result). So, these functions have more information than their correspondings in Keras. I prefer not to have the second output since it can be easily obtained with other functionalities of Keras, but it is only my opinion.
Thanks again.

PR #952, opened 2019-04-17
Changing the output of ctc_decode_operation to be compatible with tensorflow

scheherzade:
The tensorflow backend of keras returns a tuple where the first element is a list of size one.


hkaiser:
We need to rethink how we should handle this situation. The problem currently is that the ctc_decode primitive now returns a pair of values once evaluated, however Keras expects a pair of things that can be evaluated separately.

taless474:
Here is an example that should produce two output tensors.

PR #969, opened 2019-04-26
Adding conv2d operation

taless474:
This PR adds conv2d which imitates the functionality of Keras [conv2d](https://keras.io/backend/#conv2d).
It also modifies conv1d for the cases that the image is shorter than the filter having `same` padding.
Toward solving #857


hkaiser:
This was superseded by #1042

PR #972, opened 2019-05-04
K8 update

ct-clmsn:
added kubernetes remote deployment scripts.


python/phylanx/ast/deploy.py
```diff
@@ -0,0 +1,63 @@
+from kubernetes import client, config, utils
```
hkaiser: Please add copyright and license

python/phylanx/ast/physl.py
```diff
@@ -11,6 +11,7 @@
 import numpy as np
 import phylanx.execution_tree
 from phylanx import PhylanxSession
+from phylanx.ast.deploy import *
```
hkaiser: Please be as specific while importing as possible, e.g.
```
from phylanx.ast.deploy import deployer
```
That will help avoiding name clashing problems down the road.

python/phylanx/ast/deploy.py
ct-clmsn: @hkaiser all set.

python/phylanx/ast/physl.py
ct-clmsn: @hkaiser fixed!

ct-clmsn:
@hkaiser and now to finalize randomforest!

PR #981, opened 2019-05-10
Support lazy compilation.

### LINKED ISSUE #979: @Phylanx functions can't call other @Phylanx functions anymore
Demonstration:

```python
from phylanx import Phylanx

@Phylanx
def fx_inner():
    return 40 + 2

@Phylanx
def fx():
    return fx_inner()

fx()
```
```pytb
  File "C:\Repos\phylanx_python_scratchpad\phylanx_python_scratchpad\phylanx_python_scratchpad.py", line 11, in <module>
    fx()
  File "C:\Repos\phylanx\cmake-build-debug\python\build\lib.win-amd64-3.6\phylanx\ast\transducer.py", line 145, in __call__
    result = self.backend.call(map(self.map_decorated, args))
  File "C:\Repos\phylanx\cmake-build-debug\python\build\lib.win-amd64-3.6\phylanx\ast\physl.py", line 535, in call
    return self.lazy(args).eval()
  File "C:\Repos\phylanx\cmake-build-debug\python\build\lib.win-amd64-3.6\phylanx\ast\physl.py", line 526, in lazy
    self.wrapped_function.__name__, self.__src__)
RuntimeError: C:\Repos\phylanx_python_scratchpad\phylanx_python_scratchpad\phylanx_python_scratchpad.py(9, 11): couldn't find function 'fx_inner' in symbol table: HPX(bad_parameter)
```
Commit: ff576e03b1e7d90d4847deee10c03bf4fdd62fa1
Related to #975

rtohid:
When Phylanx is not initialized, a list of all transformed functions is
created but the compilation is postponed. Once Phylanx is initialized,
all the functions are compiled in the order they were discovered.

Fixes #979

### DIFF
--- a/python/phylanx/ast/physl.py
+++ b/python/phylanx/ast/physl.py
@@ -355,6 +355,7 @@ def check_return(ir):
         for s in ir:
             check_return(s)
 
+
 class PhySLFunction:
     def __init__(self, file_name, function_name, src):
         self.file_name = file_name
@@ -427,7 +428,6 @@ class PhySL:
             func_name = self.wrapped_function.__name__
             func = PhySLFunction(self.file_name, func_name, self.__src__)
             PhySL.functions.append(func)
-
 
     def generate_physl(self, ir):
         if len(ir) == 2 and isinstance(ir[0], str) and isinstance(
@@ -532,7 +532,7 @@ class PhySL:
 
         if not PhylanxSession.is_initialized:
             PhylanxSession.init(1)
-                    
+
             for func in PhySL.functions:
                 func.compile()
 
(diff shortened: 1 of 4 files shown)


python/phylanx/ast/physl.py
```diff
@@ -514,6 +533,9 @@ def lazy(self, args=()):
         if not PhylanxSession.is_initialized:
             PhylanxSession.init(1)
 
+            for func in PhySL.functions:
+                func.compile()
+
```
hkaiser: Will the functions be compiled also if the user explicitly calls `PhylanxSession.Init()` (possibly before any of the decorated functions are being seen)?

rtohid: Not if the user manually initializes Phylanx in between defining Phylanx functions. I just pushed a fix.

hkaiser: Thanks!

PR #994, opened 2019-06-27
Working on distributed LRA implementation

hkaiser:
- added annotation support to primitive_argument_type
- added annotation_operation primitive
- factored out common code for transpose and dot
- added distributed transpose_d and dot_d primitives
- added num_localities primitive
- added distributed LRA example


hkaiser:
@stevenrbrandt would you have the time to have a look at the failure generated by the help-check step (https://circleci.com/gh/STEllAR-GROUP/phylanx/28480), please? The help strings that the tool complains about look correct to me (see https://github.com/STEllAR-GROUP/phylanx/blob/distributed_lra/src/execution_tree/primitives/annotate_primitive.cpp#L40, for instance).

stevenrbrandt:
@hkaiser The error message is telling you that the help message doesn't have a "Returns" section (which is required). See ./src/plugins/arithmetics/cumsum.cpp (for an example) around line 40.

hkaiser:
> The error message is telling you that the help message doesn't have a "Returns" section (which is required). See ./src/plugins/arithmetics/cumsum.cpp (for an example) around line 40.

@stevenrbrandt the relevant error message does have a Returns: section, see: https://github.com/STEllAR-GROUP/phylanx/blob/distributed_lra/src/execution_tree/primitives/annotate_primitive.cpp#L48.

hkaiser:
@stevenrbrandt ping?

stevenrbrandt:
Something is very strange here... still working on it.

stevenrbrandt:
The check_help.py script has been broken since PhylanxSession was added. A number of primitives were not being validated and so there are now tons of non-conforming help strings.

However, I think what you need to get you through your current problem is the following change:
```
diff --git a/tools/check_help/check_help.py b/tools/check_help/check_help.py
index 6088cf49..12472cff 100644
--- a/tools/check_help/check_help.py
+++ b/tools/check_help/check_help.py
@@ -51,6 +51,8 @@ for p in all:

         # The the phylanx help string for the function
         h = phyhelpex(n)
+        if "No help available" in h:
+            h = phyhelpex(p)

         # Strip off leading whitespace
         g = re.search(r'^([ \t]*)Args:', h, re.MULTILINE)
```

PR #1011, opened 2019-08-02
Adding 4D support to batch_dot operation

taless474:
This PR adds 4D support to the batch_dot operation in Keras support.


taless474:
@hkaiser, I think there is a memory problem with this branch (It works on docker).  Would you please guide me on how to fix it?

hkaiser:
@taless474 CircleCI fails with what looks like the compiler running out of memory. You might want to split the file into smaller pieces to circumvent this.

hkaiser:
@taless474 Could you please rebase this branch on master. Everything should compile now.

PR #1050, opened 2019-10-10
Add Lanczos solver

NanmiaoWu:
Change get _lin_solver_map_ul_ to make it more general. Originally, it only support string type; now it also support int64_t type.

### DIFF
--- a/src/plugins/solvers/linear_solver.cpp
+++ b/src/plugins/solvers/linear_solver.cpp
@@ -1,6 +1,6 @@
 // Copyright (c) 2018 Shahrzad Shirzad
 //               2018 Patrick Diehl
-//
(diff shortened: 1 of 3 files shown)


src/plugins/solvers/linear_solver.cpp
```diff
@@ -552,7 +553,7 @@ namespace phylanx { namespace execution_tree { namespace primitives
                             "the linear_solver_operation primitive "
                             "requires "
                             "that first operand to be a matrix and "
-                            "the second "
+                            "the secotond "
```
hkaiser: This was most likely an accidental change.

NanmiaoWu: Yes. I will change it.

PR #1066, opened 2019-10-31
Cannon product

folshost:
Adding Cannon's algorithm for distributed matrix multiplication.


hkaiser:
@folshost Please rebase this PR on master. Things should go through after that.

src/execution_tree/primitives/annotate_primitive.cpp
```diff
@@ -108,7 +108,8 @@ namespace phylanx { namespace execution_tree { namespace primitives {
             annotation tmp(args);
```
hkaiser: Why do we need this copy of `args`?

src/execution_tree/primitives/annotate_primitive.cpp
```diff
@@ -118,25 +119,23 @@ namespace phylanx { namespace execution_tree { namespace primitives {
...
                 HPX_THROW_EXCEPTION(hpx::bad_parameter,
                     "annotate_primitive::annotate_d",
-                    generate_error_message("locality and/or tile annotation not found"));
+                    generate_error_message(
+                        "locality and/or tile annotation not found"));
             }
         }
+        else if (phylanx::execution_tree::extract_string_value(*args.begin()) ==
+            "tile")
+        {
```
hkaiser: Is `the first element in `args` guaranteed to refer to the `tile` annotation?

hkaiser:
@folshost Phylanx master has changed quite a bit since you branched off your work. Could you please rebase onto (or at least merge from) master to have those changes on your branch as well, please?

src/execution_tree/primitives/annotate_primitive.cpp
folshost: Good catch, it would appear that we don't.

src/execution_tree/primitives/annotate_primitive.cpp
folshost: This is in the case where there is no args wrapper, and only the tile annotation is supplied.

PR #1070, opened 2019-11-06
With this test, binary_crossentropy passes when compared to the newest Keras

stevenrbrandt:
The way the test for binary_crossentropy() works is as follows.

First, we create a pure python function that behaves the same as binary_crossentropy() and we test that it gives the same answers as Keras's binary_crossentropy().

Second, we test that the pure python function matches what the Phylanx plugin does.

That way, we don't need Keras to test the plugin.

Apparently, however, the Keras implementation changed because the test stopped passing on the system I built with Keras installed. This pull request updates the pure python and the C++ codes to make them agree with Keras.


hkaiser:
@stevenrbrandt Could you please rebase this on master. After that we can go ahead and merge it.

stevenrbrandt:
@hkaiser @taless474 CircleCI is telling me all tests have passed. Can one of you approve? Thanks.

PR #1074, opened 2019-11-10
Storing generated PhySL string and AST in database

hkaiser:
- this speeds up repeated script startup by about 10-15%


ct-clmsn:
Does this store into a directory like '$HOME/.physl' ? It's a great idea.

hkaiser:
> Does this store into a directory like '$HOME/.physl' ? It's a great idea.

@ct-clmsn This stores the database in a directory named `__physlcache__` next to the script that imported Phylanx (for now). We can change this however, if needed.

PR #1086, opened 2019-12-27
Add the NO_EXTRAS flag for the pybind11_add_module call

### LINKED ISSUE #1085: Phylanx Release build fails to link
When building Phylanx with gcc 7.3 (Release build), pybind11 is adding additional flags that break linking.  This is the error:  (see http://omega.nic.uoregon.edu:8020/#/builders/2/builds/57 for example)

```
[ 60%] Linking CXX shared module ../../bin/_phylanx.cpython-36m-x86_64-linux-gnu.so
/bin/ld: /tmp/ccvkdP3c.ltrans12.ltrans.o: relocation R_X86_64_PC32 against undefined symbol `_ZNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEED1Ev' can not be used when making a shared object; recompile with -fPIC
/bin/ld: final link failed: Bad value
collect2: error: ld returned 1 exit status
make[2]: *** [bin/_phylanx.cpython-36m-x86_64-linux-gnu.so] Error 1
make[1]: *** [python/src/CMakeFiles/phylanx_py.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
```

A workaround is to use the `NO_EXTRAS` setting for the `pybind11_add_module()` call in the `python/src/CMakeLists.txt` file.  Not sure if this is a "fix" though.

khuck:
Fixes #1085.  The NO_EXTRAS flag prevents the addition of `-flto` which is unfortunate, but more importantly prevents the stripping of binaries which causes the `std::basic_string::~basic_string` destructor to be undefined (for Release and RelWithDebInfo builds).

### DIFF
--- a/python/src/CMakeLists.txt
+++ b/python/src/CMakeLists.txt
@@ -59,7 +59,7 @@ if(APPLE) # OSX requires this to be a shared library, not a module (default)
       ${phylanx_py_SOURCES}
       ${phylanx_py_HEADERS})
 else(APPLE)
-  pybind11_add_module(phylanx_py
+  pybind11_add_module(phylanx_py NO_EXTRAS
     ${phylanx_py_SOURCES}
     ${phylanx_py_HEADERS}
     ${phylanx_python_SOURCES})


khuck:
@hkaiser can I be authorized to merge this PR?  Thanks!

hkaiser:
I will merge it. Would you please try using `SHARED` instead of `NO_EXTRAS` first?

khuck:
> I will merge it. Would you please try using `SHARED` instead of `NO_EXTRAS` first?

will do

khuck:
> I will merge it. Would you please try using `SHARED` instead of `NO_EXTRAS` first?

SHARED created a bunch of different linking errors when linking `bin/_phylanx.cpython-37m-x86_64-linux-gnu.so`.

khuck:
@hkaiser also, adding -flto to the link didn't change the shared object size, so I doubt that it was doing anything to begin with.

hkaiser:
Thanks @khuck.

PR #1095, opened 2020-01-31
Add missing PLUGIN statement to common plugin

parsa:


hkaiser:
@parsa Are you sure this is the actual issue? Why does it work without that on CircleCI? Also, this shared library is not a PLUGIN, but just a 'normal' library the primitives are linked against...

PR #1103, opened 2020-02-05
Making sure primitive names are consistent for non-distributed runs

### LINKED ISSUE #1102: Unify primitive names
Primitive names for non-distributed runs currently are different from the names generated for distributed executions. This is to remind us to unify those.

hkaiser:
- flyby: tabs-->spaces

Fixes #1102


hkaiser:
@parsa would you be able to have another look at the failing build step? I think that the Python environment is not quite right yet.

parsa:
The latest commit ought to do it. 

Also, I see that our CI is set to build Phylanx with C++17. Does that mean we have dropped C++14 support for Phylanx?

hkaiser:
> Also, I see that our CI is set to build Phylanx with C++17. Does that mean we have dropped C++14 support for Phylanx?

HPX in the base image is built using C++17, so we need to do the same here.

PR #1105, opened 2020-02-11
Fix #1104.

### LINKED ISSUE #1104: No arg lambda's don't work
```
from phylanx import Phylanx

@Phylanx
def foo():
    f = lambda : print("Hello")

foo()
```
creates
```
--------------------------------------------------------------------------
[[38797,1],0]: A high-performance Open MPI point-to-point messaging module
was unable to find any relevant network interfaces:

Module: OpenFabrics (openib)
  Host: 29dd8de5e0ff

Another transport will be used instead, although this may result in
lower performance.
--------------------------------------------------------------------------
PAPI lmsensors component found...
PAPI RAPL component found...
No RAPL events found.
Can't open fd for cpu0: Permission denied.
  File "/home/jovyan/.local/lib/python3.6/site-packages/phylanx-0.0.1-py3.6-linux-x86_64.egg/phylanx/ast/physl.py", line 374, in _compile_or_load
    physl_db = db(_name_of_importing_file)
  File "/home/jovyan/.local/lib/python3.6/site-packages/phylanx-0.0.1-py3.6-linux-x86_64.egg/phylanx/ast/physl_db.py", line 95, in __init__
    self.dbname, self.db, created = self.ensure_db(name, version)
  File "/home/jovyan/.local/lib/python3.6/site-packages/phylanx-0.0.1-py3.6-linux-x86_64.egg/phylanx/ast/physl_db.py", line 34, in ensure_db
    os.makedirs(dbdir)
  File "/usr/local/lib/python3.6/os.py", line 220, in makedirs
    mkdir(name, mode)
PermissionError: [Errno 13] Permission denied: '/__physlcache__'

During handling of the above exception, another exception occurred:

  File "x0.py", line 7, in <module>
    foo()
  File "/home/jovyan/.local/lib/python3.6/site-packages/phylanx-0.0.1-py3.6-linux-x86_64.egg/phylanx/ast/transducer.py", line 162, in __call__
    result = self.backend.call(*mapped_args, **mapped_kwargs)
  File "/home/jovyan/.local/lib/python3.6/site-packages/phylanx-0.0.1-py3.6-linux-x86_64.egg/phylanx/ast/physl.py", line 548, in call
    self._ensure_global_state()
  File "/home/jovyan/.local/lib/python3.6/site-packages/phylanx-0.0.1-py3.6-linux-x86_64.egg/phylanx/ast/physl.py", line 421, in _ensure_global_state
    PhySLFunction.compile()
  File "/home/jovyan/.local/lib/python3.6/site-packages/phylanx-0.0.1-py3.6-linux-x86_64.egg/phylanx/ast/physl.py", line 348, in compile
    func.compile_function()
  File "/home/jovyan/.local/lib/python3.6/site-packages/phylanx-0.0.1-py3.6-linux-x86_64.egg/phylanx/ast/physl.py", line 342, in compile_function
    self.physl._ensure_is_compiled()
  File "/home/jovyan/.local/lib/python3.6/site-packages/phylanx-0.0.1-py3.6-linux-x86_64.egg/phylanx/ast/physl.py", line 434, in _ensure_is_compiled
    self._compile_or_load()
  File "/home/jovyan/.local/lib/python3.6/site-packages/phylanx-0.0.1-py3.6-linux-x86_64.egg/phylanx/ast/physl.py", line 400, in _compile_or_load
    self.__ast__ = phylanx.ast.generate_ast(self.__src__)
RuntimeError: Error! Expecting ")" line 1:
define$4$0(foo$4$0, lambda$4$0(define$5$4(f$5$4, lambda$5$8(, cout$5$17("Hello")))))
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
Error! Expecting ")" line 1:
define$4$0(foo$4$0, lambda$4$0(define$5$4(f$5$4, lambda$5$8(, cout$5$17("Hello")))))
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
Error! Expecting ")" line 1:
define$4$0(foo$4$0, lambda$4$0(define$5$4(f$5$4, lambda$5$8(, cout$5$17("Hello")))))
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
Error! Expecting ")" line 1:
define$4$0(foo$4$0, lambda$4$0(define$5$4(f$5$4, lambda$5$8(, cout$5$17("Hello")))))
~~~~~~~~~~~~~~~~~~^
Error! Incomplete parse: line 1:
define$4$0(foo$4$0, lambda$4$0(define$5$4(f$5$4, lambda$5$8(, cout$5$17("Hello")))))
^
```

rtohid:
- If `lambda` has no argument, generate PhySL accordingly- ignore the `arg` entry.

### DIFF
--- a/tests/regressions/python/1104_no_arg_lambda.py
+++ b/tests/regressions/python/1104_no_arg_lambda.py
@@ -1,0 +1,17 @@
+#  Copyright (c) 2020 R. Tohid
+#  Copyright (c) 2020 Steven R. Brandt
+#
+#  Distributed under the Boost Software License, Version 1.0. (See accompanying
+#  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+# #1104 No arg lambdas don't work
+
+from phylanx import Phylanx
+
+@Phylanx
+def foo():
+    f = lambda : print("Hello")
+
+    return True
+
+assert True == foo()
--- a/tests/regressions/python/1104_no_arg_lambda.py
+++ b/tests/regressions/python/1104_no_arg_lambda.py
@@ -8,10 +8,10 @@
 
 from phylanx import Phylanx
 
+
 @Phylanx
 def foo():
-    f = lambda : print("Hello")
+    return lambda: True
 
-    return True
 
-assert True == foo()
+assert foo()()
--- a/tests/regressions/python/1104_no_arg_lambda.py
+++ b/tests/regressions/python/1104_no_arg_lambda.py
@@ -11,7 +11,8 @@ from phylanx import Phylanx
 
 @Phylanx
 def foo():
-    return lambda: True
+    f = lambda: True    # noqa E731
+    return f
 
 
 assert foo()()
--- a/python/phylanx/ast/physl.py
+++ b/python/phylanx/ast/physl.py
@@ -1083,8 +1083,10 @@ class PhySL:
         symbol = get_symbol_info(node, 'lambda')
         args = self._apply_rule(node.args)
         body = self._block(node.body)
-
-        return [symbol, (args, body)]
+        if args:
+            return [symbol, (args, body)]
+        else:
+            return [symbol, (body)]
 
     def _List(self, node):
         """class List(elts, ctx)"""
--- a/python/src/bindings/execution_tree.cpp
+++ b/python/src/bindings/execution_tree.cpp
@@ -238,6 +238,17 @@ void phylanx::bindings::bind_execution_tree(pybind11::module m)
                     });
             },
             "evaluate execution tree")
+        .def("__call__", [](phylanx::execution_tree::primitive const& p)
+            {
+                pybind11::gil_scoped_release release;       // release GIL
+                return hpx::threads::run_as_hpx_thread(
+                    [&]() {
+                        using namespace phylanx::execution_tree;
+                        return value_operand(primitive_argument_type{p},
+                            primitive_argument_type{}).get();
+                    });
+            },
+            "evaluate execution tree")
         .def("assign", [](phylanx::execution_tree::primitive p, double d)
             {
                 pybind11::gil_scoped_release release;       // release GIL
--- a/tests/regressions/python/CMakeLists.txt
+++ b/tests/regressions/python/CMakeLists.txt
@@ -50,6 +50,7 @@ set(tests
     1052_variable_set_dtype
     1054_variable_shape
     1056_client_base
+    1104_no_arg_lambda
     array_len_494
     array_shape_486
     array_subscript_403


tests/regressions/python/1104_no_arg_lambda.py
```diff
@@ -0,0 +1,17 @@
...
+
+from phylanx import Phylanx
+
+@Phylanx
+def foo():
+    f = lambda : print("Hello")
+
+    return True
+
+assert True == foo()
```
hkaiser: This misses an EOL character.

hkaiser:
@parsa could you please have a look what happened to our build system on CircleCI for Phylanx? Has anything changed in the base image? @msimberg?

msimberg:
@hkaiser it looks like you would need the `python3.8-dev` package in the image. It seems like the default `python3` is `python3.7` and it has the associated headers also installed from `python3.7-dev`. For some reasone `python3.8` is also installed but not the headers. Is it possible that there's some custom handling of the Python paths in Phylanx or pybind11? Ideal would be if it would just find `python3.7` instead, but we can of course just install `python3.8-dev` in the image.

hkaiser:
> @hkaiser it looks like you would need the python3.8-dev package in the image. It seems like the default python3 is python3.7 and it has the associated headers also installed from python3.7-dev. For some reasone python3.8 is also installed but not the headers. Is it possible that there's some custom handling of the Python paths in Phylanx or pybind11? Ideal would be if it would just find python3.7 instead, but we can of course just install python3.8-dev in the image.

I think we detect Python using `find_package(Python)` (https://cmake.org/cmake/help/v3.16/module/FindPython.html). So it should be possible to explicitly specify the path and/or version of Python to use.

parsa:
Bita reported this issue to me on Monday. The issue is that both Python 3.7 and 3.8 are installed in the new HPX base image. The only workaround I have come up with is installing `libpython3.8-dev` but this is not a sustainable solution as the Python versions installed in the base image would likely keep increasing in the future. We can install `libpython3.*-dev` instead, but I am not convince if we should do that. If possible we should do something that does not install multiple Python versions in the base image.

msimberg:
Indeed. I tried to look for what caused `python3.8` to be installed but didn't come up with anything useful. @parsa would you know how to best find out what's caused it to be installed?

parsa:
@msimberg Still investigating. I am currently in the process of running every `apt` command we have in the base image until Phylanx until I find the culprit.

msimberg:
> @msimberg Still investigating. I am currently in the process of running every `apt` command we have in the base image to Phylanx. This should probably expose the culprit.

Ugh, thanks a lot!

parsa:
Update: Just doing `apt install -y python3 python3-pip` alone will install Python 3.7  and 3.8. It's something `ubuntu:focal` itself does.

hkaiser:
> Update: Just doing apt install -y python3 python3-pip alone will install Python 3.7  and 3.8. It's something ubuntu:focal itself does.

So let's install `python3.8-dev` and be done with it. 

@parsa thanks for investigating!

parsa:
@hkaiser Can we switch to `ubuntu:eoan` (rolling) instead of `focal`(devel)? This issue does not seem to happen with `eoan` and it is probably a bit more stable than the devel Ubuntu?

hkaiser:
> @hkaiser Can we switch to ubuntu:eoan (rolling) instead of focal(devel)? This issue does not seem to happen with eoan and it is probably a bit more stable than the devel Ubuntu.

Fine by me. @msimberg?

rtohid:
I created a PR <https://github.com/STEllAR-GROUP/docker_build_env/pull/20>
on, it should resolve the issue. If it gets merged, we can remove python
installation on Phylanx Dockerfile.

parsa:
@rtohid with this PR we would still have both 3.7 and 3.8 installed, with headers for only 3.7 installed.

rtohid:
@parsa  did you remove Python installation commands from Phylanxs' Dockerfile?

If the issue is still persists, we probably need to report it to Ubunutu folks.

msimberg:
It seems like it's actually `python3-pip` that ends up installing `python3.8` (there may be others; I hope not). I think we could manage with the Python packages in the Ubuntu repos now. I used `pip` to install them originally because we were on Ubuntu LTS, which lead to quite old Python packages. @parsa, @rtohid does that sound good? This is all assuming nothing else ends up installing `python3.8` of course...

parsa:
@msimberg We use Pip to install Phylanx dependencies like `flake8` and `astpretty` that I doubt are Ubuntu packages. I mean, we can also use the pip installer instead of the Ubuntu package to work around this issue, but this is an Ubuntu bug any way and will probably be fixed at some point. It only shows in the `focal` builds. `rolling` and older images don't have this issue so if we switch to say `eoan` it should work for us because the software is new enough for us and is also a bit more stable.

msimberg:
@parsa, ah, that's a very good point. It seems like I've made an off by one error... I thought `focal` was 19.10. I didn't actually mean to use an unstable release of Ubuntu. `eoan` is fine for us since it has clang 9 and CMake 3.13. Can you update the PR?

msimberg:
Does it look like builds are back to normal here after the docker image was updated?

parsa:
@msimberg That particular issue with multiple Python versions is indeed fixed, and so are several other unrelated ones that @hkaiser has fixed. We are still not green though. [runs/453724033](https://github.com/STEllAR-GROUP/phylanx/runs/453724033) is the latest CI test we have at the moment.

msimberg:
@parsa at least for the `leaf` compilation issue you should most likely replace that with `filename` as it's been deprecated (search for `leaf` over here: https://www.boost.org/doc/libs/1_71_0/libs/filesystem/doc/reference.html). I can't help you with the Python problem...

hkaiser:
> @parsa at least for the leaf compilation issue you should most likely replace that with filename as it's been deprecated (search for leaf over here: https://www.boost.org/doc/libs/1_71_0/libs/filesystem/doc/reference.html). I can't help you with the Python problem...

Yes, I missed that in my fixes that are part of #1103, should be ok now. I'm focusing on making that go through first and will rebase all the other pending PRs afterwards.

PR #1108, opened 2020-02-12
Fix string output

stevenrbrandt:
If quotes are embedded in a string, they need to be formatted with escapes intact in the PhySL source. This PR fixes that issue.

### DIFF
--- a/python/phylanx/ast/physl.py
+++ b/python/phylanx/ast/physl.py
@@ -1234,7 +1234,7 @@ class PhySL:
     def _Str(self, node):
         """class Str(s)"""
 
-        return '"' + node.s + '"'
+        return '"' + re.sub(r'(["|\\])',r'\\\1',node.s) + '"'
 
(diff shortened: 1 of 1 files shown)


python/phylanx/ast/physl.py
```diff
@@ -1234,7 +1234,7 @@ def _Slice(self, node):
     def _Str(self, node):
         """class Str(s)"""
 
-        return '"' + node.s + '"'
+        return '"' + re.sub(r'(["|\\])',r'\\\1',node.s) + '"'
```
rtohid: Flake8 will fail as a couple of spaces are missing. In case no more changes are needed, here is the formatted version:
```
 return '"' + re.sub(r'(["|\\])', r'\\\1', node.s) + '"'
```

stevenrbrandt:
@hkaiser you will need to re-review as pushing the whitespace fix to make flake8 happy "dismissed" your review. :(

PR #1112, opened 2020-02-18
Use HPX targets

msimberg:
These are changes intended to go with https://github.com/STEllAR-GROUP/hpx/pull/4246. It contains two commits:

- The first (`phylanx_handle_component_dependencies`) is the absolute minimum required once the target PR is merged. Alternatively you can just forward to `hpx_handle_component_dependencies`.
- The second contains changes I would recommend you do in any case such as:
    - Prefer `DEPENDENCIES` over `COMPONENT_DEPENDENCIES`
    - Use `HPX::hpx` instead `HPX_LIBRARIES`, and `HPX::component` to create components

On top of this you may want to clean up things like `phylanx_setup_target` etc. to use the new targets (e.g. `HPX::component`, similarly to how `hpx_setup_target` has changed) but I'll leave that up to you. I'd like to deprecate the macros but we won't do that very soon. It would be very helpful if you try to move away from the macros as there's bound to be problems with the targets.

I've managed to build some tests correctly in Phylanx (I haven't built all of them) both with the first and the second commit. Feel free to do whatever you want with this PR.


hkaiser:
Wow, many thanks for starting this!

msimberg:
I've just merged https://github.com/STEllAR-GROUP/hpx/pull/4246, and the docker image should be updated later today. This PR probably needs to be retriggered to run with the new image. Please let me know if I can help on this side.

hkaiser:
@msimberg I started to experiment with this and ran into the problem, that the Debug configuration of Phylanx links against both, the Debug and the Release versions of the HPX libraries (hpx.lib and hpxd.lib). Would you know from the top of your head what might be causing this? How can I debug this?

msimberg:
#1140 superceded this, closing.

msimberg:
Was this opened on purpose?

hkaiser:
> Was this opened on purpose?

Yes, just a reminder for us to go back a look more carefully at Phylanx' build system. I'm not 100% sure we have adjusted everything properly to HPX targets yet.

msimberg:
I'm cleaning up old PRs. Feel free to open an actual issue if this is still needed.

PR #1117, opened 2020-02-24
Distributing

stevenrbrandt:
This pull request makes the find_here() and find_all_localitiies() primitives available to Python / Phylanx.


hkaiser:
As it turns out, we already have two primitives: `locality()` and `num_localities()` (see: https://github.com/STEllAR-GROUP/phylanx/blob/master/src/execution_tree/primitives/debug_output.cpp#L81-L145) We should rename `locality()` to `find_here()`, obviously, and just add your `find_all_localities()`.

hkaiser:
@stevenrbrandt there are probably tests relying on `locality()`. Those need fixing as well.

hkaiser:
@stevenrbrandt please rebase onto master. That should make the tests compile again.

stevenrbrandt:
@hkaiser all checks are passed, waiting for a review.

PR #1126, opened 2020-03-03
Adding distributed constant

taless474:
This PR adds distributed constant for vector and matrices as `constant_d`.
As some functionalities are common with `constant`, they are moved to the common library.
Three `tiling_type`s are defined for `constant_d`: `sym`, `row` and `column`. `sym` tiles the array in a close to the symmetric way. `row` and `column` tiling types are tile a matrix only on rows and columns respectively.


src/execution_tree/primitives/base_primitive.cpp
```diff
@@ -2035,6 +2035,110 @@ namespace phylanx { namespace execution_tree
                 name, codename));
     }
 
+    ///////////////////////////////////////////////////////////////////////////
+    std::int64_t extract_scalar_nonneg_integer_value_strict(
+        primitive_argument_type const& val, std::string const& name,
+        std::string const& codename)
+    {
```
hkaiser: Would it be easier to implement this in terms of `extract_scalar_integer_value_strict` and do the non-neg check if that one succeeded?

PR #1127, opened 2020-03-06
Adding missing PHYLANX_COMMON_EXPORT

hkaiser:


taless474:
I compiled on the docker and forgot about it. Thanks a lot.

Issue #1125, opened 2020-03-03
`full` does not derive its dtype from its given value


taless474:
`full`, `ones` and `zeros` are created by Phylanx `constant` primitive. While NumPy [ones](https://docs.scipy.org/doc/numpy/reference/generated/numpy.ones.html) and [zeros](https://docs.scipy.org/doc/numpy/reference/generated/numpy.zeros.html) have a default type of `double`, [full](https://docs.scipy.org/doc/numpy/reference/generated/numpy.full.html) always derived its dtype from the given value.
Phylanx `full` however, also uses `double` if an explicit dtype is not given.
```python
import numpy as np
from phylanx import Phylanx

@Phylanx
def fx_1(dim):
	return np.ones(dim)
	
@Phylanx
def fx_2(dim, n):
	return np.full(dim, n)
	
	
n = 2
dim = (3,2)

assert(fx_1(dim) == np.ones(dim)).all()
assert(fx_1(dim).dtype == np.ones(dim).dtype)

assert(fx_2(dim, n) == np.full(dim, n)).all()
assert(fx_2(dim, n).dtype == np.full(dim, n).dtype)
```
The last assertion fails.
```pytb
  File "test19.py", line 20, in <module>
    assert(fx_2(dim, n).dtype == np.full(dim, n).dtype)
AssertionError
```

taless474:
Thanks for fixing this is #1128

PR #1128, opened 2020-03-06
Making sure the full() primitive extracts default dtype from value argument

hkaiser:
Fixies #1125

### DIFF
--- a/tests/unit/python/execution_tree/primitives/numpy_dtype.py
+++ b/tests/unit/python/execution_tree/primitives/numpy_dtype.py
@@ -113,6 +113,14 @@ assert (test_full() == np.full((3, 3), 2, dtype=int)).all()
 
 
 @Phylanx
+def test_full_nodtype():
+    return np.full((3, 3), 2.0)
+
+
+assert (test_full_nodtype() == np.full((3, 3), 2.0)).all()
+
+
+@Phylanx
 def test_full_like():
     a = np.zeros((3, 3), dtype='int')
     return np.full_like(a, 42, dtype='int')
--- a/tests/unit/python/execution_tree/primitives/numpy_dtype.py
+++ b/tests/unit/python/execution_tree/primitives/numpy_dtype.py
@@ -109,7 +109,9 @@ def test_full():
     return np.full((3, 3), 2, dtype='int')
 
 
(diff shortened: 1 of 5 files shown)


tests/unit/python/execution_tree/primitives/numpy_dtype.py
```diff
@@ -112,6 +112,14 @@ def test_full():
 assert (test_full() == np.full((3, 3), 2, dtype=int)).all()
 
 
+@Phylanx
+def test_full_nodtype():
+    return np.full((3, 3), 2.0)
+
+
+assert (test_full_nodtype() == np.full((3, 3), 2.0)).all()
```
taless474: I think this test only compare values (not dtypes)

taless474:
I am sorry, I was wrong about `full_like`. It gets its dtype from the given array, not the given value.
```python
import numpy as np
from phylanx import Phylanx

@Phylanx
def fx_1(dim):
	return np.ones(dim)
	
@Phylanx
def fx_2(dim, n):
	return np.full(dim, n)
	
@Phylanx
def fx_3(a, n):
	return np.full_like(a, n)
	
n = np.array(2,dtype="int64")
dim = (3,2)
a = np.array([[1,2],[3,4],[5,6]], dtype="float64")


assert(fx_1(dim) == np.ones(dim)).all()
assert(fx_1(dim).dtype == np.ones(dim).dtype)        # float64 and float64

assert(fx_2(dim, n) == np.full(dim, n)).all()
assert(fx_2(dim, n).dtype == np.full(dim, n).dtype)  # int64 and int64

assert(fx_3(a, n) == np.full_like(a, n)).all()
assert(fx_3(a, n).dtype == np.full_like(a, n).dtype) # int64 and float64
```
```pytb
  File "test19.py", line 28, in <module>
    assert(fx_3(a, n).dtype == np.full_like(a, n).dtype) # int64 and float64
AssertionError
```

tests/unit/python/execution_tree/primitives/numpy_dtype.py
hkaiser: Fair point, will fix.

taless474:
It works perfectly, thanks.

PR #1131, opened 2020-03-10
Fix macro processing issues

taless474:
This PR moves hpx_init included libraries out of namespaces.


hkaiser:
How did this ever work before? **puzzled**

Issue #1129, opened 2020-03-09
Annotation comparison does not compare the names


taless474:
The following test
```c++
#include <phylanx/phylanx.hpp>

#include <hpx/hpx_main.hpp>
#include <hpx/testing.hpp>

#include <string>

phylanx::execution_tree::primitive_argument_type compile_and_run(
    std::string const& name, std::string const& codestr)
{
    phylanx::execution_tree::compiler::function_list snippets;
    phylanx::execution_tree::compiler::environment env =
        phylanx::execution_tree::compiler::default_environment();

    auto const& code =
        phylanx::execution_tree::compile(name, codestr, snippets, env);
    return code.run().arg_;
}
void test_annotation_equality_1()
{
    std::string annotation_0 = R"(
            annotate_d([[91, 91]], "test2d2d_4_1/1",
                list("tile", list("columns", 0, 2), list("rows", 0, 1)))
        )";
    std::string annotation_1 = R"(
            annotate_d([[91, 91]], "something",
                list("tile", list("rows", 0, 1), list("columns", 0, 2)))
        )";

    HPX_TEST_EQ(compile_and_run("annotation_0", annotation_0),
        compile_and_run("annotation_1", annotation_1));
}
int main(int argc, char* argv[])
{
    test_annotation_equality_1();

    return hpx::util::report_errors();
}
```
passes on one locality. Also in [constant_6_loc tests](https://github.com/STEllAR-GROUP/phylanx/blob/master/tests/unit/plugins/dist_matrixops/dist_constant_6_loc.cpp#L48) the name should be `constant_matrix_...` instead of `constant_vector_...`, but those tests also pass.

taless474:
Fixed by #1133.

PR #1133, opened 2020-03-13
Make distributed span indices compatible with phylanx conventions

taless474:
This PR changes the starting dimension of tiling spans in a way that they are compatible with node data conventions. For example, in a 3d array, the first dimension is the page dimension, the second one is the row dimension and then we have the column dimension (pages x rows x columns).
The distributed version of `dot` and `cannon_product` uses `get_span` directly, so these primitives are modified.
Also, tiling_information dimensions are adapted for the PHYLANX_MAX_DIMENSIONS, 4.
To comply with node_data, the name of `dimension()` member function of the `localities_information` is changed to `num_dimensions()` (It returns a `size_t` that is the number of dimensions).

### DIFF
--- a/phylanx/execution_tree/tiling_annotations.hpp
+++ b/phylanx/execution_tree/tiling_annotations.hpp
@@ -67,7 +67,8 @@ namespace phylanx { namespace execution_tree
     {
         static char const* get_span_name(std::size_t i)
         {
-            static const char* const names[] = {"columns", "rows", "pages"};
+            static const char* const names[] = {
+                "columns", "rows", "pages", "quats"};
             return names[i];
         }
 
@@ -81,10 +82,10 @@ namespace phylanx { namespace execution_tree
             std::string const& name, std::string const& codename) const
         {
             annotation ann("tile");
-            for (std::size_t i = 0; i != spans_.size(); ++i)
+            for (std::size_t i = spans_.size(); i != 0; --i)
             {
-                ann.add_annotation(get_span_name(i),
-                    ir::range(spans_[i].start_, spans_[i].stop_), name,
+                ann.add_annotation(get_span_name(i - 1),
+                    ir::range(spans_[i - 1].start_, spans_[i - 1].stop_), name,
                     codename);
             }
             return ann;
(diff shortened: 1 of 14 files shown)


phylanx/execution_tree/tiling_annotations.hpp
```diff
@@ -67,7 +67,8 @@ namespace phylanx { namespace execution_tree
     {
         static char const* get_span_name(std::size_t i)
         {
-            static const char* const names[] = {"columns", "rows", "pages"};
+            static const char* const names[] = {
+                "columns", "rows", "pages", "quats"};
```
hkaiser: Should this be `"quads"` instead?

phylanx/execution_tree/tiling_annotations.hpp
```diff
@@ -179,7 +180,7 @@ namespace phylanx { namespace execution_tree
         PHYLANX_EXPORT void transpose(
             std::int64_t const* data, std::size_t count);
 
-        tiling_span spans_[3];    // column, row and page spans
+        tiling_span spans_[3];    // page, row and column spans
```
hkaiser: If we support `quads` now, should we increase the size of this array?

hkaiser: I take that back, this type is for 3D data only anyways...

phylanx/execution_tree/tiling_annotations.hpp
taless474: I think we named it `quats` as it was the first dimension of quaternions ([here](https://github.com/STEllAR-GROUP/phylanx/blob/master/phylanx/util/serialization/blaze.hpp#L196)).

taless474:
Please let me fix #1137 problem here. I have not changed anything in `constant` and `random` since the tests were passed.

hkaiser:
> Please let me fix #1137 problem here. I have not changed anything in constant and random since the tests were passed.

Please let me know once this is ready to go.

taless474:
This PR would fix #1129 and #1137.
It assumes that for a column vector spans size is 1  (e.g. spans =[[start, stop]]). 
For a row vector spans has a unit number of dimensions, but it is represented as spans = [[0,0],[start,stop]].

PR #1142, opened 2020-03-29
Making sure random_d produces correctly sized 2d results

### LINKED ISSUE #1139: Incorrect results of extract_numeric_value_dimensions for `random` matrices
I ran the following code on two localities:
(I use `unify_indices` branch, however, I got the same result on the master before)
```c++
#include <phylanx/phylanx.hpp>

#include <hpx/hpx_init.hpp>
#include <hpx/include/iostreams.hpp>
#include <hpx/include/lcos.hpp>
#include <hpx/testing.hpp>

#include <cstdint>
#include <string>
#include <utility>
#include <vector>

#include<iostream>

void execution_time(std::string const& name, std::string const& codestr)
{
    phylanx::execution_tree::compiler::function_list snippets;
    phylanx::execution_tree::compiler::environment env =
        phylanx::execution_tree::compiler::default_environment();

    auto const& compiled_code =
        phylanx::execution_tree::compile(name, codestr, snippets, env);

    hpx::cout << "Locality: " << hpx::get_locality_id()
              << "  Starting execution for: " << name << hpx::endl;
    std::uint64_t t = hpx::util::high_resolution_clock::now();
    phylanx::execution_tree::primitive_argument_type a = compiled_code.run().arg_;
    t = hpx::util::high_resolution_clock::now() - t;

    std::cout << *(a.annotation()) << "\n\n\n";
    hpx::cout << "       " << (t / 1e6) << " ms.\n" << hpx::flush;
}

////////////////////////////////////////////////////////////////////////////////
void test__0()
{
    if (hpx::get_locality_id() == 0)
    {
        execution_time("perftest_0", R"(
            dot_d(
                constant_d(13., list(100, 100), 0, 2, "const_1"),
                constant_d(13., list(100, 100), 0, 2, "const_2")
            )
        )");
    }
    else if (hpx::get_locality_id() == 1)
    {
        execution_time("perftest_0", R"(
            dot_d(
                constant_d(13., list(100, 100), 1, 2, "const_1"),
                constant_d(13., list(100, 100), 1, 2, "const_2")
            )
        )");
    }
}

void test__1()
{
    if (hpx::get_locality_id() == 0)
    {
        execution_time("perftest_1", R"(
            dot_d(
                random_d(list(100, 100), 0, 2, "rand_1"),
                random_d(list(100, 100), 0, 2, "rand_2")
            )
        )");
    }
    else if (hpx::get_locality_id() == 1)
    {
        execution_time("perftest_1", R"(
            dot_d(
                random_d(list(100, 100), 1, 2, "rand_1"),
                random_d(list(100, 100), 1, 2, "rand_2")
            )
        )");
    }
}

////////////////////////////////////////////////////////////////////////////////
int hpx_main(int argc, char* argv[])
{
    test__0();
    test__1();

    hpx::finalize();
    return hpx::util::report_errors();
}

int main(int argc, char* argv[])
{
    std::vector<std::string> cfg = {"hpx.run_hpx_main!=1"};

    return hpx::init(argc, argv, cfg);
}
```
test__0 passes while test__1 fails throwing this exception:
```console
inconsistent dimensionalities between data and tile information
```
which is located in `localities_annotation.cpp`. Tracking the issue, I think `extract_numeric_value_dimensions(arg, name, codename)` does not produce the expected output when is called by `random_d`. In the above example, we expect the dimensions to be an array of (50,100,0,0) while we get (100,100,0,0) in test__1 for the first tile (we have a row tiling with numtiles=2). (Note that if you are on the master you would see (100,50,0,0))
Both data types are double in the above example, and I think both tests use the first two matrix definitions [here](https://github.com/STEllAR-GROUP/phylanx/blob/master/src/ir/node_data.cpp#L1610). I don't understand why we get (100,100) for the second test.

hkaiser:
- flyby: added missing std::move, remove superfluous std::move invocations

Fixes #1139

### DIFF
--- a/tests/regressions/execution_tree/random_d_1139.cpp
+++ b/tests/regressions/execution_tree/random_d_1139.cpp
@@ -1,0 +1,101 @@
+// Copyright (c) 2020 Bita Hasheminezhad
+//
+// Distributed under the Boost Software License, Version 1.0. (See accompanying
+// file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+#include <phylanx/phylanx.hpp>
+
+#include <hpx/hpx_init.hpp>
+#include <hpx/include/iostreams.hpp>
+#include <hpx/include/lcos.hpp>
+#include <hpx/include/util.hpp>
+#include <hpx/testing.hpp>
+
+#include <cstdint>
+#include <string>
+#include <utility>
+#include <vector>
+
+#include<iostream>
+
+///////////////////////////////////////////////////////////////////////////////
+void execution_time(std::string const& name, std::string const& codestr)
+{
+    phylanx::execution_tree::compiler::function_list snippets;
+    phylanx::execution_tree::compiler::environment env =
+        phylanx::execution_tree::compiler::default_environment();
+
+    auto const& compiled_code =
+        phylanx::execution_tree::compile(name, codestr, snippets, env);
+
+    hpx::cout << "Locality: " << hpx::get_locality_id()
+              << "  Starting execution for: " << name << hpx::endl;
+    std::uint64_t t = hpx::util::high_resolution_clock::now();
+    phylanx::execution_tree::primitive_argument_type a = compiled_code.run().arg_;
+    t = hpx::util::high_resolution_clock::now() - t;
+
+    std::cout << *(a.annotation()) << "\n\n\n";
+    hpx::cout << "       " << (t / 1e6) << " ms.\n" << hpx::flush;
+}
+
+////////////////////////////////////////////////////////////////////////////////
+void test__0()
+{
+    if (hpx::get_locality_id() == 0)
+    {
+        execution_time("perftest_0", R"(
+            dot_d(
+                constant_d(13., list(100, 100), 0, 2, "const_1"),
+                constant_d(13., list(100, 100), 0, 2, "const_2")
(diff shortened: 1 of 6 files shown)


tests/regressions/execution_tree/random_d_1139.cpp
```diff
@@ -0,0 +1,101 @@
...
+    auto const& compiled_code =
+        phylanx::execution_tree::compile(name, codestr, snippets, env);
+
+    hpx::cout << "Locality: " << hpx::get_locality_id()
+              << "  Starting execution for: " << name << hpx::endl;
+    std::uint64_t t = hpx::util::high_resolution_clock::now();
+    phylanx::execution_tree::primitive_argument_type a = compiled_code.run().arg_;
+    t = hpx::util::high_resolution_clock::now() - t;
+
+    std::cout << *(a.annotation()) << "\n\n\n";
```
taless474: I think this std::cout was good for demonstration, but we don't need it anymore and we can remove #include<iostream> too.

hkaiser: @taless474 right. Please feel free to push to this branch to apply these changes.

taless474:
I have used only `#include<iostream>` components and lowered the matrix size. Also removed the unused `using` in dist_random.hpp file.

PR #1143, opened 2020-04-02
Adding cannon product performance tests

taless474:
This PR adds a couple of tests that measure the execution time of cannon_product of different sized square matrices on 4, 9 and 16 localities.
The current version of cannon_product only works if the number of localities is a perfect square. To have the same matrix sizes for 4, 9 and 16 localities, the selected sizes are divisible by 12.


tests/performance/dist_cannon_16_loc.cpp
```diff
@@ -0,0 +1,71 @@
...
+///////////////////////////////////////////////////////////////////////////////
+char const* const cannon_product_code = R"(block(
+    define(cannon, dim_size,
+        block(
+            define(array1, random_d(list(dim_size, dim_size), find_here(), 16)),
+            define(array2, random_d(list(dim_size, dim_size), find_here(), 16)),
+            define(product,
+                constant_d(0.0, list(dim_size, dim_size), find_here(), 16)),
+            store (product, cannon_product(array1, array2)),
+            product
```
hkaiser, line 30: I think this can be simplified: 
```suggestion
            cannon_product(array1, array2)
```

tests/performance/dist_cannon_4_loc.cpp
```diff
@@ -0,0 +1,455 @@
...
+//    else
+//    {
+//        execution_time("perftest_cannon_4loc_960000", R"(
+//            cannon_product(
+//                random_d(list(960000, 960000), 3, 4, "rand_9_1"),
+//                random_d(list(960000, 960000), 3, 4, "rand_9_2")
+//            )
+//        )");
+//    }
+//}
```
hkaiser, line 453: Is this code still needed? If not, please remove it.

tests/performance/CMakeLists.txt
```diff
@@ -5,9 +5,16 @@
...
     blaze_benchmarks
+    dist_cannon_4_loc
+    dist_cannon_9_loc
+    dist_cannon_16_loc
     simple_loop
    )
 
+set(dist_cannon_4_loc_PARAMETERS LOCALITIES 4)
+set(dist_cannon_9_loc_PARAMETERS LOCALITIES 9)
+set(dist_cannon_16_loc_PARAMETERS LOCALITIES 16)
```
hkaiser, line 16: Not sure if our testing infrastructure can handle this. But we'll see.

tests/performance/dist_cannon_16_loc.cpp
taless474: thanks

tests/performance/CMakeLists.txt
taless474: I think performance tests are not added for typical runs, we don't put `  add_phylanx_unit_test("performance" ${test} ${${test}_PARAMETERS})` in their cmake, so we can run them only when we want.

hkaiser: Fair enough!

PR #1144, opened 2020-04-07
Add the dist_identity primitive

NanmiaoWu:
The dist_identity consider three tiling cases: 1. row-tiling with any number of tiles; 2. column-tiling with any number of tiles; 3. symmetric-tiling with 4 tiles.

### DIFF
--- a/phylanx/plugins/dist_matrixops/dist_identity.hpp
+++ b/phylanx/plugins/dist_matrixops/dist_identity.hpp
@@ -3,14 +3,16 @@
 // Distributed under the Boost Software License, Version 1.0. (See accompanying
 // file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
 
-#if !defined(PHYLANX_PRIMITIVES_DIST_RANDOM)
-#define PHYLANX_PRIMITIVES_DIST_RANDOM
+#if !defined(PHYLANX_PRIMITIVES_DIST_IDENTITY)
+#define PHYLANX_PRIMITIVES_DIST_IDENTITY
 
 #include <phylanx/config.hpp>
+#include <phylanx/execution_tree/annotation.hpp>
+#include <phylanx/execution_tree/localities_annotation.hpp>
 #include <phylanx/execution_tree/primitives/base_primitive.hpp>
 #include <phylanx/execution_tree/primitives/node_data_helpers.hpp>
 #include <phylanx/execution_tree/primitives/primitive_component_base.hpp>
-#include <phylanx/util/random.hpp>
+
 
 #include <hpx/lcos/future.hpp>
 
--- a/phylanx/plugins/dist_matrixops/dist_identity.hpp
+++ b/phylanx/plugins/dist_matrixops/dist_identity.hpp
@@ -1,0 +1,72 @@
+// Copyright (c) 2020 Nanmiao Wu
+//
+// Distributed under the Boost Software License, Version 1.0. (See accompanying
+// file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+#if !defined(PHYLANX_PRIMITIVES_DIST_RANDOM)
+#define PHYLANX_PRIMITIVES_DIST_RANDOM
+
+#include <phylanx/config.hpp>
+#include <phylanx/execution_tree/primitives/base_primitive.hpp>
+#include <phylanx/execution_tree/primitives/node_data_helpers.hpp>
+#include <phylanx/execution_tree/primitives/primitive_component_base.hpp>
+#include <phylanx/util/random.hpp>
+
+#include <hpx/lcos/future.hpp>
+
+#include <array>
+#include <cstddef>
+#include <cstdint>
+#include <memory>
+#include <random>
+#include <string>
+#include <tuple>
+#include <utility>
+#include <vector>
+
+#include <blaze/Math.h>
+
+namespace phylanx { namespace dist_matrixops { namespace primitives
+{
+    class dist_identity
+      : public execution_tree::primitives::primitive_component_base
+      , public std::enable_shared_from_this<dist_identity>
+    {
+    public:
+        static execution_tree::match_pattern_type const match_data;
+
+        dist_identity() = default;
+
+        dist_identity(execution_tree::primitive_arguments_type&& operands,
+            std::string const& name, std::string const& codename);
+
+    protected:
+        hpx::future<execution_tree::primitive_argument_type> eval(
+            execution_tree::primitive_arguments_type const& operands,
+            execution_tree::primitive_arguments_type const& args,
+            execution_tree::eval_context ctx) const override;
+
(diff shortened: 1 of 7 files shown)


NanmiaoWu:
The dist_identity consider three tiling cases: 1. row-tiling with any number of tiles; 2. column-tiling with any number of tiles; 3. symmetric-tiling with 4 tiles.

phylanx/plugins/dist_matrixops/dist_identity.hpp
```diff
@@ -0,0 +1,74 @@
...
+#include <phylanx/execution_tree/primitives/node_data_helpers.hpp>
+#include <phylanx/execution_tree/primitives/primitive_component_base.hpp>
+
+#include <hpx/lcos/future.hpp>
+
+#include <array>
+#include <cstddef>
+#include <cstdint>
+#include <memory>
+#include <random>
```
taless474, line 24: I think we do not need random here

tests/unit/plugins/dist_matrixops/dist_identity_2_loc.cpp
```diff
@@ -0,0 +1,193 @@
...
+void test_identity_d_operation(std::string const& name, std::string const& code,
+    std::string const& expected_str)
+{
+    phylanx::execution_tree::primitive_argument_type result =
+        compile_and_run(name, code);
+    phylanx::execution_tree::primitive_argument_type comparison =
+        compile_and_run(name, expected_str);
+
+    // comparing annotations
+    HPX_TEST_EQ(*(result.annotation()), *(comparison.annotation()));
```
taless474, line 47: We need to compare results and comparison, since we need their values to be the same too. Please change this to `HPX_TEST_EQ(result, comparison);`

src/plugins/dist_matrixops/dist_identity.cpp
```diff
@@ -0,0 +1,310 @@
...
+    {
+        // verify arguments
+        if (operands.size() < 1 || operands.size() > 6)
+        {
+            HPX_THROW_EXCEPTION(hpx::bad_parameter, "dist_identity::eval",
+                generate_error_message("the identity_d primitive requires "
+                                       "at least 1 and at most 6 operands"));
+        }
+
+        if (!valid(operands[1]) || !valid(operands[2]) || !valid(operands[3]))
```
taless474, line 223: I think as only the first operand should be given, we must check only `if(!valid(operands[0]))` (if I am not mistaken, the other are initialized and will be checked as args to see if a value is given or `nil` is assigned to them). 
I am sure this is due to my mistake in `random_d` (it should have been `operands[0]` instead of `operands[3]`) and sorry I wasn't clear about this.

src/plugins/dist_matrixops/dist_identity.cpp
```diff
@@ -0,0 +1,310 @@
...
+                tile_info.as_annotation(name_, codename_), ann_info, name_,
+                codename_));
+
+        blaze::DynamicMatrix<T> m(row_size, column_size, 0.0);
+        if (tiling_type == "row")
+        {
+            for (std::size_t i = 0; i != row_size; ++i)
+            {
+                std::size_t j = i + row_start;
+                m(i, j) = 1.0;
```
taless474, line 140: I am not sure but I think we might be able to use a `blaze::band`. Please refer to `eye_operation`

src/plugins/dist_matrixops/dist_identity.cpp
```diff
@@ -0,0 +1,310 @@
...
+    dist_identity::dist_identity(
+        execution_tree::primitive_arguments_type&& operands,
+        std::string const& name, std::string const& codename)
+      : primitive_component_base(std::move(operands), name, codename)
+    {
+    }
+
+    ///////////////////////////////////////////////////////////////////////////
+    namespace detail {
+        static std::atomic<std::size_t> const_count(0);
```
taless474, line 84: @hkaiser , as I think the namespace detail is common for `dist_identity` and other primitives in dist_matrixops, I have chosen different names for this variable in `dist_random` and `dist_constant`. This, `const_count`, is the same as `dist_constant`'s. Does it make a problem?

phylanx/plugins/dist_matrixops/dist_identity.hpp
NanmiaoWu: Yes, I already deleted that.

tests/unit/plugins/dist_matrixops/dist_identity_2_loc.cpp
NanmiaoWu: Changed to HPX_TEST_EQ(result, comparison).

src/plugins/dist_matrixops/dist_identity.cpp
NanmiaoWu: Changed to if(!valid(operands[0])).

src/plugins/dist_matrixops/dist_identity.cpp
NanmiaoWu: Changed to blaze::band.

src/plugins/dist_matrixops/dist_identity.cpp
NanmiaoWu: Changed to identity_count.

src/plugins/dist_matrixops/dist_identity.cpp
```diff
@@ -0,0 +1,297 @@
...
+        blaze::DynamicMatrix<T> m(row_size, column_size, T(0));
+        if (tiling_type == "row")
+        {
+            blaze::band(m, row_start) = T(1);
+        }
+        else if (tiling_type == "column")
+        {
+            blaze::band(m, -column_start) = T(1);
+        }
+        else if (tiling_type == "sym" && numtiles == 4)
```
hkaiser, line 143: Why is this limited to four tiles only?

src/plugins/dist_matrixops/dist_identity.cpp
```diff
@@ -0,0 +1,297 @@
...
+        {
+            if (column_start - row_start == 0)
+            {
+                blaze::band(m, 0) = T(1);
```
hkaiser, line 147: I think this will not do the right thing as it will fill all tiles with a diagonal of ones.

src/plugins/dist_matrixops/dist_identity.cpp
NanmiaoWu: Bita and I think that it may be easier to start with some simple cases. I could figure out the sym tiling with any number of tiles if needed.

src/plugins/dist_matrixops/dist_identity.cpp
NanmiaoWu: It works for sym tiling with 4 tiles: only the loc 0 and loc 3 will be filled with diagonal ones while loc 1 and loc 2 will have all zeros. But if we consider the general case, sym with any number of tiles, it will not work.

NanmiaoWu:
I am gonna close this pr and create a new one with general case for sym-tiling.

PR #1146, opened 2020-04-09
Add the dist_identity primitive

NanmiaoWu:
Consider three cases: row-tiling, column-tiling, and sym-tiling. All of the three cases are able to deal with any number of tiles, which means they are general.

### DIFF
--- a/src/plugins/dist_matrixops/dist_identity.cpp
+++ b/src/plugins/dist_matrixops/dist_identity.cpp
@@ -239,41 +239,24 @@ namespace phylanx { namespace dist_matrixops { namespace primitives {
 
                     std::int64_t sz = extract_scalar_integer_value_strict(
                         std::move(args[0]), this_->name_, this_->codename_);
-/*                        
-                    std::uint32_t tile_idx =
-                        extract_scalar_nonneg_integer_value_strict(
-                            std::move(args[1]), this_->name_, this_->codename_);
-                    std::uint32_t numtiles =
-                        extract_scalar_positive_integer_value_strict(
-                            std::move(args[2]), this_->name_, this_->codename_);
-                    if (tile_idx >= numtiles)
-                    {
-                        HPX_THROW_EXCEPTION(hpx::bad_parameter,
-                            "dist_identity::eval",
-                            this_->generate_error_message(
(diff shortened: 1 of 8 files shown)


src/plugins/dist_matrixops/dist_identity.cpp
```diff
@@ -0,0 +1,302 @@
...
+            blaze::band(m, row_start) = T(1);
+        }
+        else if (tiling_type == "column")
+        {
+            blaze::band(m, -column_start) = T(1);
+        }
+        else if (tiling_type == "sym")
+        {
+            std::int64_t num_band = row_start - column_start;
+            std::int64_t upper_band = static_cast<std::size_t>(column_size - 1);
```
hkaiser, line 146: Why do you cast to a `std::size_t` before assigning to a `std::int64_t`?

NanmiaoWu: In line 149 and line 150, when I use max and min, the two operands in max or min should be the same type. So I cast them to std::int64_t.

NanmiaoWu: ah. I misunderstand your question. Seems like I dont need to cast to a std::size_t before assigning to a std::int64_t.

src/plugins/dist_matrixops/dist_identity.cpp
```diff
@@ -0,0 +1,302 @@
...
+///////////////////////////////////////////////////////////////////////////////
+namespace phylanx { namespace dist_matrixops { namespace primitives {
+    ///////////////////////////////////////////////////////////////////////////
+    execution_tree::match_pattern_type const dist_identity::match_data = {
+
+        hpx::util::make_tuple("identity_d", std::vector<std::string>{R"(
+                identity_d(
+                    _1_sz,
+                    __arg(_2_tile_index, nil),
+                    __arg(_3_numtiles, nil),
```
hkaiser, line 46: As @taless474 has suggested in #1147:
```suggestion
                    __arg(_2_tile_index, find_here()),
                    __arg(_3_numtiles, num_localities()),
```

NanmiaoWu: Will do.

PR #1147, opened 2020-04-15
Modify random_d and constant_d to have tile_idx and numtiles as optional arguments

taless474:
This PR gets the values of `tile_idx` and `numtiles` from `find_here()` and `num_localities()` for the case that user does not provide them using `random_d` or `constant_d`.
Note that the shape for `constant` can contain zero while `constant_d` and `random_d` do not create empty arrays and do not accept zeros in their shapes.


src/plugins/dist_matrixops/dist_constant.cpp
```diff
@@ -45,8 +44,8 @@ namespace phylanx { namespace dist_matrixops { namespace primitives
                 constant_d(
                     _1_value,
                     _2_shape,
-                    __arg(_3_tile_index, nil),
-                    __arg(_4_numtiles, nil),
+                    __arg(_3_tile_index, find_here()),
+                    __arg(_4_numtiles, num_localities()),
```
hkaiser, line 48: Wow, I would never have thought of using a primitive invocation as the default value for an argument - nice!

PR #1149, opened 2020-04-20
Adding Retile for distributed arrays

taless474:
This PR adds `retile` for distributed arrays in 4 tiling types: `sym`, `row`, ` column` and `user`.
In any of the `sym`, `row`, and ` column` tilings, if an intersection is given, retile will produce tiles that have overlapped parts (with the length of the given intersection on that dimension).
`user` tiling type uses the `new_tiling` argument which specifies the spans for each dimension.
A change in the number of tiles is not possible in this PR.

### DIFF
--- a/phylanx/plugins/dist_matrixops/tile_calculation_helper.hpp
+++ b/phylanx/plugins/dist_matrixops/tile_calculation_helper.hpp
@@ -55,6 +55,41 @@ namespace tile_calculation
         return std::make_tuple(start, size);
     }
 
+    inline std::tuple<std::int64_t, std::size_t> tile_calculation_overlap_1d(
+        std::int64_t start, std::size_t size, std::size_t dim,
+        std::size_t intersection)
+    {
+        if (start == 0)
+        {
+        }
+        else if (start + size == dim)
+        {
+            start -= intersection;
+        }
+        else
+        {
+            start -= static_cast<std::size_t>(intersection / 2);
+        }
+        size += intersection;
+
+        if (start < 0)
+        {
+            HPX_THROW_EXCEPTION(hpx::bad_parameter,
+                "tile_calculation::tile_calculation_overlap_1d",
+                phylanx::util::generate_error_message(
+                    "the given intersection produces negative start for the "
+                    "array"));
+        }
+        if (start + size > dim)
+        {
+            HPX_THROW_EXCEPTION(hpx::bad_parameter,
+                "tile_calculation::tile_calculation_overlap_1d",
+                phylanx::util::generate_error_message(
+                    "the given intersection need an end point larger than the "
+                    "array size"));
+        }
+        return std::make_tuple(start, size);
+    }
     ///////////////////////////////////////////////////////////////////////////
     inline std::tuple<std::size_t, std::size_t> middle_divisors(std::size_t num)
     {
--- a/phylanx/plugins/dist_matrixops/tile_calculation_helper.hpp
+++ b/phylanx/plugins/dist_matrixops/tile_calculation_helper.hpp
@@ -90,6 +90,7 @@ namespace tile_calculation
         }
         return std::make_tuple(start, size);
     }
+
     ///////////////////////////////////////////////////////////////////////////
     inline std::tuple<std::size_t, std::size_t> middle_divisors(std::size_t num)
     {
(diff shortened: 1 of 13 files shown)


phylanx/plugins/dist_matrixops/tile_calculation_helper.hpp
```diff
@@ -55,6 +54,42 @@ namespace tile_calculation
         return std::make_tuple(start, size);
     }
 
+    inline std::tuple<std::int64_t, std::size_t> tile_calculation_overlap_1d(
+        std::int64_t start, std::size_t size, std::size_t dim,
+        std::size_t intersection)
+    {
+        if (start == 0)
+        {
+        }
```
hkaiser, line 63: Could you rewrite this `if()` such that we don't have empty branches, please?

src/plugins/dist_matrixops/retile_annotations.cpp
```diff
@@ -0,0 +1,811 @@
...
+
+        std::tuple<bool, std::int64_t, std::int64_t> tile_extraction_1d(
+            ir::range&& new_tiling)
+        {
+            ir::range_iterator it = new_tiling.begin();
+            std::string label = extract_string_value_strict(*it);
+            if (label == "args")
+            {
+                // extracting the "tile" tag
+                ++it; // passing the "locality" tag
```
hkaiser, line 123: Are you sure `it != end`?

src/plugins/dist_matrixops/retile_annotations.cpp
```diff
@@ -0,0 +1,811 @@
...
+        execution_tree::primitive_arguments_type&& operands,
+        std::string const& name, std::string const& codename)
+      : primitive_component_base(std::move(operands), name, codename)
+    {}
+
+    ///////////////////////////////////////////////////////////////////////////
+    namespace detail
+    {
+        std::tuple<bool, std::int64_t, std::int64_t> tile_extraction_1d_helper(
+            ir::range_iterator&& it)
```
hkaiser, line 99: Does this have to be an rvalue? I think passing iterators by value is normally safe.

phylanx/plugins/dist_matrixops/tile_calculation_helper.hpp
taless474: Actually, I don't know how to write the condition for `else` in this code. Can you guide me?

taless474: I take that back. I got it

src/plugins/dist_matrixops/retile_annotations.cpp
```diff
@@ -0,0 +1,811 @@
...
+        {
+            ir::range tile_data = extract_list_value_strict(*++it);
+            ir::range_iterator inner_it = tile_data.begin();
+
+            // type_: columns = 0, rows = 1
+            bool type_ =
+                extract_string_value_strict(*inner_it) == "rows" ? 1 : 0;
+            std::int64_t start =
+                extract_scalar_nonneg_integer_value_strict(*++inner_it);
```
hkaiser, line 108: All `extract_...` functions take the name and filename argument (they default to `"<unknown>"`). Error messages are much more readable if those are provided, however.

src/plugins/dist_matrixops/retile_annotations.cpp
```diff
@@ -0,0 +1,811 @@
...
+
+        // creating the updated array as result
+        auto v = arr.vector();
+        blaze::DynamicVector<T> result(des_size);
+        util::distributed_vector<T> v_data(
+            arr_localities.annotation_.name_, v, num_localities, loc_id);
+
+        // relative start
+        std::int64_t rel_start = des_start - cur_start;
+        if (rel_start < 0 || des_stop > cur_stop)
```
hkaiser, line 321: Should this be `des_stop >= cur_stop`?

taless474: I think it should be `>`. If they are equal, still `else` part can get it. Am I missing something?

src/plugins/dist_matrixops/retile_annotations.cpp
taless474: Okay, sure

src/plugins/dist_matrixops/retile_annotations.cpp
hkaiser: Ok, I was merely asking as normally the upper limit is not part of a (open) range.

phylanx/util/detail/range_dimension.hpp
```diff
@@ -0,0 +1,109 @@
...
+    extract_positive_range_dimensions(phylanx::ir::range const& shape)
+    {
+        std::array<std::size_t, PHYLANX_MAX_DIMENSIONS> result = {0};
+        if (!shape.empty())
+        {
+            if (shape.size() == 1)
+            {
+                result[0] = extract_scalar_positive_integer_value_strict(
+                    *shape.begin());
+            }
```
hkaiser, line 32: Could you add the name and filename to these `extract_` functions as well, please?

taless474: Oh, sorry I missed that

PR #1151, opened 2020-04-24
Adding locality parameter to python execution interface

hkaiser:
- flyby: adding sleep primitive


taless474:
@hkaiser, are changes on tests/unit/distributed/remote_add.cpp on purpose?

hkaiser:
> @hkaiser, are changes on tests/unit/distributed/remote_add.cpp on purpose?

Uhh, no. No idea how that happened - thanks for noticing.

stevenrbrandt:
```
import time

# This is required now.
PhylanxSession.init(2)

@Phylanx
def find_here():
    h = find_here()
    print("here is:",h)
    return h

fh = find_here.launch(locality=0)
print("here:",fh())
```
the above code prints nothing. The `find_here()` function is somehow now returning None, apparently.

Also, the `PhylanxSession.init(1)` is no mandatory. The code won't run without it, even if I try to launch with `hpxrun.py`.

stevenrbrandt:
The issue I have is this:
- There's no way to send data from one locality to another inside a phylanx function
- Arrays created in phylanx functions are deallocated when the function exits
- Therefore all distributed array functions must take place in a context where localities can't exchange data

This situation would be fixed if we had access to a global data structure inside phylanx functions. That way, data could be exchanged between invocations of phylanx functions.

Issue #1155, opened 2020-04-29
Results of the distributed arithmetic primitives


taless474:
The following test uses one of the numeric primitives, `maximum`, and is run on 2 localities. We want to get the maximum of two arrays, `[13., 42., 33.]` and `[101., 12., 65.]` . They are compared element by element, so the expected result is `[101., 42., 65.]`.
In the last example the first array is compared to `35.` and we expect to get `[35., 42., 35.]` by broadcasting.
```c++
void test_maximum_1d_0()
{
    if (hpx::get_locality_id() == 0)
    {
        test_dist_operation("test_maximum_1d_2loc_0", R"(
            maximum(
                annotate_d([13., 42.], "lhs",
                    list("tile", list("columns", 0, 2))),
                annotate_d([101., 12.], "rhs",
                    list("tile", list("columns", 0, 2))))
        )", R"(
            annotate_d([101., 42.0], "lhs/0",
                list("args",
                    list("locality", 0, 2),
                    list("tile", list("columns", 0, 2))))
        )");
    }
    else
    {
        test_dist_operation("test_maximum_1d_2loc_0", R"(
            maximum(
                annotate_d([33.], "lhs",
                    list("tile", list("columns", 2, 3))),
                annotate_d([65.], "rhs",
                    list("tile", list("columns", 2, 3))))
        )", R"(
            annotate_d([65.], "lhs/0",
                list("args",
                    list("locality", 1, 2),
                    list("tile", list("columns", 2, 3))))
        )");
    }
}

void test_maximum_1d_1()
{
    if (hpx::get_locality_id() == 0)
    {
        test_dist_operation("test_maximum_1d_2loc_1", R"(
            maximum(
                annotate_d([13., 42.], "lhs_1",
                    list("tile", list("columns", 0, 2))),
                annotate_d([12., 65.], "rhs_1",
                    list("tile", list("columns", 1, 3))))

        )", R"(
            annotate_d([101., 42.0], "lhs_1/0",
                list("args",
                    list("locality", 0, 2),
                    list("tile", list("columns", 0, 2))))
        )");
    }
    else
    {
        test_dist_operation("test_maximum_1d_2loc_1", R"(
            maximum(
                annotate_d([33.], "lhs_1",
                    list("tile", list("columns", 2, 3))),
                annotate_d([101.], "rhs_1",
                    list("tile", list("columns", 0, 1))))
        )", R"(
            annotate_d([65.], "lhs_1/0",
[...]

PR #1156, opened 2020-04-29
Fixing annotation propagation problems

hkaiser:
Fixes: #1155


hkaiser:
@taless474 this fixes two of the problems you have described in #1155. The remaining two are more difficult to address as they would require adding some communication to the primitives.

PR #1158, opened 2020-04-30
Modify statistics given an empty list for axis

taless474:
In NumPy having an empty axis is not the same as not having an axis:
```python
np.all([-1, 0, 1], axis=())              # array([ True, False,  True])
np.any([-1, 0, 1], axis=())              # array([ True, False,  True])
np.max([-1, 0, 1], axis=())              # array([-1,  0,  1])
np.min([-1, 0, 1], axis=(), initial=0)   # array([-1,  0,  0])
np.mean([-1, 0, 1], axis=())             # array([-1.,  0.,  1.])
np.prod([-1, 0, 1], axis=())             # array([-1,  0,  1])
np.std([-1, 0, 1], axis=())              # array([0., 0., 0.])
np.var([-1, 0, 1], axis=())              # array([0., 0., 0.])
np.sum([-1, 0, 1], axis=())              # array([-1,  0,  1])
```
This PR modifies statistic base to comply with the mentioned feature.


hkaiser:
@taless474 could you please rebase this on master to fix the build system issues?

Issue #1157, opened 2020-04-30
`astype` does not work with `define`


taless474:
Although `astype([1, 0, 1], "bool")` works properly, the following code (in physl) fails in an assertion:
```c++
define(a, [1, 0, 1])
astype(a, "bool")
```
```console
{version}: V1.5.0-trunk (AGAS: V3.0), Git: 6df8ceca6a
{boost}: V1.72.0
{build-type}: debug
{date}: Apr 27 2020 22:02:32
{platform}: Win32
{compiler}: Microsoft Visual C++ version 14.1
{stdlib}: Dinkumware standard library version 650
{env}: 51 entries:
  ALLUSERSPROFILE=C:\ProgramData
  APPDATA=C:\Users\Bita\AppData\Roaming
  COMPUTERNAME=UNICK
  ComSpec=C:\Windows\system32\cmd.exe
  CommonProgramFiles(x86)=C:\Program Files (x86)\Common Files
  CommonProgramFiles=C:\Program Files\Common Files
  CommonProgramW6432=C:\Program Files\Common Files
  DriverData=C:\Windows\System32\Drivers\DriverData
  FPS_BROWSER_APP_PROFILE_STRING=Internet Explorer
  FPS_BROWSER_USER_PROFILE_STRING=Default
  HOMEDRIVE=C:
  HOMEPATH=\Users\Bita
  LOCALAPPDATA=C:\Users\Bita\AppData\Local
  LOGONSERVER=\\UNICK
  MSBuildLoadMicrosoftTargetsReadOnly=true
  MSMPI_BIN=C:\Program Files\Microsoft MPI\Bin\
  NUMBER_OF_PROCESSORS=48
  OS=Windows_NT
  OneDrive=C:\Users\Bita\OneDrive
  PATHEXT=.COM;.EXE;.BAT;.CMD;.VBS;.VBE;.JS;.JSE;.WSF;.WSH;.MSC
  PROCESSOR_ARCHITECTURE=AMD64
  PROCESSOR_IDENTIFIER=Intel64 Family 6 Model 85 Stepping 4, GenuineIntel
  PROCESSOR_LEVEL=6
  PROCESSOR_REVISION=5504
  PSModulePath=C:\Program Files\WindowsPowerShell\Modules;C:\Windows\system32\WindowsPowerShell\v1.0\Modules
  PUBLIC=C:\Users\Public
[...]

PR #1159, opened 2020-04-30
Making sure all_of() always returns boolean values

hkaiser:
- fixing astype() to properly support variables
- flyby: add support for constant boolean arrays to PhySL parser

Fixes: #1157


taless474:
Thank you @hkaiser.
I think when the result is 0d, `all` and `any` result types are non-boolean.
```c++
any([[1,3,0],[1,0,0]])
any([[1,3,0],[1,0,0]],0)
```
```console
1
[true, true, false]
```

hkaiser:
> I think when the result is 0d, all and any result types are non-boolean.

The result was correct, the printing was not. Things should be better now.

PR #1161, opened 2020-05-02
Add primitive dist_diag. 

NanmiaoWu:
It is the distributed version of Numpy diag for 1-D array as input. Remove diag0d to comply with Numpy.


src/plugins/dist_matrixops/dist_diag.cpp
```diff
@@ -0,0 +1,576 @@
...
+#include <phylanx/util/distributed_matrix.hpp>
+#include <phylanx/util/distributed_vector.hpp>
+
+#include <hpx/assertion.hpp>
+#include <hpx/errors/throw_exception.hpp>
+#include <hpx/include/lcos.hpp>
+#include <hpx/include/naming.hpp>
+#include <hpx/include/util.hpp>
+
+#include <atomic>
```
taless474, line 29: I think atomic is not needed

src/plugins/dist_matrixops/dist_diag.cpp
```diff
@@ -0,0 +1,576 @@
...
+#include <phylanx/execution_tree/localities_annotation.hpp>
+#include <phylanx/execution_tree/locality_annotation.hpp>
+#include <phylanx/execution_tree/meta_annotation.hpp>
+#include <phylanx/execution_tree/primitives/annotate_primitive.hpp>
+#include <phylanx/execution_tree/primitives/node_data_helpers.hpp>
+#include <phylanx/execution_tree/tiling_annotations.hpp>
+#include <phylanx/ir/node_data.hpp>
+#include <phylanx/plugins/dist_matrixops/dist_diag.hpp>
+#include <phylanx/plugins/dist_matrixops/tile_calculation_helper.hpp>
+#include <phylanx/util/detail/range_dimension.hpp>
```
taless474, line 19: I believe range_dimension is not used

src/plugins/dist_matrixops/dist_diag.cpp
```diff
@@ -0,0 +1,576 @@
...
+        execution_tree::primitive_arguments_type&& operands,
+        std::string const& name, std::string const& codename)
+      : primitive_component_base(std::move(operands), name, codename)
+    {}
+
+    ///////////////////////////////////////////////////////////////////////////
+
+    namespace detail
+    {
+        //static std::atomic<std::size_t> diag_count(0);
```
taless474, line 88: I think it would be best if you put these functions in an hpp file, so both retile and diag access it.

src/plugins/dist_matrixops/dist_diag.cpp
```diff
@@ -0,0 +1,576 @@
...
+            tiling_span(column_start, column_start + column_size));
+
+        auto locality_ann = arr_localities.locality_.as_annotation();
+        auto attached_annotation =
+            std::make_shared<annotation>(localities_annotation(locality_ann,
+                tile_info.as_annotation(name_, codename_),
+                arr_localities.annotation_, name_, codename_));
+
+        return primitive_argument_type(result, attached_annotation);
+    }
```
hkaiser, line 432: Could you extract the various cases into separate functions, please? This function is very large and difficult to read.

src/plugins/matrixops/diag_operation.cpp
```diff
@@ -81,41 +75,6 @@ namespace phylanx { namespace execution_tree { namespace primitives
...
-            break;
-        }
-
-        HPX_THROW_EXCEPTION(hpx::bad_parameter,
-            "diag_operation::diag0d",
-            generate_error_message(
-                "the diag primitive requires for all arguments to "
-                    "be numeric data types"));
-    }
-
```
hkaiser, line 118: What's the rationale to remove `diag` for scalars? Is this change related to `diag_d`?

src/plugins/dist_matrixops/dist_diag.cpp
NanmiaoWu: will delete it

src/plugins/dist_matrixops/dist_diag.cpp
NanmiaoWu: Will delete it

src/plugins/dist_matrixops/dist_diag.cpp
NanmiaoWu: Already put them in a new hpp file, called idx_tile_calculation_helper.hpp

src/plugins/matrixops/diag_operation.cpp
NanmiaoWu: Because Numpy diag does not consider scalar case. To comply with it, I delete the diag0d case.

src/plugins/dist_matrixops/dist_diag.cpp
NanmiaoWu: Will do it.

hkaiser:
@NanmiaoWu could you please rebase this on master to fix the build system issues?

NanmiaoWu:
> @NanmiaoWu could you please rebase this on master to fix the build system issues?

Yes. Already done it.

phylanx/plugins/dist_matrixops/idx_tile_calculation_helper.hpp
```diff
@@ -0,0 +1,61 @@
...
+
+#if !defined(PHYLANX_DIST_MATRIXOPS_IDX_TILE_CALCULATION_HELPER)
+#define PHYLANX_DIST_MATRIXOPS_IDX_TILE_CALCULATION_HELPER
+
+#include <hpx/assertion.hpp>
+
+#include <cstddef>
+#include <cstdint>
+#include <string>
+#include <tuple>
```
taless474, line 16: I think tuple is not needed here

NanmiaoWu: Yes. Already deleted it.

src/plugins/dist_matrixops/dist_diag.cpp
```diff
@@ -0,0 +1,512 @@
...
+                    options are `row` or `column` tiling.
+                tile_index (int, optional): the tile index we need to generate
+                    the diag array for. A non-negative integer. If not
+                    given, it sets to current locality.
+                numtiles (int, optional): number of tiles of the returned array.
+                    If not given it sets to the number of localities in the
+                    application.
+            Returns:
+
+            A 1-D array of its k-th diagonal when a is a 2-D array; a 2-D array
```
taless474, line 72: Here we don't have dist_diag2d, so I guess we can remove "A 1-D array of its k-th diagonal when a is a 2-D array".

src/plugins/dist_matrixops/dist_diag.cpp
```diff
@@ -0,0 +1,511 @@
...
+                    if (tile_idx >= numtiles)
+                    {
+                        HPX_THROW_EXCEPTION(hpx::bad_parameter,
+                            "dist_diag::eval",
+                            this_->generate_error_message(
+                                "invalid tile index. Tile indices start from 0 "
+                                "and should be smaller than number of tiles"));
+                    }
+
+
```
taless474, line 503: The user might supply `a` with any dimension rather than 1. Would you please throw an exception in the case that `a`'s dimension is not 1?

src/plugins/dist_matrixops/dist_diag.cpp
NanmiaoWu: Yes. Will delete it.

src/plugins/dist_matrixops/dist_diag.cpp
NanmiaoWu: I am not sure I understand. Do you mean arr when mentioning a? And, why arr's dimension has to be 1? Or do you mean that it should be larger or equal to 1?

taless474: I mean if the user asks for the following (a 2-d array) what would be the result:
```
diag_d(annotate_d([[1,2,3],[4,5,6]],"name", list("tile",list("columns",0,3),list("rows",0,2))))
```

NanmiaoWu: ahh got you.because we currently only consider 1d array as input.

PR #1164, opened 2020-05-06
Second attempt at allowing raw physl code to be passed to a Python

stevenrbrandt:
This pull request allows users to create Python functions from raw PhySL code using the doc string as follows:
```
@Phylanx(doc_src=True)
def addme(a,b):
    """
    define(addme,a,b,a+b)
    """

print(addme(3,4))
```
By circumventing the Python to PhySL translation, this PR allows users to test new features more quickly and workaround bugs in the translator code.

### DIFF
--- a/python/phylanx/ast/physl.py
+++ b/python/phylanx/ast/physl.py
@@ -460,7 +460,7 @@ class PhySL:
         self.__ast__ = None
         self.ir = None
         self.python_tree = tree
-        if 'doc_src' in kwargs and kwargs['doc_src'] == True:
+        if 'doc_src' in kwargs and kwargs['doc_src']:
             self.doc_src = func.__doc__
         else:
             self.doc_src = None
--- a/python/phylanx/ast/physl.py
+++ b/python/phylanx/ast/physl.py
@@ -392,25 +392,29 @@ class PhySL:
                 physl_db.insert(
                     self.wrapped_function.__name__, self.__src__, self.__ast__)
 
-            if self.doc_src is not None:
-                self.__src__ = self.doc_src
-
             physl_db.close()
 
-        except Exception:
+        except Exception as e:
+            # close database, if needed
+            if physl_db is not None:
+                physl_db.close()
+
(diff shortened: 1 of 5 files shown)


hkaiser:
@stevenrbrandt could you please rebase this on master to fix the build system issues?

python/phylanx/ast/physl.py
```diff
@@ -383,12 +383,17 @@ def _compile_or_load(self):
                 check_return(self.ir)
 
                 self.__src__ = self._generate_physl(self.ir)
+                if self.doc_src is not None:
+                    self.__src__ = self.doc_src
```
hkaiser, line 387: I think that if the doc-string should be used as the PhySL there is no reason to go run the `_generate_physl()` function here as well.

python/phylanx/ast/physl.py
```diff
@@ -399,6 +404,9 @@ def _compile_or_load(self):
             self.__src__ = self._generate_physl(self.ir)
             self.__ast__ = phylanx.ast.generate_ast(self.__src__)
 
+            if self.doc_src is not None:
+                self.__src__ = self.doc_src
```
hkaiser, line 408: Why would we need to replace the `__src__` here if we have successfully loaded it from the database?

python/phylanx/ast/physl.py
```diff
@@ -383,12 +383,17 @@ def _compile_or_load(self):
...
+                if self.doc_src is not None:
+                    self.__src__ = self.doc_src
                 self.__ast__ = phylanx.ast.generate_ast(self.__src__)
 
                 # now store the PhySL string and AST for this function
                 physl_db.insert(
                     self.wrapped_function.__name__, self.__src__, self.__ast__)
 
+            if self.doc_src is not None:
+                self.__src__ = self.doc_src
```
hkaiser, line 395: Why is replace `__src__` necessary here? It was either successfully loaded from the database of set above.

hkaiser:
FWIW, this also fixes #1170

PR #1165, opened 2020-05-06
Modifying shape and adding shape_d

taless474:
This PR modifies the `extract_shape` primitive to always return the local shape, whether it is given a local array or a distributed one.
It also adds `shape_d` to calculate the whole shape of the given distributed array.
Having the following:
```c++
constant_d(0.0, shape_d(x, 0))
```
we expect the overall shape of the array.

### DIFF
--- a/src/plugins/matrixops/extract_shape.cpp
+++ b/src/plugins/matrixops/extract_shape.cpp
@@ -39,8 +39,9 @@ namespace phylanx { namespace execution_tree { namespace primitives
 
             Without the optional argument, it returns a list of integers
             corresponding to the size of each dimension of the array `a`. If the
-            optional `dim` argument is supplied, then the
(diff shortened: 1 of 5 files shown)


src/plugins/matrixops/extract_shape.cpp
```diff
@@ -160,7 +197,7 @@ namespace phylanx { namespace execution_tree { namespace primitives
             index += 2;
 
         std::int64_t size = 0;
-        if (arg.has_annotation())
+        if (mode_ == dist_mode)
```
hkaiser, line 200: We should probably still make sure the object actually has an annotation, otherwise we should fall back to the local case.

hkaiser:
@taless474 this needs rebasing on master as well. Could you do that, please?

Issue #1162, opened 2020-05-03
Add support for distributed kmeans


taless474:
To have a distributed version of kmeans algorithm, we can use the distributed version of the following primitives, 
- [ ] expand_dims
- [ ] slice (slice_column, slice_row)
- [ ] argmin
- [ ] sqrt, square (generic operations)
- [ ] sum (statistics)
- [ ] shape

according to:
```c++
char const* const kmeans_code = R"(
    define(closest_centroids, points, centroids,
        block(
            define(points_x,
                expand_dims(slice_column(points, 0), -1)
            ),
            define(points_y,
                expand_dims(slice_column(points, 1), -1)
            ),
            define(centroids_x,
                slice_column(centroids, 0)
            ),
            define(centroids_y,
                slice_column(centroids, 1)
            ),
            argmin(sqrt(
                        square(points_x - centroids_x)
                        + square(points_y - centroids_y)
                ), 1
            )
        )
    )

    define(move_centroids, points, closest, centroids, num_points, num_centroids,
        block(
        define(result, constant(0.0, list(num_centroids, 2))),
        define(count, 0),
        for(define(k, 0), k < num_centroids, store(k, k + 1), block(
            store(count, 0),
            for(define(i, 0), i < num_points, store(i, i + 1),block(
                if(slice(closest, i) == k, block(
                    store(slice(result, k, list(0, 2)),
                        slice_row(result, k) + slice_row(points, i)),
                    store(count, count + 1)
                ))
            )),
            if(count,
                store(slice(result, k, list(0, 2)), slice_row(result, k)/count)
            ))
        ), result
    ))

    define(kmeans, points, iterations, initial_centroids, enable_output,
        block(
            define(centroids, initial_centroids),
            define(num_centroids, shape(centroids, 0)),
            define(num_points, shape(points, 0)),
            for(define(i, 0), i < iterations, store(i, i + 1),
                block(
                    if(enable_output, block(
                        cout("centroids in iteration ", i,": ", centroids)
                    )),
                    store(centroids,
                          move_centroids(points,
                                        closest_centroids(points,
                                                         centroids),
                                        centroids,
                                        num_points,
                                        num_centroids))
                )
            ), centroids
        )
    )
)";
```

PR #1167, opened 2020-05-07
Adding 3D kmeans

taless474:
This PR:
- adds a 3d version of kmeans algorithm (only `closest_centroid` function is different)
- simplifies kmeans_fix_point to be more compatible with its primitive version; it does not require `apply`, `fmap`, `lambda`, `range` or `vstack`, however, it uses `slice_assign`.
- adds a test on 2 localities for distributed generic operations, on sqrt and square. (#1162)

### DIFF
--- a/examples/algorithms/kmeans/CMakeLists.txt
+++ b/examples/algorithms/kmeans/CMakeLists.txt
@@ -5,6 +5,7 @@
 
 set(example_programs
     kmeans
+    kmeans_fix_points_3d
     kmeans_fix_points
    )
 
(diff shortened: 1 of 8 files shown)


taless474:
In Python we have:
```py
import numpy as np
a = np.array([[1,2,3],[4,5,6]])
print(a[:][1]) # [4, 5, 6] 
print(a[1][:]) # [4, 5, 6]
print(a[:,1])  # [2, 5]
print(a[1,:])  # [4, 5, 6]
```
We represent that here with:
```
cout(slice([[1,2,3],[4,5,6]], 1, nil))
cout(slice([[1,2,3],[4,5,6]], nil, 1))
cout(slice_column([[1,2,3],[4,5,6]], 1))
cout(slice_row([[1,2,3],[4,5,6]], 1))
```
I modified the dimension for the first two cases in this PR.

PR #1176, opened 2020-05-20
Lda basic

ct-clmsn:
import of lda - white paper algorithm


ct-clmsn:
needs work on distributed portion. lda_trainer.hpp:225 shows the variable 'wp' needs to be distributed and requires a sum reduction and broadcast on that line to operate correctly.

ct-clmsn:
moved implementation code from hpp to cpp files. removed impl directory from header path. fixed outstanding flake8 issues in python code.

PR #1177, opened 2020-05-21
Distributed Inverse Operation

rhecto1:
Pull request for distributed inverse operation with tests on 2 and 3 localities. Input and output matrices tiled by columns.

### DIFF
--- a/src/plugins/dist_matrixops/dist_inverse_operation.cpp
+++ b/src/plugins/dist_matrixops/dist_inverse_operation.cpp
@@ -36,10 +36,8 @@
 
 #include <phylanx/util/distributed_matrix.hpp>
 
-// This is just boilerplate
 namespace phylanx { namespace dist_matrixops { namespace primitives {
-    // Declare the help_string details, just for explaining what
-    // a primitive is doing, if needed.
+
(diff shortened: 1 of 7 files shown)


src/plugins/dist_matrixops/dist_inverse_operation.cpp
```diff
@@ -0,0 +1,491 @@
...
+
+                        }
+                    }
+                }
+
+               //hpx::lcos::barrier b7("barrierg" + std::to_string(current_row) +
+               //         "_" + lhs_localities.annotation_.name_,
+               //     lhs_localities.locality_.num_localities_,
+               //     lhs_localities.locality_.locality_id_);
+               // b7.wait();
```
hkaiser, line 356: Would you mind removing the commented code in this file?

rhecto1: I only the removed the comments that were notes to myself (such as "this is mostly boilerplate, just change 'X,Y,Z' ") and the commented out barriers. The comments that describe the functionality are still there, but I can definitely add back the other comments if that would be better.

PR #1178, opened 2020-05-25
Implementing argmin_d and argmax_d

hkaiser:
This adds a first implementation for the `argmin_d` and `argmax_d` primitives.

@taless474 This does not have any tests yet (sorry) and only implements the (simplest) 1D variation. But I think this should get you started.


hkaiser:
@taless474 I think the problems we discussed are solved now. Please check again.

Issue #1187, opened 2020-06-02
Fixing last failed test for Release build


khuck:
The release build has one last failing test:
http://omega.nic.uoregon.edu:8020/#/builders/2/builds/167/steps/16/logs/stdio

I tracked this down to temporaries `output.tensor()` and `target.tensor()` going out of scope in `cat_cross_operation::cat_cross3d()`: https://github.com/STEllAR-GROUP/phylanx/blob/64540c34dcec64530cc5c9239fb36144b1302bee/src/plugins/keras_support/categorical_crossentropy_operation.cpp#L246-L297

I'll submit a pull request for that issue.

However, after fixing that crash, the python code fails assertions:

```
AssertionError: [[ 0.96681002  2.18922338  3.41163674  4.6340501 ]
 [ 5.85646346  7.07887682  8.30129018  9.52370354]
 [10.7461169  11.96853026 13.19094362 14.41335698]]
Uncaught exception. Entering post mortem debugging
Running 'cont' or 'step' will restart the program
> /var/lib/buildbot/slaves/phylanx/x86_64-gcc7-release/build/tests/unit/python/execution_tree/categorical_crossentropy.py(79)<module>()
-> assert np.all(np.abs(v1 - v2) < 1.0e-13), v1 - v2
(Pdb) p v1
array([[ 3.,  7., 11., 15.],
       [19., 23., 27., 31.],
       [35., 39., 43., 47.]])
(Pdb) p v2
array([[ 2.03318998,  4.81077662,  7.58836326, 10.3659499 ],
       [13.14353654, 15.92112318, 18.69870982, 21.47629646],
       [24.2538831 , 27.03146974, 29.80905638, 32.58664302]])
(Pdb) list
 74  	    o = 3 * np.ones(10)
 75  	    o[4] = 2
 76  	
 77  	    v1 = cc(t, o, flag, -1)
 78  	    v2 = cat_cross(t, o, flag)
 79  ->	    assert np.all(np.abs(v1 - v2) < 1.0e-13), v1 - v2
 80  	
 81  	    t = np.linspace(1.0e-9, 10e-9, 10)
 82  	    t[2] = 10
 83  	    o = 3 * np.ones(10)
 84  	    o[4] = 2
(Pdb) 
```

hkaiser:
Which test is failing, concretely? Could you elaborate, please?

khuck:
[...]

PR #1188, opened 2020-06-02
Fixing temporary variables that are going out of scope in Release bui…

khuck:
…ld execution, see #1187


hkaiser:
Are you sure those values returned from `.tensor()` went out of scope too early? The code here looks ok without your changes (but those shouldn't hurt in any way either).

khuck:
> Are you sure those values returned from `.tensor()` went out of scope too early? The code here looks ok without your changes (but those shouldn't hurt in any way either).

Honestly, because it's a Release build I couldn't get any info debugging.  So I just made some assumptions.  I am being a bit more disciplined with it now though - the debug build also fails with the assertion error, so I am replacing temporaries one at a time until I find the sweet spot between correct execution (no assertion error) and not crashing.

PR #1189, opened 2020-06-04
Stop-gap measure to work around broken HPX #includes

hkaiser:
Please note that this is just a short-term measure for people who can't wait for HPX being fixed. Once https://github.com/STEllAR-GROUP/hpx/pull/4710 is merged, things should be fine again.


hkaiser:
@NanmiaoWu this should fix your current issues with building Phylanx on top of HPX master

PR #1190, opened 2020-06-04
Give the option to physl to start a phylanx function at line one.

stevenrbrandt:
When running a Phylanx function, it's sometimes beneficial to start the line numbers for the function at line 1 rather than whatever the line is in the source file in which it is embedded. This PR provides an option to make that possible.

### DIFF
--- a/python/phylanx/ast/physl.py
+++ b/python/phylanx/ast/physl.py
@@ -230,13 +230,13 @@ def print_physl_src(src, with_symbol_info=False, tag=4):
     print("", sep="")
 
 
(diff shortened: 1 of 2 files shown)


hkaiser:
Wouldn't this be achieved with a lot less effort by making this conditional: 
https://github.com/STEllAR-GROUP/phylanx/blob/master/python/phylanx/ast/transducer.py#L93

By default, the source tree associated with an annotated block of Python code is numbered starting line one. We currently increment that to adjust for the position of the annotated code in the original source file.

PR #1193, opened 2020-06-16
Add distributed all_gather primitive

NanmiaoWu:
Add distributed all_gather primitive.


src/plugins/dist_matrixops/all_gather.cpp
```diff
@@ -0,0 +1,383 @@
...
+                    -> execution_tree::primitive_argument_type
+                    {
+                        using namespace execution_tree;
+
+                        switch (extract_numeric_value_dimension(
+                            std::move(args[0]), this_->name_, this_->codename_))
+                        {
+
+                            case 1:
+                                HPX_FALLTHROUGH;
```
taless474, line 365: If I understand it correctly, it should work for 1d arrays, too. I didn't find any test for that, would you please add some?

src/plugins/dist_matrixops/all_gather.cpp
```diff
@@ -0,0 +1,383 @@
...
+    execution_tree::primitive_argument_type all_gather::all_gather2d(
+        execution_tree::primitive_argument_type&& arr) const
+    {
+        using namespace execution_tree;
+
+        execution_tree::localities_information locs =
+            extract_localities_information(arr, name_, codename_);
+
+        std::size_t ndim = locs.num_dimensions();
+        if (ndim > 2 || ndim < 1)
```
taless474, line 286: It might not be necessary, but it would be good if it works with empty arrays too

src/plugins/dist_matrixops/all_gather.cpp
```diff
@@ -0,0 +1,383 @@
...
+            }
+            blaze::DynamicMatrix<T> result(total_rows, prevdim[1]);
+            std::size_t step = 0;
+            for (auto&& arg : args)
+            {
+                auto&& val =
+                    execution_tree::extract_node_data<T>(std::move(arg));
+                std::size_t num_rows = val.dimension(0);
+                auto m = val.matrix();
+                for (std::size_t j = 0; j != num_rows; ++j)
```
taless474, line 125: I think if we use a submatrix here we can get rid of the for loop.

src/plugins/dist_matrixops/all_gather.cpp
```diff
@@ -0,0 +1,383 @@
...
+#include <phylanx/execution_tree/tiling_annotations.hpp>
+#include <phylanx/ir/node_data.hpp>
+#include <phylanx/plugins/dist_matrixops/all_gather.hpp>
+#include <phylanx/plugins/dist_matrixops/tile_calculation_helper.hpp>
+#include <phylanx/plugins/matrixops/concatenate.hpp>
+#include <phylanx/util/distributed_matrix.hpp>
+#include <phylanx/util/distributed_vector.hpp>
+#include <phylanx/util/generate_error_message.hpp>
+#include <phylanx/util/index_calculation_helper.hpp>
+
```
taless474, line 24: I guess some of these includes are not needed; I didn't find utilization for `tile_calculation_helper` or `index_calculation_helper`

src/plugins/dist_matrixops/all_gather.cpp
```diff
@@ -0,0 +1,383 @@
...
+    ///////////////////////////////////////////////////////////////////////////
+    template <typename T>
+    execution_tree::primitive_argument_type all_gather::all_gather2d(
+        ir::node_data<T>&& arr,
+        execution_tree::localities_information&& locs) const
+    {
+        using namespace execution_tree;
+        auto m = arr.matrix();
+        blaze::DynamicMatrix<T> res_value(m.rows(), m.columns());
+        res_value = m;
```
taless474, line 230: can we write the above three lines as `blaze::DynamicMatrix<T> res_value = arr.matrix()` if we need to allocate memory?

src/plugins/dist_matrixops/all_gather.cpp
NanmiaoWu: I think I should delete 1d case here. Because I only consider 2d case. The reason why I added 1d here is that  I assumed a matrix which has only one row or one column is 1d. But it should still be 2d, right?

src/plugins/dist_matrixops/all_gather.cpp
NanmiaoWu: Yes. It makes sense to consider empty arrays. I assume what I need to add is that ndim can also be equal to 0

src/plugins/dist_matrixops/all_gather.cpp
taless474: yeah, I believe that's a 2d, too.

NanmiaoWu: OK. Will delete that 1d case

src/plugins/dist_matrixops/all_gather.cpp
NanmiaoWu: Yes. Will use submatrix instead.

src/plugins/dist_matrixops/all_gather.cpp
NanmiaoWu: Yes. Will remove them.

src/plugins/dist_matrixops/all_gather.cpp
NanmiaoWu: Yes. Will revise it.

NanmiaoWu:
@hkaiser Sure. Will do it.

PR #1204, opened 2020-06-28
[WIP] Change the CMakeLists.txt to output phylanx wheel

git-kale:
The changes in code will output a wheel after the build instead of the installation on system. The wheel file will only contain the python library and _phylanx(d).cpython-37m-x86_64-linux-gnu.so. The shared libraries generated during the build and phylanx dependencies are then packed with the wheel. The rpaths of these binaries is fixed to remove external dependency of wheel to make it a standalone package with no external dependencies.

### DIFF
--- a/python/CMakeLists.txt
+++ b/python/CMakeLists.txt
@@ -81,7 +81,7 @@ if (PYTHON_EXECUTABLE)
 
   # for the install step it does not matter which setup.py we use
   install(CODE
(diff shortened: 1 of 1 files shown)


hkaiser:
Will this work on other platforms (other than Linux) as well?

Also, shouldn't we leave the current install step in place (which simply copies the Python files to the proper spots in the binary directories) to ensure that Phylanx can still be run without being installed?

git-kale:
Although I've tested it only on the Linux system, the command of the creation of the wheel is the same on all operating systems so It will work on other platforms. The PR is WIP and the purpose of it is integration if wheel creation in CI

Issue #1211, opened 2020-07-16
Add support for `time` primitive in PhySL


taless474:
Having the following code:
```c++
#include <phylanx/phylanx.hpp>

#include <hpx/hpx_init.hpp>
#include <hpx/include/lcos.hpp>
#include <hpx/include/util.hpp>
#include <hpx/modules/testing.hpp>

#include <cstdint>
#include <iostream>
#include <string>
#include <utility>
#include <vector>

///////////////////////////////////////////////////////////////////////////////
char const* const conv1d_code = R"(block(
    define(conv,
        block(
            define(array,
                random_d(8,10,2),
            define(kernel,
                random(list(4, 2, 100))),
            conv1d_d(array, kernel)
        )
    ),
    conv
))";

////////////////////////////////////////////////////////////////////////////////
int hpx_main(int argc, char* argv[])
{
    using namespace phylanx::execution_tree;

    // compile the given code
    compiler::function_list snippets;
    auto const& compiled_code = compile("conv", conv1d_code, snippets);
    auto conv = compiled_code.run();


    std::cout << "Having "
        << hpx::get_num_localities(hpx::launch::sync)
        << " localities:\n";


    hpx::util::high_resolution_timer t;

    auto result = conv();
    auto elapsed = t.elapsed();

    std::cout << "Result of conv1d_d for the given csv file and the kernel is calculated in: " 
                   << elapsed << " seconds" << std::endl;


    return hpx::finalize();
}

int main(int argc, char* argv[])
{
    std::vector<std::string> cfg = {"hpx.run_hpx_main!=1"};

    return hpx::init(argc, argv, cfg);
}
```
the elapsed time contains the execution time of 3 primitives: `random_d`, `random` and `conv1d_d`. 
It would be really useful to have a primitive that enables us to separate these measurements. Having able to use hpx::util::high_resolution_timer in physl can be a solution.

hkaiser:
@taless474 I'd suggest to add something like:
```
char const* const conv1d_code = R"(block(
    define(conv,
        block(
            define(array,
                random_d(8,10,2),
            define(kernel,
                random(list(4, 2, 100))),
            timer(
                conv1d_d(array, kernel),
                lambda(time, cout(time))
        )
    ),
    conv
))";
```
i.e. a primitive that times the execution of its first argument and passes the collected time to the function that is given as the second argument.

PR #1213, opened 2020-07-17
Adding timer() primitive

hkaiser:
- flyby: fixing #includes for HPX changes

Fixes: #1211 

@taless474 could you please verify that this is doing what you need? The syntax is as outlined in #1211:
```
timer(<any_expression>, <any_unary_function>)
```
where 
- `<any_expression>` is -- well -- any PhySL expression you would like to measure the execution time for
- `<any_unary_function>` is any function that will be invoked with a single argument - the execution time of the expression in seconds (float)

The `timer()` primitive will return whatever the expression has returned. Thus this primitive can be easily inserted into any spot in an existing PhySL code without changing its semantics. A simple example would be:
```
timer(
    block(
        define(x, 3.14),
        x
    ),
    lambda(x, cout("time: ", x, " [s]"))
)
```
which would evaluate to `3.14` as expected, but would additionally print the measured time.

### DIFF
--- a/tests/regressions/execution_tree/CMakeLists.txt
+++ b/tests/regressions/execution_tree/CMakeLists.txt
@@ -11,6 +11,7 @@ set(tests
     817_cout_nil
     817_lambda_nil
     1198_inaccessible_variable
+    1213_timer_problem
     allow_empty_blocks_278
     assign_variable_twice_491
     assign_wrong_variable_245
@@ -45,6 +46,7 @@ set(out_of_scope_variable_232_FLAGS DEPENDENCIES HPX::iostreams_component)
 set(external_functions_337_FLAGS DEPENDENCIES HPX::iostreams_component)
 
 set(random_d_1139_PARAMETERS LOCALITIES 2)
+set(1213_timer_problem_PARAMETERS LOCALITIES 2)
 
 foreach(test ${tests})
   set(sources ${test}.cpp)
--- a/phylanx/execution_tree/primitives.hpp
+++ b/phylanx/execution_tree/primitives.hpp
@@ -1,4 +1,4 @@
-//  Copyright (c) 2017-2019 Hartmut Kaiser
+//  Copyright (c) 2017-2020 Hartmut Kaiser
 //                2018 R. Tohid
 //
 //  Distributed under the Boost Software License, Version 1.0. (See accompanying
@@ -28,6 +28,7 @@
 #include <phylanx/execution_tree/primitives/store_operation.hpp>
 #include <phylanx/execution_tree/primitives/string_output.hpp>
 #include <phylanx/execution_tree/primitives/target_reference.hpp>
+#include <phylanx/execution_tree/primitives/timer.hpp>
 #include <phylanx/execution_tree/primitives/variable.hpp>
 #include <phylanx/execution_tree/primitives/variable_factory.hpp>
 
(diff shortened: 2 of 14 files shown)


taless474:
Thank you. This test:
```C++
#include <phylanx/phylanx.hpp>

#include <hpx/hpx_init.hpp>
#include <hpx/include/lcos.hpp>
#include <hpx/modules/testing.hpp>

#include <cstdint>
#include <iostream>
#include <string>
#include <utility>
#include <vector>

char const* const conv1d_code = R"(block(
    define(conv,
        block(
            define(array,
                random_d(list(8, 10, 2))),
            define(kernel,
                random(list(4, 2, 100))),
            timer(
                conv1d_d(array, kernel),
                lambda(time, cout(time))
            )
        )
    ),
    conv
))";

////////////////////////////////////////////////////////////////////////////////
int hpx_main(int argc, char* argv[])
{
    using namespace phylanx::execution_tree;

    // compile the given code
    compiler::function_list snippets;
    auto const& compiled_code = compile("conv", conv1d_code, snippets);
    auto conv = compiled_code.run();


    std::cout << "Having "
        << hpx::get_num_localities(hpx::launch::sync)
        << " localities:\n";

    auto result = conv();

    return hpx::finalize();
}

int main(int argc, char* argv[])
{
    std::vector<std::string> cfg = {"hpx.run_hpx_main!=1"};

    return hpx::init(argc, argv, cfg);
}
```
on docker made this error:
```console
1: {locality-id}: 1
1: {hostname}: [ (tcp:127.0.0.1:7911) ]
1: {process-id}: 3742
1: {os-thread}: 0, locality#1/worker-thread#0
1: {thread-id}: 00000000e8036c90
1: {thread-description}: phylanx_primitive_eval_single_action
1: {state}: state_running
1: {auxinfo}:
1: {file}: /repos/phylanx/src/execution_tree/primitives/access_argument.cpp
1: {line}: 80
1: {function}: virtual hpx::future<primitive_argument_type> phylanx::execution_tree::primitives::access_argument::eval(const phylanx::execution_tree::primitive_arguments_type &, phylanx::execution_tree::eval_context) const
1: {what}: Assertion 'operands_.size() > 1 && (valid(operands_[1]) || is_explicit_nil(operands_[1]))' failed: HPX(assertion_failure)
```

It's probably my problem, but I am having trouble building Phylanx on Windows currently.

hkaiser:
For me, your example fires this exception: https://github.com/STEllAR-GROUP/phylanx/blob/timer_primitive/src/plugins/dist_keras_support/dist_conv1d.cpp#L280. I'm not sure what to do about that.

hkaiser:
@stevenrbrandt I'm not sure what's wrong with my help text (see here: https://app.circleci.com/pipelines/github/STEllAR-GROUP/phylanx/628/workflows/88d287af-09b9-44f1-ac52-a648e3f3317b/jobs/39953/tests) Would you be able to have a look?

hkaiser:
> For me, your example fires this exception: https://github.com/STEllAR-GROUP/phylanx/blob/timer_primitive/src/plugins/dist_keras_support/dist_conv1d.cpp#L280. I'm not sure what to do about that.

I have your problem reproduced now. The exception above occurs when the example is run on one locality only.

hkaiser:
@taless474 please try again, I hope the problem is solved now.

taless474:
It works properly, thanks.

tests/regressions/execution_tree/CMakeLists.txt
```diff
@@ -11,6 +11,7 @@ set(tests
     817_cout_nil
     817_lambda_nil
     1198_inaccessible_variable
+    1213_timer_problem
     allow_empty_blocks_278
```
taless474, line 15: I think we missed adding set(1213_timer_problem_PARAMETERS LOCALITIES 2)

PR #1215, opened 2020-07-22
Add functions for slicing distributed arrays locally

NanmiaoWu:
This PR:

1. Renames `slice_row_d` and `slice_column_d` to denote the slice operation that works on distributed arrays with index on the whole  array.

2. Adds functions for assigning 1d vector to distributed array with index working locally. This will return the distributed array (with some rows being modified) still with annotation.

3. Adds tests for extracting 1d vector from distributed array with local index by using the functions for non-distributed arrays. This will return some 1d vectors without annotation.


src/execution_tree/primitives/slice.cpp
```diff
@@ -275,16 +280,85 @@ namespace phylanx { namespace execution_tree
                     name, codename));
         }
 
-        if (data.has_annotation() || value.has_annotation())
+//        if (data.has_annotation() || value.has_annotation())
```
taless474, line 283: I guess we can remove these commented lines

src/plugins/matrixops/slicing_operation.cpp
```diff
@@ -74,27 +74,71 @@ namespace phylanx {namespace execution_tree {    namespace primitives
...
 
+    ///////////////////////////////////////////////////////////////////////////
+    namespace detail
+    {
+        slicing_operation::slice_mode slicing_operation_mode(
+            std::string const& name)
+        {
+            slicing_operation::slice_mode result
+                = slicing_operation::local_mode;
+            if ((name.find("slice_row_d") != std::string::npos) ||
```
taless474, line 123: There is something that I am not sure I understand. Does `slice_d` work on its own?

src/execution_tree/primitives/slice.cpp
NanmiaoWu: Yes. Will delete it.

src/plugins/matrixops/slicing_operation.cpp
NanmiaoWu: No. I think I need to delete slice_d here.

PR #1222, opened 2020-07-31
Make sure physl interpreter works in spmd mode by default

hkaiser:
- flyby: enable mult_cannond_d to run on one locality


hkaiser:
@stevenrbrandt this removes the need for the `hpx:run-hpx-main` command line option when running the physl interpreter on more than one locality.

PR #1231, opened 2020-08-12
Allow a null values to be assigned to a dict, e.g. d["k"] = None

stevenrbrandt:
Allow a null values to be assigned to a dict, e.g. d["k"] = None

### DIFF
--- a/src/execution_tree/primitives/store_operation.cpp
+++ b/src/execution_tree/primitives/store_operation.cpp
@@ -76,7 +76,7 @@ namespace phylanx { namespace execution_tree { namespace primitives
                         "two operands"));
         }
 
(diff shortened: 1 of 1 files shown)


src/execution_tree/primitives/store_operation.cpp
```diff
@@ -76,7 +76,7 @@ namespace phylanx { namespace execution_tree { namespace primitives
                         "two operands"));
         }
 
-        if (!valid(operands_[0]) || !valid(operands_[1]))
+        if (!valid(operands_[0]))
```
hkaiser, line 79:
```suggestion
        if (!valid(operands_[0]) || is_implicit_nil(operands[1])) 
```
This would still throw if for some reason the right hand side is invalid, but would allow for assigning explicit `None`.

Issue #1252, opened 2020-09-16
buildbot build error


khuck:
Buildbot build is failing with the following error:

http://omega.nic.uoregon.edu:8020/#/builders/5/builds/258/steps/15/logs/stdio

```
In file included from /var/lib/buildbot/workers/phylanx/x86_64-gcc7-debug/build/phylanx/execution_tree/primitives/base_primitive.hpp:1193:0,
                 from /var/lib/buildbot/workers/phylanx/x86_64-gcc7-debug/build/src/execution_tree/locality_annotation.cpp:9:
/var/lib/buildbot/workers/phylanx/x86_64-gcc7-debug/build/phylanx/execution_tree/primitives/primitive_component_base.hpp: In constructor ‘phylanx::execution_tree::match_pattern_type::match_pattern_type(phylanx::execution_tree::match_pattern_type::match_pattern_data&&)’:
/var/lib/buildbot/workers/phylanx/x86_64-gcc7-debug/build/phylanx/execution_tree/primitives/primitive_component_base.hpp:188:61: error: no matching function for call to ‘get<0>(phylanx::execution_tree::match_pattern_type::match_pattern_data&)’
           : primitive_type_(std::move(hpx::util::get<0>(data)))
                                                             ^
```

PR #1253, opened 2020-09-17
Attempting to fix compilation problem related to hpx::tuple

hkaiser:
- also adapt to new namespace for hpx::tuple and related facilities

@khuck could you please verify whether this fixes the issue you reported (#1252)? I'm not sure what's causing the problem, so this is just a blind attempt to fix things.


khuck:
@hkaiser yeah, this seems to fix things.  I am still compiling, but it's past 60%, whereas the build failure from the issue was happening 10% in.  thanks!

PR #1262, opened 2020-09-25
Teaching the front end about assert

### LINKED ISSUE #1260: Assert does not work
If one uses

`assert data["input_nodes_sz"] != 0`

it does not work, however adding parenthesis works

`assert(data["input_nodes_sz"] != 0)`

we might, could add them in the frontend instead.

hkaiser:
- adding second (optional) argument to assert_condition()
  primitive (message to print)
- flyby: fix compiler output if number of arguments does not match
  expected interface

Fixes #1260


hkaiser:
Merging despite problem in traverse_ast. This will be fixed separately.

PR #1297, opened 2021-01-06
Fixing collecting performance counters for distributed runs of physl

hkaiser:
- flyby: fixing warnings from changes in HPX

### DIFF
--- a/examples/interpreter/physl.cpp
+++ b/examples/interpreter/physl.cpp
@@ -20,7 +20,7 @@
 #include <utility>
 #include <vector>
 
-#include <hpx/filesystem.hpp>
+#include <hpx/modules/filesystem.hpp>
 #include <hpx/format.hpp>
 #include <hpx/program_options.hpp>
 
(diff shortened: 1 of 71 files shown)


examples/interpreter/physl.cpp
```diff
@@ -513,9 +513,30 @@ void print_newick_tree(std::string const& code_source_name,
...
+    std::string const& code_source_name, std::string const& _dot_file,
+    std::string const& _newick_tree_file, std::string const& _counter_file)
 {
+    std::int64_t _locality =
+        hpx::naming::get_locality_id_from_id(hpx::find_here());
+
+    std::ostringstream cf;
+    cf << _counter_file;
+    if(_locality > 0)
+        cf << "." << _locality;
```
hkaiser, line 525: Using an `ostringstream is a bit overkill for this situation, I think. Could you just concatenate like:
```suggestion
    if(_locality > 0)
        _counter_file += std::to_string(_locality);
```

PR #1298, opened 2021-01-07
Address the removal of Python 2 from CI image

parsa:
Addresses:
* Removal of Python 2 from the Docker image
* Update `PYTHONPATH` to reflect Python 3.7 to 3.8 upgrade for the `help_format` check.

Note:
* The test fails do not look relevant. Need to be investigated separately:
    * <https://gist.github.com/parsa/d603f60b8827229ccc476e3e0a6dc5a8>

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -291,7 +291,7 @@ jobs:
             - run:
                 name: Check the formatting of Phylanx's Python files
                 environment:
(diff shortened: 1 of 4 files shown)


hkaiser:
@parsa thanks a lot!

hkaiser:
Let's merge this as it fixes part of the issues.

PR #1319, opened 2021-03-08
Transferred bytes counters

hkaiser:
Add performance counters that return the number of transferred bytes for the distributed primitives


taless474:
Is it possible for you to enable `--performance` option for tests written in C++?

PR #1320, opened 2021-03-25
Fixing problem that started to show up with newer cmake versions

hkaiser:


kmoham6:
Thanks, It solved my problem.

